PR #43, opened 2022-02-21
[#42] Automatic PID file path definition

### LINKED ISSUE #42: Automatically define a `pidfile` value when none is set in the configuration
When the user does not specify any `pidfile` configuration option, the software should set automatically to a predefined value.
In particular, the value should be dependedant on other configuration options, like `host`.

fluca1978:
In the case the user does not set the `pidfile` configuration
parameter, the software should define one path itself.
In particular the change makes the `pidfile` equal to the concatenation
of the `unix_socket_dir` (mandatory) and the string `pgmoneta.<host>.pid`
where `<host>` is the value of the `host` configuration parameter
or `all` if `host = *`.

Changed the `pidfile` variable size.

Updated the documentation and the AUTHORS file.

Close #42.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -3,3 +3,4 @@ pgmoneta was created by the following authors:
 Jesper Pedersen <jesper.pedersen@redhat.com>
 David Fetter <david@fetter.org>
 Will Leinweber <will@bitfission.com>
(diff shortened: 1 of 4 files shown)


jesperpedersen:
Squashed and merged.

PR #45, opened 2022-02-21
[#44] Handle configuration file with no newline at the end.

### LINKED ISSUE #44: Parse configuration file even if it does not end with a new-lne
In the case of a configuration file `pgmoneta.conf` that does end with a newline, the application does not read the last line of the configuration.
The end result is that the program could fail with a misleading error message.
For instance

```
% pgmoneta -c /etc/pgmoneta/pgmoneta.conf -u /etc/pgmoneta/pgmoneta_users.conf
    FATAL configuration.c:813 pgmoneta: No user defined for primary
```

with a configuration file that ends with

```
% cat /etc/pgmoneta/pgmoneta.conf
...
[primary]
host = venkman
port = 5432
user = luca%
```

(without a new line).
Adding a new line makes the error disappear.

fluca1978:
In the case the last line of the configuration file does not have
a new line, such line is not parsed at all. This means the
key/value pair is not extracted at all.
This could cause misleading errors.

Close #44

### DIFF
--- a/src/libpgmoneta/configuration.c
+++ b/src/libpgmoneta/configuration.c
@@ -1279,7 +1279,7 @@ extract_key_value(char* str, char** key, char** value)
       while (str[c] != ' ' && str[c] != '\r' && str[c] != '\n' && c < length)
          c++;
 
-      if (c < length)
+      if (c <= length)
       {
          v = malloc((c - offset) + 1);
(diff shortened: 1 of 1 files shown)


jesperpedersen:
Rebased and merged.

PR #83, opened 2022-08-26
[pgmoneta#77] Adding SSH Link functionality

MariamFahmy98:
Implementing links for the remote backups to limit backup size.

### DIFF
--- a/src/libpgmoneta/se_ssh.c
+++ b/src/libpgmoneta/se_ssh.c
@@ -30,9 +30,12 @@
 
 /* pgmoneta */
 #include <pgmoneta.h>
+#include <hashmap.h>
+#include <info.h>
 #include <logging.h>
-#include <stdio.h>
+#include <string.h>
 #include <utils.h>
+#include <security.h>
 #include <storage.h>
 
 /* system */
@@ -46,13 +49,26 @@ static int ssh_storage_setup(int, char*, struct node*, struct node**);
 static int ssh_storage_execute(int, char*, struct node*, struct node**);
 static int ssh_storage_teardown(int, char*, struct node*, struct node**);
 
-static int sftp_make_dir(char* local_dir, char* remote_dir);
-static int sftp_copy_directory(char* from, char* to);
-static int sftp_copy_file(char* from, char* to);
+static char* get_remote_server_basepath(int server);
+static char* get_remote_server_backup(int server);
+static char* get_remote_server_backup_identifier(int server, char* identifier);
+
+static int read_latest_backup_sha256(char* path);
+
+static int sftp_make_directory(char* local_dir, char* remote_dir);
+static int sftp_copy_directory(char* local_root, char* remote_root, char* relative_path);
+static int sftp_copy_file(char* local_root, char* remote_root, char* relative_path);
 
 static ssh_session session = NULL;
 static sftp_session sftp = NULL;
+
+static struct hashmap* hash_map = NULL;
+
 static bool is_error = false;
+
+static char** file_paths = NULL;
+static char** hashes = NULL;
+static char* latest_remote_root = NULL;
 
 struct workflow*
 pgmoneta_storage_create_ssh(void)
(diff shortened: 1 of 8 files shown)


MariamFahmy98:
It is neither completed nor tested.

MariamFahmy98:
It is now ready to be reviewed. Link functionality works good. However the code needs some refactoring and clean ups.
The retention isn't supported yet.

src/libpgmoneta/se_ssh.c
```diff
@@ -234,38 +254,90 @@ ssh_storage_execute(int server, char* identifier,
...
-   struct configuration* config;
+   char* latest_backup_sha256 = NULL;
+   char* current_backup_sha256 = NULL;
+   char* remote_backup_dir = NULL;
+   int next_newest = -1;
+   int number_of_backups = 0;
+   struct backup** backups = NULL;
 
-   config = (struct configuration*)shmem;
+   current_remote_paths = (char**)malloc(sizeof(char*) * 970);
```
jesperpedersen, line 264: 970 ?

src/libpgmoneta/se_ssh.c
```diff
@@ -382,85 +477,640 @@ sftp_copy_directory(char* from, char* to)
...
 
-   memset(buffer, 0, sizeof(buffer));
+   map_size = pgmoneta_hashmap_size(hash_map);
+
+   if (pgmoneta_hashmap_key_set(hash_map, &keys))
+   {
+      goto error;
+   }
+
+   for (int i = 0; i < map_size; i++)
```
jesperpedersen, line 933: Can't we just write the entire file ?

src/libpgmoneta/se_ssh.c
```diff
@@ -382,85 +477,640 @@ sftp_copy_directory(char* from, char* to)
...
+   }
+
+   while ((read_bytes = sftp_read(file, &buffer[0], sizeof(buffer))) > 0)
+   {
+      fwrite(&buffer[0], read_bytes, 1, tmp_file);
+   }
+
+   fclose(tmp_file);
+
+   latest_remote_paths = (char**)malloc(sizeof(char*) * 970);
```
jesperpedersen, line 1019: 970 ... I wonder what the reference is :)

src/libpgmoneta/se_ssh.c
```diff
@@ -382,85 +477,640 @@ sftp_copy_directory(char* from, char* to)
...
+{
+   char* d = NULL;
+
+   d = get_remote_server_basepath(server);
+   d = pgmoneta_append(d, "backup/");
+
+   return d;
+}
+
+char*
```
jesperpedersen, line 1107: static ?

jesperpedersen (changes requested):

src/libpgmoneta/se_ssh.c
MariamFahmy98: We can directly write into the remote `backup.sha256`. However, for each file we will open the remote file and then append it a new entry because files are copied/linked in `sftp_copy_file()`. This is why I find it better to store all file paths with its corresponding hashes first in a hashmap and then write into `backup.sha256` at the end. What do you think ?

jesperpedersen: Sure, but we need `backup.sha256` locally as well, right ?

MariamFahmy98: We can write `backup.sha256` locally at the beginning, and once all files are copied/linked to the remote server. We can copy the local `backup.sha256` to the remote server, and delete the file locally. As the data directory is removed after the remote backup finished.
At this point, whenever `sftp_copy_file()` is triggered, I create a new entry of the file path with its hash and store it in a hashmap. After the remote backup finishes, I write the whole hashmap into the remote `backup.sha256` without the need of having a `backup.sha256` locally.

jesperpedersen: But, `backup.sha256` lives outside of `data/` - next to `backup.info`. So, I'm missing why we need a SHA256 of `backup.sha256` itself

MariamFahmy98: There is 2 approaches we can follow one of them:
1- Creating `backup.sha256` locally, and whenever `sftp_copy_file()` is triggered, The file path with its hash will be written into `backup.sha256` locally. At the end of `ssh_storage_execute()` we will copy the `backup.sha256` to the remote server and then we delete the file locally in `ssh_storage_teardown()` as we do the backup/data directory.

2- Storing each file path with its hash in a hashmap and after the remote backup finishes, we iterate over the hashmap and write the entry in `backup.sha256` remotely.

If there is a better approach, you can share it with me.

jesperpedersen: Separate workflow with just the SHA-256 stuff after compression, but before SFTP.

jesperpedersen: https://github.com/pgmoneta/pgmoneta/blob/main/src/libpgmoneta/workflow.c#L133

MariamFahmy98: I get it. We will create `backup.sha256` locally in a separate workflow. Is it right ?
If it is, `backup.sha256` stores the remote file path with its hash so that we can just make a link to the latest remote file.
But if we create `backup.sha256` locally before SFTP, we will store the local file path not the remote one.

MariamFahmy98: Oh I messed up. I am actually storing the reference path not the full path. You are right. I will create a separate workflow for it.

jesperpedersen: Can't we just store the relative path from `data/` ? We know if we are operating locally or remotely so we can build the full paths from that

jesperpedersen: So,

```
SHA256............:postgresql.conf
```

f.ex

jesperpedersen: Or the other way around

jesperpedersen: Great, lets see if that works better :)

MariamFahmy98: I get it.  we will create a separate workflow for writing `backup.sha256` locally before the SFTP, and at the end we will copy this file to the remote server.
In case of reading `backup.sha256` of the latest backup to decide if we either make a copy or do a link. We will read this file locally too ?

jesperpedersen: Well, we can if decide to keep it locally (initial commit I would say yes), otherwise it is a matter of just reading `x` and `x - 1` from the remote server. I think you will get a better idea of what is best during the refactoring

MariamFahmy98: Hashmaps will be used only when reading `backup.sha256` ?
We will read it at the beginning and store its entries in hashmap so that we can use it later to decide if we will copy the file or do a link ?

jesperpedersen: Yes

MariamFahmy98: Great, thanks. I will work on it and I will let you know once I finish it :)

MariamFahmy98:
You can take a look any time. Thanks.

src/libpgmoneta/security.c
```diff
@@ -2719,3 +2720,54 @@ create_ssl_server(SSL_CTX* ctx, int socket, SSL** ssl)
...
+   SHA256_Final(hash, &sha256_ctx);
+
+   for (i = 0; i < SHA256_DIGEST_LENGTH; i++)
+   {
+      sprintf(&sha256_buf[i * 2], "%02x", hash[i]);
+   }
+
+   sha256_buf[64] = 0;
+
+   pgmoneta_log_debug("sha_buff: %s", sha256_buf);
```
jesperpedersen, line 2766: Lets do trace or remove it

src/libpgmoneta/workflow.c
```diff
@@ -131,6 +131,9 @@ wf_backup(void)
    current->next = pgmoneta_workflow_create_permissions(PERMISSION_TYPE_BACKUP);
    current = current->next;
 
+   current->next = pgmoneta_workflow_create_sha256();
```
jesperpedersen, line 134: Lets stick it under `if (config->storage_engine == STORAGE_ENGINE_SSH)` - local doesn't need it atm

jesperpedersen (changes requested):

jesperpedersen:
Rebased, and merged.

PR #113, opened 2023-03-14
allow prefix matching for incomplete identifier

Jubilee101:
If an identifier cannot be found, try matching prefix of existing backups

### DIFF
--- a/src/libpgmoneta/wf_restore.c
+++ b/src/libpgmoneta/wf_restore.c
@@ -169,8 +169,28 @@ restore_execute(int server, char* identifier, struct node* i_nodes, struct node*
 
    if (!pgmoneta_exists(base))
    {
-      pgmoneta_log_error("Restore: Unknown identifier for %s/%s", config->servers[server].name, id);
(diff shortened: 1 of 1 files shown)


src/libpgmoneta/wf_restore.c
```diff
@@ -169,8 +169,28 @@ restore_execute(int server, char* identifier, struct node* i_nodes, struct node*
 
    if (!pgmoneta_exists(base))
    {
-      pgmoneta_log_error("Restore: Unknown identifier for %s/%s", config->servers[server].name, id);
-      goto error;
+       if (pgmoneta_get_backups(root, &number_of_backups, &backups))
```
jesperpedersen, line 172: Indent by 3

jesperpedersen (changes requested):

Jubilee101:
I think I fixed it. Not entirely sure though. Let me know if there's problem.

jesperpedersen:
Rebased, and merged.

PR #122, opened 2023-05-11
Apply existing patch

Jubilee101:
The first commit consists of changes only from the patch. 

I thought I'd do some cleanup or modifications on the basis of the patch directly, but on second thought it might be better if we could have some discussions first @jesperpedersen. I've already made some comments in the code (some are just plain questions about the code).

I'll leave the commits unsquashed to tell the changes apart. And I will squash the changes once we agree it's good enough to merge.

### DIFF
--- a/src/include/message.h
+++ b/src/include/message.h
@@ -55,6 +55,26 @@ struct message
    void* data;        /**< The message data */
 } __attribute__ ((aligned (64)));
 
+/** @struct
+ * Defines a tuple
+ */
+struct tuple
+{
+   char** data;                   /**< The data */
+   struct tuple* next;            /**< The next tuple */
+} __attribute__ ((aligned (64)));
+
+/** @struct
+ * Defines the response to a query
+ */
+struct query_response
+{
+   char names[MAX_NUMBER_OF_COLUMNS][MISC_LENGTH]; /**< The column names */
+   int number_of_columns;                          /**< The number of columns */
+
+   struct tuple* tuples;
+} __attribute__ ((aligned (64)));
+
 /**
  * Read a message in blocking mode
  * @param ssl The SSL struct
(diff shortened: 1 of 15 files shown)


src/include/message.h
```diff
@@ -55,6 +55,26 @@ struct message
...
+/** @struct
+ * Defines a query
+ */
+struct query
+{
+   char names[MAX_NUMBER_OF_COLUMNS][MISC_LENGTH]; /**< The column names */
+   int number_of_columns;                          /**< The number of columns */
+
+   struct tuple* tuples;
+} __attribute__ ((aligned (64)));
```
Jubilee101, line 76: I'm not sure what this means. Does the struct contains all the tuples the query touches?

src/libpgmoneta/memory.c
```diff
@@ -122,3 +122,40 @@ pgmoneta_memory_destroy(void)
    data = NULL;
    message = NULL;
 }
+
+void*
+pgmoneta_memory_dynamic_create(size_t* size)
+{
+   *size = 0;
+
```
Jubilee101, line 130: All it does is set the size? What about the memory it created? @jesperpedersen

src/libpgmoneta/memory.c
```diff
@@ -122,3 +122,40 @@ pgmoneta_memory_destroy(void)
...
+pgmoneta_memory_dynamic_append(void* orig, size_t orig_size, void* append, size_t append_size, size_t* new_size)
+{
+   void* d = NULL;
+   size_t s;
+
+   if (append != NULL)
+   {
+      s = orig_size + append_size;
+      d = realloc(orig, s);
+      memcpy(d + orig_size, append, append_size);
```
Jubilee101, line 150: can we assume `append_size` is always at most the size of memory `append` is pointing to?

src/libpgmoneta/message.c
```diff
@@ -535,6 +540,11 @@ pgmoneta_create_startup_message(char* username, char* database, struct message**
    ds = strlen(database);
    size = 4 + 4 + 4 + 1 + us + 1 + 8 + 1 + ds + 1 + 17 + 9 + 1;
```
Jubilee101, line 541: This is very confusing for me

src/libpgmoneta/message.c
```diff
@@ -552,11 +562,518 @@ pgmoneta_create_startup_message(char* username, char* database, struct message**
    pgmoneta_write_string(m->data + 13 + us + 1 + 9 + ds + 1, "application_name");
    pgmoneta_write_string(m->data + 13 + us + 1 + 9 + ds + 1 + 17, "pgmoneta");
```
Jubilee101, line 563: Same as above

src/libpgmoneta/message.c
```diff
@@ -535,6 +540,11 @@ pgmoneta_create_startup_message(char* username, char* database, struct message**
    ds = strlen(database);
    size = 4 + 4 + 4 + 1 + us + 1 + 8 + 1 + ds + 1 + 17 + 9 + 1;
 
+   if (replication)
+   {
+      size += 14;
```
Jubilee101, line 545: same comment as above

src/libpgmoneta/message.c
```diff
@@ -552,11 +562,518 @@ pgmoneta_create_startup_message(char* username, char* database, struct message**
...
+/*    for (int i = 0; i < query->number_of_columns; i++) */
+/*    { */
+/*       if (!strcmp(query->names[i], name)) */
+/*       { */
+/*          return pgmoneta_get_column(i, tuple); */
+/*       } */
+/*    } */
+
+/*    return NULL; */
+/* } */
```
Jubilee101, line 756: I think we need to clean this chunk of comments

src/libpgmoneta/message.c
```diff
@@ -552,11 +562,518 @@ pgmoneta_create_startup_message(char* username, char* database, struct message**
...
+         data = pgmoneta_memory_dynamic_append(data, data_size, reply->data, reply->length, &data_size);
+
+         if (pgmoneta_has_message('Z', data, data_size))
+         {
+            cont = false;
+         }
+      }
+      else if (status == MESSAGE_STATUS_ZERO)
+      {
+         SLEEP(1000000L);
```
Jubilee101, line 799: Just a plain question, what does `MESSAGE_STATUS_ZERO` mean and why do we sleep here?

src/libpgmoneta/message.c
```diff
@@ -552,11 +562,518 @@ pgmoneta_create_startup_message(char* username, char* database, struct message**
...
+   cont = true;
+   while (cont)
+   {
+      status = pgmoneta_read_block_message(NULL, socket, &reply);
+
+      if (status == MESSAGE_STATUS_OK)
+      {
+         data = pgmoneta_memory_dynamic_append(data, data_size, reply->data, reply->length, &data_size);
+
+         if (pgmoneta_has_message('Z', data, data_size))
```
Jubilee101, line 792: How can I find more info about this message type such as 'Z'?

src/libpgmoneta/message.c
```diff
@@ -552,11 +562,518 @@ pgmoneta_create_startup_message(char* username, char* database, struct message**
...
+bool
+pgmoneta_has_message(char type, void* data, size_t data_size)
+{
+   int offset;
+
+   offset = 0;
+
+   while (offset < data_size)
+   {
+      char t = (char)pgmoneta_read_byte(data + offset);
```
Jubilee101, line 897: I'm not familiar with the mechanism here. Is there some agreement on where the type of message will be within the message?

src/include/message.h
```diff
@@ -55,6 +55,26 @@ struct message
...
+   char** data;                   /**< The data */
+   struct tuple* next;            /**< The next tuple */
+} __attribute__ ((aligned (64)));
+
+/** @struct
+ * Defines a query
+ */
+struct query
+{
+   char names[MAX_NUMBER_OF_COLUMNS][MISC_LENGTH]; /**< The column names */
```
Jubilee101, line 72: I feel that we probably need to change the note here, judging from the code below this is more like a response to a query instead of the query itself? And is 8 big enough as the max column number?

src/libpgmoneta/memory.c
jesperpedersen: Yeah, we will start with `NULL`

src/libpgmoneta/memory.c
jesperpedersen: It should always be the current size, and `realloc` will do the rest

src/libpgmoneta/message.c
jesperpedersen: Its the protocol,

Main: https://www.postgresql.org/docs/devel/protocol-message-formats.html
Replication: https://www.postgresql.org/docs/current/protocol-replication.html

src/libpgmoneta/message.c
jesperpedersen: Protocol again

src/include/message.h
jesperpedersen: I think we are ok with 8 for now

Jubilee101: Do you want to reconsider the naming here? `query_response` makes more sense to me.

Jubilee101: Hmm I think that is fine, since we have the comment there. I just feel that this struct is more of the response to a query, judging from the usage in `pgmoneta_query_execute`, unless we have other usage later.

jesperpedersen: Type it up, and lets take a look at it

Jubilee101:
Update:
For the two new commits, the first one is just removing some comments, the second one is to see whether we like it if we change `query` to `query_response`. I personally don't have much confidence about it after I made the change, so let me know what you think @jesperpedersen 
I will squash the commits once we are happy about it.

jesperpedersen:
Lets get this squashed

Jubilee101:
Commits squashed

jesperpedersen:
Added an issue ref, commit message, rebased, and merged.

Issue #123, opened 2023-05-17
Native backup


jesperpedersen:
Replace `pg_basebackup` with a native solution based on the replication protocol

jesperpedersen:
@Jubilee101 Please, fill this out

Jubilee101:
> @Jubilee101 Please, fill this out

Uhh just to make sure, you mean with commits right? Or do I need to put doc here?

jesperpedersen:
It should be the brief summary of your doc

Jubilee101:
Currently pgmoneta generates backups by sending pg_basebackup command to Postgres database server. We want to rid of this dependency on Postgres binaries by implementing the process this command triggers. This roughly involves:

1. Sending the BASE_BACKUP command to the server.
2. Implementing the Streaming Replication Protocol to receive the backup and the necessary write-ahead log.
3. Saving the backup locally, verifying checksum and performing compression if necessary.

Jubilee101:
Hey should we close this issue? Or do we want to use this for more advanced backups as well?

PR #124, opened 2023-05-18
[#123] Implement a native basebackup solution

Jubilee101:
This PR is to address #123, currently it only removes the old implementation.


Jubilee101:
Sorry about that. It's there then, let me how you want me to approach this.

PR #127, opened 2023-06-17
[#123] Receive and consume copy stream messages

Jubilee101:
Finally, I resolved the bug (that was a stupid mistake)! This PR cosists of two functionalities:

One to read data from server into a buffer, it will left shift consumed data to reuse space, I tested locally and it was perfect, did not once exceed its original buffer size. For the sake of simplicity I just use the default buffer size (65535B), in the future we could have some configuration for the user to specify how big they want the stream buffer to be.

Another one is to parse data in the buffer, consume data into messages. It will block and return one message by another, and it will filter out those messages we are not interested in the stream replication process. And it will automatically read new data if current message is not complete.

I tested locally and it works perfectly, just as the protocal says, two DataRow messages, one CopyOutResponse then a bunch of CopyData, then finishes by a CopyDone message.

Let me know if you have questions.

### DIFF
--- a/src/include/memory.h
+++ b/src/include/memory.h
@@ -37,6 +37,18 @@ extern "C" {
 
 #include <stdlib.h>
 
+/** @struct
+ * Defines a streaming buffer
+ */
+struct stream_buffer
+{
+   char* buffer;  /**< allocated buffer holding streaming data */
+   int size;      /**< allocated buffer size */
+   int start;     /**< offset to the first unconsumed data in buffer */
+   int end;       /**< offset to the first position after available data */
+   int cursor;    /**< next byte to consume */
+} __attribute__ ((aligned (64)));
+
 /**
  * Initialize a memory segment for the process local message structure
  */
@@ -95,6 +107,29 @@ pgmoneta_memory_dynamic_destroy(void* data);
  */
 void*
 pgmoneta_memory_dynamic_append(void* orig, size_t orig_size, void* append, size_t append_size, size_t* new_size);
+
+/**
+ * Initialize a stream buffer
+ * @param buffer The stream buffer to be initialized
+ */
+void
+pgmoneta_memory_stream_buffer_init(struct stream_buffer** buffer);
+
+/**
+ * Enlarge the buffer, doesn't guarantee success
+ * @param buffer The stream buffer
+ * @param bytes_needed The number of bytes needed
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_memory_stream_buffer_enlarge(struct stream_buffer* buffer, int bytes_needed);
+
+/**
+ * Free a stream buffer
+ * @param buffer The stream buffer to be freed
+ */
+void
+pgmoneta_memory_stream_buffer_free(struct stream_buffer* buffer);
 
 #ifdef __cplusplus
 }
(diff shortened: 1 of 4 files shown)


src/include/memory.h
```diff
@@ -36,6 +36,17 @@ extern "C" {
 #include <pgmoneta.h>
 
 #include <stdlib.h>
+/** @struct
```
jesperpedersen, line 39: Missing line feed above

src/include/memory.h
```diff
@@ -36,6 +36,17 @@ extern "C" {
...
+ * Defines a streaming buffer
+ */
+struct stream_buffer
+{
+   char* buf;  /**< allocated buffer holding streaming data */
```
jesperpedersen, line 44: Lets call it `buffer`

src/include/memory.h
```diff
@@ -36,6 +36,17 @@ extern "C" {
...
+   int size;  /**< allocated buffer size */
```
jesperpedersen, line 45: Lets align documentation comments

src/include/memory.h
```diff
@@ -96,6 +107,29 @@ pgmoneta_memory_dynamic_destroy(void* data);
 void*
 pgmoneta_memory_dynamic_append(void* orig, size_t orig_size, void* append, size_t append_size, size_t* new_size);
 
+/**
+ * Initialize a stream buffer
+ * @param stream_buf The stream buffer to be initialized
+ */
+void
+pgmoneta_memory_stream_buffer_init(struct stream_buffer** stream_buf);
```
jesperpedersen, line 115: `struct stream_buffer** stream_buf` -> `struct stream_buffer** buffer`

jesperpedersen (changes requested):

src/include/pgmoneta.h
```diff
@@ -107,14 +107,14 @@ extern "C" {
 #define unlikely(x)  __builtin_expect (!!(x), 0)
 
 #define MAX(a, b)               \
-        ({ __typeof__ (a) _a = (a);  \
-           __typeof__ (b) _b = (b);  \
-           _a > _b ? _a : _b; })
+   ({ __typeof__ (a) _a = (a);  \
```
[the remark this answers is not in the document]
Jubilee101: I was using 0.75... Anyway it's fixed now.

src/include/memory.h
```diff
@@ -37,6 +37,18 @@ extern "C" {
 
 #include <stdlib.h>
 
+/** @struct
+ * Defines a streaming buffer
+ */
+struct buffer
```
jesperpedersen, line 43: buffer -> stream_buffer

src/include/memory.h
```diff
@@ -96,6 +108,29 @@ pgmoneta_memory_dynamic_destroy(void* data);
 void*
 pgmoneta_memory_dynamic_append(void* orig, size_t orig_size, void* append, size_t append_size, size_t* new_size);
 
+/**
+ * Initialize a stream buffer
+ * @param buffer The stream buffer to be initialized
+ */
+void
+pgmoneta_memory_stream_buffer_init(struct buffer** buffer);
```
jesperpedersen, line 116: buffer** -> stream_buffer**

jesperpedersen (changes requested):

Jubilee101:
Changes made

jesperpedersen:
Rebased, and merged.

PR #128, opened 2023-06-23
[#123] Receive and parse DataRow messages in stream buffer

Jubilee101:
I had to re-implement this functionality similar to `pgmoneta_query_execute`. Because now all the data is in stream buffer. I also made some changes to `pgmoneta_consume_copy_stream` so that I can make use of existing APIs. I think it's better to submit this patch as a sperate PR because the next one (archive_streamer) should be really big and it'll take some time to finish.


jesperpedersen:
Rebased, and merged.

PR #137, opened 2023-07-16
Replace archive functionality with libarchive

Jubilee101:
I verified the correctness and this seems alright. Server starts normally. Note that this replacement doesn't do anything special with symlinks or tablespaces currently. I think this part should be first handled in restore. So it only does what our archive functionality used to do, nothing more and nothing less.


jesperpedersen:
I'm going to leave this pull request open until the full tablespace support is in

Jubilee101:
@jesperpedersen Another problem, we probably need to add another layer of directory to wrap base directory and other tablespaces. Because now we need to tar them all together. If we just tar everything under their root directory, it will cause two problems: First tar functionality will try to tar itself since it's also under this directory. Secondly if there are multiple backups waiting to be archived, tar functionality will tar them all. I propose wrap it under a directory called `archive-timestamp/`

Jubilee101:
Changes to accomodate tablespaces have been submitted. I had to make a lot changes and decisions to make this work:
1. Support relative path when restoring backup. Because creating symlink only takes absolute path, I have to convert relative path using realpath()
2. When restoring before making an archive, base directory and other tablespaces are wrapped under a directory called `archive-server_name-id`. And the tar directory is like this:
.                                                                                       .
= archive-server_name-id                                              = server_name-id 
    == server_name-id                               instead of      = server_name-id-tablespace1
    == server_name-id-tablespace1                                = server_name-id-tablespace2
    == server_name-id-tablespace2
This extra layer will ensure that the symlinks are valid when untared.
Restoring without making an archive is not affected. There won't be this extra archive layer.
4. There is a nasty bug in `pgmoneta_delete_directory`. Since we now support symlinks, using `stat` will cause it to trace the link. And error occurs if the target directory is deleted before it. Leaving the rest of the directory not deleted. The solution is simple however, replacing with `lstat` solves the problem.

src/libpgmoneta/archive.c
```diff
@@ -39,27 +39,35 @@
 
 #include <archive.h>
 #include <archive_entry.h>
+#include <dirent.h>
 #include <stdio.h>
 
+void write_tar_file(struct archive* a, char* current_real_path, char* current_save_path);
```
jesperpedersen, line 45: static

jesperpedersen (changes requested):

Jubilee101:
> The tablespace link in the archive doesn't work - it assume a path once extracted.
> 
> Create the archive in one location, move it to another and extract -> link is wrong

It is correct if you don't move it... The link is created during restoring. So it doesn't know where you are going to move and extract the archive to. We have little control on the extraction process. Unless we provide our own extraction functionality in the cli. Then we can change the symlink when extracting it. Otherwise we can only assume user makes the archive to the places they want in the first place.

jesperpedersen:
Could you make the symlink relative ? Like, `13898 -> ../../primary-xyz-ts1/` ?

Jubilee101:
> Could you make the symlink relative ? Like, `13898 -> ../../primary-xyz-ts1/` ?

Unfortunately no. Symlinks have to use absolute path..

jesperpedersen:
I changed the API a bit... maybe `man 2 symlink` could work...

Jubilee101:
I'll be damned... Took me quite some time to get right but it actually worked! I feel so dumb just arbitrarily saying no based on my limited experience.

Jubilee101:
The link is like this: `16397 -> ../../primary-20230717193201-table_space_2/`
Tricky part is that you still have to use absolute path for the symlink location when creating symlink.

Jubilee101:
> I changed the API a bit... maybe man 2 symlink could work...

Oh crap I didn't see this. I created a new util api using symlinkat. It does the trick

Jubilee101:
I did some further verification. Looks like postgres takes symlinks using relative path. I can query tablespaces or create db under existing tablespaces normally. And moving archive around doesn't affect things, as long as the file structures under archive remain unchanged.

jesperpedersen:
Rebased, and merged.

Issue #139, opened 2023-07-30
Native receivewal 


Jubilee101:
This issue is for removing our dependency on pg_receivewal. We will create a native receivewal solution based on our previous stream buffer infrastructure.

PR #140, opened 2023-07-30
[#139] Native receivewal solution

Jubilee101:
Hi @jesperpedersen, this is only an initial version of our receivewal solution. I pushed it because I could use some advice before going further. This solution works beautifully and smoothly as I tested locally, I think. Our stream buffer adapts really well, it automatically expands itself just enough to suit the larger xlog messages. I also dealt with cases like server sends data across segment boundary. And it exits on server shutdown or connection close. I also started a server on the received xlogs and everything seems to be normal.

By "initial" I mean it doesn't have the ability to deal with older timelines yet. If server sends CommandComplete for reaching the end of timeline, the while loop will simply break instead of starting again on new timeline. And I wonder if it's necessary since the callback is triggerred regularly, if the process exits because of reaching the end of the timeline, next callback will naturally start the process on the next timeline.

Also there is a serious problem, the process doesn't end when I stop pgmoneta by pressing `ctrl+c`. It seems that SIGINT is not propagating to this child process. I'm not very familiar with how signals are handled in pgmoneta, so I could really use your help here.

### DIFF
--- a/src/include/message.h
+++ b/src/include/message.h
@@ -346,6 +346,14 @@ int
 pgmoneta_create_base_backup_message(int server_version, char* label, bool include_wal, char* checksum_algorithm, struct message** msg);
 
 /**
+ * Send a CopyDone message
+ * @param socket The socket
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_send_copy_done_message(int socket);
+
+/**
  * Create a query message for a simple query
  * @param query The query to be executed on server
  * @param msg The resulting message
--- a/src/include/utils.h
+++ b/src/include/utils.h
@@ -737,6 +737,14 @@ pgmoneta_get_y2000_timestamp(void);
 int
 pgmoneta_convert_base32_to_hex(unsigned char* base32, int base32_length, unsigned char** hex);
 
+/**
+ * Get the file size of a given file
+ * @param file_path The file path
+ * @return The file size, 0 if error occurred
+ */
+size_t
+pgmoneta_get_file_size(char* file_path);
+
 #ifdef DEBUG
 
 /**
--- a/src/libpgmoneta/message.c
+++ b/src/libpgmoneta/message.c
@@ -715,22 +715,24 @@ pgmoneta_create_standby_status_update_message(int64_t received, int64_t flushed,
    struct message* m = NULL;
    size_t size;
 
-   size = 1 + 8 + 8 + 8 + 8 + 1;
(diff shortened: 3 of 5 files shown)


jesperpedersen:
I think it is ok for each WAL process to deal with one specific timeline - so having the process exit is ok.

As to shutdown - can you use `config->running` or `getppid()`  ?

Jubilee101:
Updates:
1. `config->running` works, though sometimes it will lag a little if it's blocking to wait for a message. I don't think killing the process in shutdown callback is a better idea, because by checking `config->running` we can do some cleanup work such as closing the file and the socket. I also logged a warn about partially received WAL segments just like postgres does.
2. I enhanced our handling on server reaching the end of timeline. By protocal server will send us a CopyDone message, upon which we should reply with a CopyDone message as well. Then it sends two consecutive CommandComplete. I add some lines to address this part.

jesperpedersen:
1. Killing the process isn't ideal, so - if we can - lets stay away from it
2. Sounds good

Jubilee101:
> Killing the process isn't ideal, so - if we can - lets stay away from it
> Sounds good

Yeah it's already done. Could you take a look at the code and let me know if there's any problem?

jesperpedersen:
The WAL process is running at 100% CPU...

Jubilee101:
Updates:
1. CPU usage rate is now on average about 3%, and doesn't exceed 4% after using `sleep()`, even when postgres is benchmarking and there's a base backup going on.
2. I took the liberty to change `WAL` in `BASE_BACKUP` command back to false as we talked about. I worry that the backup will take even longer time if there are too many wal segments under `pg_wal`.
3. Serious problem, base backup now is taking too much time even though the outcome is correct. It now takes about 280 secs to finsih. I suspect it has something to do with the backup process and the wal process sharing the socket.

Jubilee101:
Ok, as I mentioned in the forum, I guess my wal implementation doesn't increase the base backup time, larger database does. So I think we are good to go with this wal implemention if there's no other problem. 

We could in the future look into ways to accelerate our base backup process. I don't know how long it normally should take.

jesperpedersen:
I think we should leave WAL as `true`, as copying WAL is optional in the `restore` command

Jubilee101:
> I think we should leave WAL as true, as copying WAL is optional in the restore command

Alright, I changed it back.

jesperpedersen:
Rebased, and merged.

PR #142, opened 2023-08-08
Add clang sanitizer compile options

Jubilee101:


Jubilee101:
Hi @jesperpedersen , sorry about the delay. Clang 16.0.6 is only supported on fedora38 so I have to launch a new instance and go over installation again. The problem complained at runtime is simple, when we try to convert the byte array to an int32 we used `bytes[0] << 24`, when bytes[0] is 255(0b1111) the signed bit is toggled when left shift 24 bits. I suppressed the error by treating it as unsigned. I think this solves the problem. Running backup command after 1 min or so yeilds the correct result.

jesperpedersen:
Merged.

PR #144, opened 2023-08-08
Verify base backup with received manifest

Jubilee101:
Used cJSON to parse the manifest.

### DIFF
--- a/pgmoneta.spec
+++ b/pgmoneta.spec
@@ -7,8 +7,8 @@ URL:           https://github.com/pgmoneta/pgmoneta
 Source0:       https://github.com/pgmoneta/pgmoneta/archive/%{version}.tar.gz
 
 BuildRequires: gcc cmake make python3-docutils zlib zlib-devel libzstd libzstd-devel lz4 lz4-devel bzip2 bzip2-devel
-BuildRequires: libev libev-devel openssl openssl-devel systemd systemd-devel libssh libssh-devel libarchive libarchive-devel
-Requires:      libev openssl systemd zlib libzstd lz4 libssh
+BuildRequires: libev libev-devel openssl openssl-devel systemd systemd-devel libssh libssh-devel libarchive libarchive-devel cjson cjson-devel
+Requires:      libev openssl systemd zlib libzstd lz4 bzip2 libssh libarchive cjson
 
 %description
 pgmoneta is a backup / restore solution for PostgreSQL.
--- a/src/include/manifest.h
+++ b/src/include/manifest.h
@@ -1,0 +1,82 @@
+/*
+ * Copyright (C) 2023 Red Hat
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
+ * are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this list
+ * of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice, this
+ * list of conditions and the following disclaimer in the documentation and/or other
+ * materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its contributors may
+ * be used to endorse or promote products derived from this software without specific
+ * prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+ * OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+ * THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT
+ * OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR
+ * TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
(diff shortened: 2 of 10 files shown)


pgmoneta.spec
```diff
@@ -7,7 +7,7 @@ URL:           https://github.com/pgmoneta/pgmoneta
 Source0:       https://github.com/pgmoneta/pgmoneta/archive/%{version}.tar.gz
 
 BuildRequires: gcc cmake make python3-docutils zlib zlib-devel libzstd libzstd-devel lz4 lz4-devel bzip2 bzip2-devel
-BuildRequires: libev libev-devel openssl openssl-devel systemd systemd-devel libssh libssh-devel libarchive libarchive-devel
+BuildRequires: libev libev-devel openssl openssl-devel systemd systemd-devel libssh libssh-devel libarchive libarchive-devel cjson cjson-devel
 Requires:      libev openssl systemd zlib libzstd lz4 libssh
```
jesperpedersen, line 11: Could you fix `Requires` to include all run-time libraries as well ?

src/include/manifest.h
```diff
@@ -0,0 +1,74 @@
...
+} __attribute__ ((aligned (64)));
+
+/**
+ * Parse the manifest json into manifest struct
+ * @param manifest_path The path to manifest
+ * @param manifest The manifest
+ * @return 0 on parsing success, otherwise 1
+ */
+int
+pgmoneta_parse_manifest(char* manifest_path, struct manifest* manifest);
```
jesperpedersen, line 60: Isn't `pgmoneta_parse_manifest(char* manifest_path, struct manifest** manifest);` better ?

src/libpgmoneta/manifest.c
```diff
@@ -0,0 +1,295 @@
...
+#include <cjson/cJSON.h>
+#include <stdio.h>
+#include <string.h>
+
+static void manifest_init(struct manifest** manifest);
+static void manifest_file_init(struct manifest_file** file, char* path, char* checksum, size_t size);
+static void manifest_free(struct manifest* manifest);
+static void manifest_file_free(struct manifest_file* file);
+
+static void
```
jesperpedersen, line 46: Move `static` functions to the end of the file

src/libpgmoneta/manifest.c
```diff
@@ -0,0 +1,295 @@
...
+{
+   struct manifest_file* f = NULL;
+   f = (struct manifest_file*) malloc(sizeof(struct manifest_file));
+   f->next = NULL;
+   f->checksum = (char*)malloc(strlen(checksum) + 1);
+   f->path = (char*)malloc(strlen(path) + 1);
+   f->size = size;
+
+   memcpy(f->checksum, checksum, strlen(checksum));
+   f->checksum[strlen(checksum)] = '\0';
```
jesperpedersen, line 68: Use `memset` or `cmalloc` instead

src/libpgmoneta/manifest.c
```diff
@@ -0,0 +1,295 @@
...
+   free(file->path);
+   free(file->checksum);
+   free(file);
+}
+
+int
+pgmoneta_manifest_checksum_verify(char* root)
+{
+   char manifest_path[MAX_PATH];
+   char* manifest_sha256 = NULL;
```
jesperpedersen, line 110: -> `algorithm` - and we can't assume SHA256

src/libpgmoneta/manifest.c
```diff
@@ -0,0 +1,295 @@
...
+   {
+      snprintf(manifest_path, MAX_PATH, "%s/%s", root, "backup_manifest");
+   }
+   manifest_init(&manifest);
+   if (pgmoneta_parse_manifest(manifest_path, manifest))
+   {
+      goto error;
+   }
+   // first check manifest checksum
+   pgmoneta_generate_string_sha256_hash(manifest->content, &manifest_sha256);
```
jesperpedersen, line 129: We need to support all algorithms

src/libpgmoneta/manifest.c
```diff
@@ -0,0 +1,295 @@
...
+         snprintf(file_path, MAX_PATH, "%s%s", root, file->path);
+      }
+      else
+      {
+         snprintf(file_path, MAX_PATH, "%s/%s", root, file->path);
+      }
+      file_size = pgmoneta_get_file_size(file_path);
+      if (file_size != file->size)
+      {
+         pgmoneta_log_error("File size mismatch: %s, getting %lu,should be %lu", file_size, file->size);
```
jesperpedersen, line 154: Missing space

src/libpgmoneta/manifest.c
```diff
@@ -0,0 +1,295 @@
...
+         tmp->next = manifest_file;
+         tmp = tmp->next;
+      }
+   }
+
+   ptr = strstr(json, "\"Manifest-Checksum\"");
+
+   // save preceding lines before the manifest checksum field,
+   // in order to verify the checksum of the manifest itself
+   content_size = ptr - json;
```
jesperpedersen, line 269: This seems hacky - isn't there a better way ?

We can't assume that the manifest format to be stable

jesperpedersen (changes requested):

jesperpedersen:
I don't see `MANIFEST yes MANIFEST_CHECKSUMS 'SHA256'` being sent for PostgreSQL >= 13

src/libpgmoneta/manifest.c
Jubilee101: Oh it looks hacky but is actually not, because the checksum of the manifest itself has to be calculated last. So it's always the last one to be added to the manifest json. According to https://www.postgresql.org/docs/current/backup-manifest-toplevel.html: "This key is always present on the last line of the backup manifest file. "

I could also just count backwards until getting '\n', but I don't know how stable the format is, for example is the feilds always separated by new line? Or will there always be a trailing '\n', so searching for this field directly is actually easier and more reliable I think. Let me know your thoughts.

Jubilee101:
@jesperpedersen I checked on Postgres15; the message sent is `BASE_BACKUP (LABEL 'pgmoneta_base_backup_20230809121918', WAL true, WAIT false, CHECKPOINT 'fast', MANIFEST 'yes', MANIFEST_CHECKSUMS 'SHA256')`. It won't appear on postgres log. 

I verified this also with Postgres13 as well earlier. And I'm sure the manifest is indeed received and verified.

The rest of change requests have addressed. We don't have hash algorithm functionalities other than sha256 yet. So I just left a log message as a placeholder there. We could open another issue to add support for other hash algorithms.

jesperpedersen:
Yeah, for now lets just push `MANIFEST yes MANIFEST_OPTION 'SHA256'` as well in our implementation

Jubilee101:
> Yeah, for now lets just push MANIFEST yes MANIFEST_OPTION 'SHA256' as well in our implementation

I'm not following.. These options is already in our backup infrastructure. And I already verified the message sent to postgres server. Can you be more specific on what needs to change?

jesperpedersen:
Merged.

PR #150, opened 2023-11-03
add build/installation instructions on different platforms

Jubilee101:
Add build and installation instructions on Fedora (37, 38), Rocky (8, 9) and RHEL 8. For RHEL 9, libev is not supported, we'll have to remove the dependency on that in the future, in order to support RHEL 9. 

The main differences are the ways to fulfill depency requirements on different platforms in order to build pgmoneta. Luckily building and installing pgmoneta is the same on all the current primarily supported platforms.


Jubilee101:
So turns out I got the wrong version of EPEL installed on RHEL 9... So there is no need to remove dependency on anything for now 😃.

jesperpedersen:
Rebased, and merged.

Issue #158, opened 2023-12-05
WAL: Extract error message


jesperpedersen:
Currently `wal.c` uses `pgmoneta_log_message` in case of an error.

The error message should be extracted instead and displayed using an ERROR log

Jubilee101:
Got it. It should be easy as the format is declared in Postgres' documents.

PR #165, opened 2023-12-11
[#158] extract and log CopyFail/ErrorResponse message

Jubilee101:
ErrorResponse message usually consists of multiple fields, each with a type and a string, referring to https://www.postgresql.org/docs/current/protocol-error-fields.html.

Aside from the fields that tell you what the error is, postgres also tells you things like where in the source code the error was reported. It would be useful for debugging but messy and confusing for users. So I only log the "always present" fields (message, severity and error code) in ERROR level and leave the rest to DEBUG level.

At INFO level above, we shall see:
9dfdd20e2013c62f85ab87e7250e62a

At DEBUG level, we'll see:
a17373aad779e86e42c6266312cd848

### DIFF
--- a/src/libpgmoneta/message.c
+++ b/src/libpgmoneta/message.c
@@ -149,6 +149,58 @@ pgmoneta_log_message(struct message* msg)
    }
 }
 
+void
+pgmoneta_log_copyfail_message(struct message* msg)
+{
+   if (msg == NULL || msg->kind != 'f')
+   {
+      return;
+   }
+
+   pgmoneta_log_error("COPY-failure: %s", (char*) msg->data);
+}
+
+void
+pgmoneta_log_error_response_message(struct message* msg)
+{
+   size_t offset = 0;
+   signed char field_type = 0;
+   char* error = NULL;
+   char* error_code = NULL;
+
+   if (msg == NULL || msg->kind != 'E')
+   {
+      return;
+   }
+
+   pgmoneta_extract_error_fields('M', msg, &error);
+   pgmoneta_extract_error_fields('C', msg, &error_code);
+
+   pgmoneta_log_error("error response message: %s (SQLSTATE code: %s)", error, error_code);
+
+   while (offset < msg->length)
+   {
+      field_type = pgmoneta_read_byte(msg->data + offset);
+
+      if (field_type == '\0')
+      {
+         break;
+      }
+
+      offset += 1;
+
+      if (field_type != 'M' && field_type != 'C')
+      {
+         pgmoneta_log_debug("error response field type: %c, message: %s", field_type, msg->data + offset);
+      }
+
+      offset += (strlen(msg->data + offset) + 1);
+   }
+
+   free(error_code);
+   free(error);
+}
+
 int
 pgmoneta_write_empty(SSL* ssl, int socket)
 {
@@ -1750,9 +1802,16 @@ pgmoneta_consume_data_row_messages(int socket, struct stream_buffer* buffer, str
          goto error;
       }
 
-      if (msg == NULL || msg->kind == 'E' || msg->kind == 'f')
+      if (msg == NULL)
       {
-         pgmoneta_log_message(msg);
+         pgmoneta_log_error("wal: received NULL message");
+         goto error;
+      }
+
+      if (msg->kind == 'E' || msg->kind == 'f')
+      {
+         pgmoneta_log_copyfail_message(msg);
+         pgmoneta_log_error_response_message(msg);
          goto error;
       }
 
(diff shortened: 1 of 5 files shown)


jesperpedersen:
I think we should only log one line per error, aka

```
error response message: requested WAL .... (SQLSTATE code: 58P01
```

Maybe we don't need the severity from the PostgreSQL side, since it is always an ERROR on our side.

Could be an idea to use `pgmoneta_extract_message()`

Jubilee101:
> I think we should only log one line per error, aka

Is it still ok to log multiple lines for debug log?

> Maybe we don't need the severity from the PostgreSQL side, since it is always an ERROR on our side.

That makes sense, do you want to keep it in debug level?

> Could be an idea to use pgmoneta_extract_message()

Nah the format is quite different. For error feilds there is no length info in it.

jesperpedersen:
Yeah, I think we can keep `DEBUG` as is, including severity.

Only the `ERROR` is different since it should only trigger one alert in the monitoring system

Jubilee101:
New error log sample:
3ec3afbcafec7753f1f1fd0b2c4a369

src/libpgmoneta/message.c
```diff
@@ -1752,7 +1804,9 @@ pgmoneta_consume_data_row_messages(int socket, struct stream_buffer* buffer, str
 
       if (msg == NULL || msg->kind == 'E' || msg->kind == 'f')
       {
-         pgmoneta_log_message(msg);
+         pgmoneta_log_error("received NULL or error message");
```
jesperpedersen, line 1807: Remove

src/libpgmoneta/message.c
```diff
@@ -1883,7 +1937,9 @@ pgmoneta_receive_archive_files(int socket, struct stream_buffer* buffer, char* b
          pgmoneta_consume_copy_stream(socket, buffer, &msg);
          if (msg->kind == 'E' || msg->kind == 'f')
          {
-            pgmoneta_log_message(msg);
+            pgmoneta_log_error("received NULL or error message");
```
jesperpedersen, line 1940: Remove

src/libpgmoneta/message.c
```diff
@@ -2026,7 +2084,9 @@ pgmoneta_receive_archive_stream(int socket, struct stream_buffer* buffer, char*
       pgmoneta_consume_copy_stream(socket, buffer, &msg);
       if (msg->kind == 'E' || msg->kind == 'f')
       {
-         pgmoneta_log_message(msg);
+         pgmoneta_log_error("received NULL or error message");
```
jesperpedersen, line 2087: Remove

src/libpgmoneta/message.c
```diff
@@ -2036,7 +2096,9 @@ pgmoneta_receive_archive_stream(int socket, struct stream_buffer* buffer, char*
...
```
jesperpedersen, line 2099: Remove

src/libpgmoneta/message.c
```diff
@@ -2288,7 +2352,9 @@ pgmoneta_receive_manifest_file(int socket, struct stream_buffer* buffer, char* b
...
```
jesperpedersen, line 2355: Remove

src/libpgmoneta/wal.c
```diff
@@ -199,7 +199,9 @@ pgmoneta_wal(int srv, char** argv)
 
       if (msg == NULL || msg->kind == 'E' || msg->kind == 'f')
       {
-         pgmoneta_log_message(msg);
+         pgmoneta_log_error("received NULL or error message");
```
jesperpedersen, line 202: Remove

src/libpgmoneta/message.c
```diff
@@ -149,6 +149,58 @@ pgmoneta_log_message(struct message* msg)
...
+
+void
+pgmoneta_log_error_response_message(struct message* msg)
+{
+   size_t offset = 0;
+   signed char field_type = 0;
+   char* error = NULL;
+   char* error_code = NULL;
+
+   if (msg == NULL || msg->kind != 'E')
```
jesperpedersen, line 171: Split this into two `if`s.

Log in case if `msg == NULL`

jesperpedersen (changes requested):

jesperpedersen:
We just need to make sure only one ERROR is logged, and we know about `NULL` and invalid messages as well

Jubilee101:
> We just need to make sure only one ERROR is logged, and we know about `NULL` and invalid messages as well

Got it. I kept that message simply because if we only log it within a specific logging function, we won't be able to know which line exactly that error happened.

jesperpedersen:
If we do

```
pgmoneta_log_copyfail_message(msg);
pgmoneta_log_error_response_message(msg);
```

then 2 ERRORs will show up - even though there is only 1 real error

jesperpedersen:
Merged.

Issue #157, opened 2023-12-05
Verify replication slots upon startup


jesperpedersen:
We should verify that `backup_slot` and `wal_slot` are valid on the server in question.

If not, then it is a FATAL and pgmoneta should be stopped.

jesperpedersen:
F.ex.

```
postgres=# SELECT slot_name, slot_type FROM pg_replication_slots;
 slot_name | slot_type 
-----------+-----------
 repl      | physical
(1 row)
```

Jubilee101:
Got it! I'll work on this. One problem we found constantly is missing WAL segments when there's no replication slot configured. Currently we do nothing when there is no replication slot configured. Should we make wal_slot a required configuration? Or can we create a replication slot automatically for user if there's none?

jesperpedersen:
I added #160 for the first part, and #161 for the latter

Jubilee101:
Got it!

PR #166, opened 2023-12-12
[#157] verify replication slot upon startup if it is configured

Jubilee101:
Hi @jesperpedersen , please note that this is still a draft because there's tiny problem there. pgmoneta is able to verify the replication slot and exit(1) if it's wrong. But it seems that the pid file is not removed. As a result, the next time it cannot startup even when the config is correct. Here's the log:
```
2023-12-11 19:44:09 FATAL main.c:1954 PID file [/tmp//pgmoneta.all.pid] exists, is there another instance running ?
```
I checked the pid inside /tmp/pgmoneta.all.pid, and seems that there's no corresponding process. Should I manually remove the file before exit, or is there anything I'm doing not quite right?

### DIFF
--- a/src/libpgmoneta/message.c
+++ b/src/libpgmoneta/message.c
@@ -913,6 +913,36 @@ pgmoneta_create_replication_slot_message(char* create_slot_name, struct message*
 }
 
 int
+pgmoneta_create_search_replication_slot_message(char* slot_name, struct message** msg)
+{
+   char cmd[1024];
+   struct message* m = NULL;
+   size_t size;
+
+   memset(&cmd[0], 0, sizeof(cmd));
+
+   snprintf(cmd, sizeof(cmd), "SELECT slot_name, slot_type FROM pg_replication_slots WHERE slot_name = '%s';", slot_name);
+
+   size = 1 + 4 + strlen(cmd) + 1;
+
+   m = (struct message*)malloc(sizeof(struct message));
+   m->data = malloc(size);
+
+   memset(m->data, 0, size);
+
+   m->kind = 'Q';
+   m->length = size;
+
+   pgmoneta_write_byte(m->data, 'Q');
+   pgmoneta_write_int32(m->data + 1, size - 1);
+   memcpy(m->data + 5, &cmd[0], strlen(cmd));
+
+   *msg = m;
+
+   return MESSAGE_STATUS_OK;
+}
+
+int
 pgmoneta_send_copy_done_message(int socket)
 {
    struct message* msg = NULL;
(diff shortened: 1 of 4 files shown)


jesperpedersen:
The PID file is remove here - https://github.com/pgmoneta/pgmoneta/blob/main/src/main.c#L678

You need to look at how much of pgmoneta has been initialized at this point and instead of a direct `exit(1)` set an `exit_code` and do a `goto` instead.

Also the logic around verification needs a bit of work - if a WAL slot is defined, then the search message for a server can say that it is defined (no further action), or isn't defined; in that case is `create_slot` defined - run it. If `create_slot` isn't defined then it is a `FATAL`.

src/libpgmoneta/message.c
```diff
@@ -912,6 +912,36 @@ pgmoneta_create_replication_slot_message(char* create_slot_name, struct message*
...
+int
+pgmoneta_create_search_replication_slot_message(char* slot_name, struct message** msg)
+{
+   char cmd[1024];
+   struct message* m = NULL;
+   size_t size;
+
+   memset(&cmd[0], 0, sizeof(cmd));
+
+   snprintf(cmd, sizeof(cmd), "SELECT slot_name, slot_type FROM pg_replication_slots WHERE slot_name = '%s';", slot_name);
```
jesperpedersen, line 924: We don't need the `slot_type` since we already know it will be a `physical` one

jesperpedersen (changes requested):

Jubilee101:
> You need to look at how much of pgmoneta has been initialized at this point and instead of a direct exit(1) set an exit_code and do a goto instead.

That's where it confuses me. I saw a lot of direct exit(1) after pid file is created and before it is destroyed. I thought it'll be ok to just exit.

> Also the logic around verification needs a bit of work - if a WAL slot is defined, then the search message for a server can say that it is defined (no further action), or isn't defined; in that case is create_slot defined - run it. If create_slot isn't defined then it is a FATAL.

Got it. It is a little arbitrary right now. I'll fix the rule.

src/libpgmoneta/message.c
Jubilee101: What if the user configured the wrong slot type?

jesperpedersen:
Yeah, we should fix that now, and guard the key variables with `NULL` checks in case they havn't been initialized.

src/libpgmoneta/message.c
jesperpedersen: Sure, we can add that check - ideally users should use `create_slot = yes` to control slots

jesperpedersen:
Rebased, and merged.

Issue #168, opened 2023-12-13
Segfault in `delete_wal_older_than`


johnaohara:
**Describe the bug**

A segfault occurs when pgmoneta attempts to delete historic WALs

```
Dec 13 01:25:00 server02.example.com kernel: pgmoneta[3527739]: segfault at 5565a6579 ip 00007f125d869e41 sp 00007ffdaab1de90 error 4 in libc-2.28.so[7f125d7ce000+1bb000]
Dec 13 01:25:00 server02.example.com kernel: Code: 2e 0f 1f 84 00 00 00 00 00 f3 0f 1e fa 53 48 83 ec 10 48 8b 05 c8 30 32 00 48 8b 00 48 85 c0 0f 85 84 00 00 00 48 85 ff 74 6f <48> 8b 47 f8 48 8d 77 f0 a8 02 75 >
Dec 13 01:25:00 server02.example.com systemd[1]: Starting Check pmlogger instances are running...
Dec 13 01:25:00 server02.example.com systemd[1]: Started Process Core Dump (PID 3527741/UID 0).
Dec 13 01:25:00 server02.example.com systemd[1]: Started Check pmlogger instances are running.
Dec 13 01:25:00 server02.example.com systemd-coredump[3527743]: Process 3527739 (pgmoneta) of user 1001 dumped core.
                                                                                        
                                                                                        Stack trace of thread 3527739:
                                                                                        #0  0x00007f125d869e41 __libc_free (libc.so.6)
                                                                                        #1  0x00007f125dba817a delete_wal_older_than (libpgmoneta.so.0)
                                                                                        #2  0x00007f125dba83cb pgmoneta_delete_wal (libpgmoneta.so.0)
                                                                                        #3  0x00007f125dbcd0c4 retain_execute (libpgmoneta.so.0)
                                                                                        #4  0x00007f125dbbae7d pgmoneta_retention (libpgmoneta.so.0)
                                                                                        #5  0x00007f125f477f63 ev_invoke_pending (libev.so.4)
                                                                                        #6  0x00007f125f47bff8 ev_run (libev.so.4)
                                                                                        #7  0x00005565a601f55e ev_loop (pgmoneta)
                                                                                        #8  0x00007f125d808d85 __libc_start_main (libc.so.6)
                                                                                        #9  0x00005565a601ff6e _start (pgmoneta)
```

**To Reproduce**

Steps to reproduce the behavior.

**Version**

```
$ pgmoneta -V
[...]

PR #169, opened 2023-12-13
[#168] delect_wal_older_than should return before entering error section

Jubilee101:


jesperpedersen:
Merged.

Issue #164, opened 2023-12-08
Decrease memory allocation rate


Jubilee101:
Valgrind reports our base backup mallocs too many bytes. We need to go over base backup implementation and reuse some buffer, to decrease the number of bytes allocated.

Jubilee101:
Heaptrack reports that most allocations come from untar, cjson parsing and extracting messages from the copy stream. We'll try to decrease the memory allocated for libarchive and cjson, and reuse one message struct for extraction. We will also try to use the stream buffer directly to hold the payload.

PR #176, opened 2024-01-02
[#164] Reuse message buffer to hold consumed messages

Jubilee101:
Sorry for the delay.. I tested it out and seems that after replacing the old pgmoneta_cosume_copy_stream, both wal and base backup are working fine. And we reduced a lot of mallocs:

62cf36c3f2c15a61a5aaf07745e37c1


jesperpedersen:
Merged.

PR #178, opened 2024-01-07
set message buffer payload to NULL before freeing

Jubilee101:
Since our message buffer is just pointing to the middle of some malloced space, there's no point of freeing the payload. We need to set it to NULL or it'll cause invalid free.


jesperpedersen:
Merged with a small style change.

Issue #187, opened 2024-02-15
Timeline/Backup LSN Support


Jubilee101:
Currently pgmoneta has little support for fetching/extracting LSN info related to timeline and backup. We would like the following feature:

- [x] .history file streaming
- [x] .history file parsing 
- [x] Corresponding prometheus support
- [x] Make sure pgmoneta starts fetching WAL segments of the next timeline when the current one ends
- [x] Extracting starting LSN from backup_label into backup.info

Some references: 
https://www.highgo.ca/2021/11/01/the-postgresql-timeline-concept/
https://developer.aliyun.com/article/234
https://www.postgresql.org/docs/current/protocol-replication.html

Jubilee101:
Close this as most tasks have been done. I'll later submit a separate issue for server side LSN. That'll tie into the extension project, and require looking into postgres internals - base backup command as well as mvcc stuff.

PR #188, opened 2024-02-19
[#187] Make WAL streaming follow timeline change

Jubilee101:
Hi,

This is a big improvement on our WAL functionality. Previously we get our WAL streaming starting position using IDENTIFY_SYSTEM. However in postgres this is actually the last resort. The reason is that, consider the case when server goes into recovery mode and bumps its timeline from 1 to 2. When pgmoneta connects to server again, it will start streaming segments on timeline 2 directly, instead of what's remaining on timeline 1. This is because IDENTIFY_SYSTEM always gives us the current timeline and xlog position. And the server is currently on timeline 2. 

In the new design, on streaming starts, pgmoneta will go into WAL archive directory and starts looking for the latest segment, which it will use as the starting position, disregarding which timeline it's on. And when server finished streaming WAL segments on an old timeline, it will send command complete and indicate the next xlogpos on the next timeline. We also added support to read that info and automatically start the next WAL streaming process on the next timeline:
23113e1cfecc2c69c78e28017d17025

Some minor bugs are also fixed along the way, including our error response logging, sometimes the error response message's payload does not include type and length header, causing parsing error. As well as our create_slot functionality, create_slot flag was permanently set to false. More config->running checkings are added to prevent the child process not terminated properly.

Note that pgmoneta restore currently doesn't support having the server going into recovery mode. To make the server jump to the next timeline, I need to manually add recovery.signal, and change restore_command in postgresql.conf.

### DIFF
--- a/src/libpgmoneta/wal.c
+++ b/src/libpgmoneta/wal.c
@@ -40,8 +40,10 @@
 #include <utils.h>
 
 /* system */
+#include <dirent.h>
 #include <errno.h>
 #include <ev.h>
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <unistd.h>
@@ -49,13 +51,15 @@
 #include <sys/types.h>
 #include <openssl/ssl.h>
 
-static char* wal_file_name(int timeline, size_t segno, int segsize);
+static char* wal_file_name(uint32_t timeline, size_t segno, int segsize);
+static int wal_fetch_history(char* basedir, int timeline, SSL* ssl, int socket);
 static FILE* wal_open(char* root, char* filename, int segsize);
 static int wal_close(char* root, char* filename, bool partial, FILE* file);
 static int wal_prepare(FILE* file, int segsize);
 static int wal_send_status_report(SSL* ssl, int socket, int64_t received, int64_t flushed, int64_t applied);
 static int wal_xlog_offset(size_t xlogptr, int segsize);
-static int wal_convert_xlogpos(char* xlogpos, int* high32, int* low32, int segsize);
+static int wal_convert_xlogpos(char* xlogpos, uint32_t* high32, uint32_t* low32, int segsize);
+static int wal_find_streaming_start(char* basedir, uint32_t* timeline, uint32_t* high32, uint32_t* low32, int segsize);
 static int wal_shipping_setup(int srv, char** wal_shipping);
 
 void
@@ -63,14 +67,14 @@ pgmoneta_wal(int srv, char** argv)
 {
    int usr;
    int auth;
-   int cnt = 0;
    SSL* ssl = NULL;
    int socket = -1;
-   int high32 = 0;
-   int low32 = 0;
+   uint32_t high32 = 0;
+   uint32_t low32 = 0;
    char* d = NULL;
    char* wal_shipping = NULL;
-   int timeline = -1;
+   uint32_t timeline = 0;
+   uint32_t cur_timeline = 0;
    int hdrlen = 1 + 8 + 8 + 8;
    int bytes_left = 0;
    char* xlogpos = NULL;
@@ -90,6 +94,7 @@ pgmoneta_wal(int srv, char** argv)
    FILE* wal_shipping_file = NULL;
    struct message* identify_system_msg = NULL;
    struct query_response* identify_system_response = NULL;
+   struct query_response* end_of_timeline_response = NULL;
    struct message* start_replication_msg = NULL;
    struct message* msg = (struct message*)malloc(sizeof (struct message));
    struct configuration* config;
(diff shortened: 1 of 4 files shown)


jesperpedersen:
Merged.

src/libpgmoneta/wal.c
```diff
@@ -649,10 +763,96 @@ wal_convert_xlogpos(char* xlogpos, int* high32, int* low32, int segsize)
...
+   sscanf(pos, "%08X%08X%08X", timeline, high32, low32);
+
+   // high32 is the segment id, low32 is the number of segments
+   int segments_per_id = 0x100000000ULL / segsize;
+   // if the latest wal segment is partial, we start with this one
+   // otherwise we start with the next one
+   if (!high_is_partial)
+   {
+      // handle possible overflow
+      if (*low32 == segments_per_id)
```
ashu3103, line 833: Hi @Jubilee101, 

Here `*low32` is the last 32 bits (8 Hex characters) of the wal file name (in backup base_dir).

At this line we are limiting the value of `*low32` to `segments_per_id` and assuming `segsize = 16MB`,  `segments_per_id = 256`.

Which means for the wal segment -

`pos = 000000010000000100000100`

`*low32` will be `0` instead of `257`
and
`*high32` will be `2` instead of `1`.

Please correct me if I'm in wrong direction

Jubilee101: Hey,

Yes, this is a bit like the page table. The WAL address is 64 bits which is byte addressing. The higher 32 bits are group addressing, and lower 32 bits are in-group offset. So the address space for each group is at most 4GB, with each xlogpos pointing to a byte within. And since by default the segsize is 16MB, the in-segment offset is 24 bits, leaving 8 bits for segment addressing. That's why each group you have 256 segments. 

Say you already have 256 segments (each segment is a WAL file in the directory) in group 1, and you add a new one. It carries to the next group. So it's the 0th segment in group 2. Note that the group is a virtual concept, you don't actually group the segments into separate directories.

Back to your question, it's almost right, but not quite accurate:

> *low32 will be 0 instead of 257

It's actually just the 8 bits between 24 - 31 that will be 0. Since 0 - 23 is in-segment offset.

ashu3103: Ahh, it all makes sense now, so this will get us the 8 MSB of the lower 32 bits of wal address and the rest 24 bits (segment offset) we will get from lsn (every thing after `/`).

Just a bit curious, how things would change if `wal_segment_size = 64MB` or greater because in that case we can't be having a 24 bit offset.

Thanks a lot!

Jubilee101: > Ahh, it all makes sense now, so this will get us the 8 MSB of the lower 32 bits of wal address and the rest 24 bits (segment offset) we will get from lsn (every thing after /).

If an xlogpos or lsn is 1/5D7B7530, it points to the byte inside the 5D segment in group 1 at the offset 7B7530.

> Just a bit curious, how things would change if wal_segment_size = 64MB or greater because in that case we can't be having a 24 bit offset.

It won't change much ideally. We calculate the number of segments dynamically, there will be more segments and you don't need that many in-segment offest. That said, I don't know if you can configure it to be smaller, I think most of the time you make it bigger (if you do it at all, no clue why ppl change it). The bottom line is the size has to be an exponent of 2.

ashu3103: Crystal clear, Thanks!

ashu3103: One more doubt @Jubilee101, 

In the code sequence -

```
wal_find_streaming_start(d, segsize, &timeline, &high32, &low32);
   pgmoneta_log_info("wal_find_streaming_start: timeline: %u, low: %u, high: %u", timeline, low32, high32);
   if (timeline == 0)
   {
      pgmoneta_log_info("xlogpos: %s", pgmoneta_query_response_get_data(identify_system_response, 2));
      read_replication = (config->servers[srv].version >= 15) ? 1 : 0;

      // query the replication slot to get the starting LSN and timeline ID
      if (read_replication)
      {
         if (wal_read_replication_slot(ssl, socket, config->servers[srv].wal_slot, config->servers[srv].name, segsize, &high32, &low32, &timeline))
         {
            read_replication = 0;   // Fallback if not PostgreSQL 15+
         }
      }

      // use current xlogpos as last resort
      if (!read_replication)
      {
         timeline = cur_timeline;
         xlogpos_size = strlen(pgmoneta_query_response_get_data(identify_system_response, 2)) + 1;
         xlogpos = (char*)malloc(xlogpos_size);

         if (xlogpos == NULL)
         {
            goto error;
         }
         memset(xlogpos, 0, xlogpos_size);
         memcpy(xlogpos, pgmoneta_query_response_get_data(identify_system_response, 2), xlogpos_size);
         if (wal_convert_xlogpos(xlogpos, segsize, &high32, &low32))
         {
            goto error;
         }
         free(xlogpos);
         xlogpos = NULL;
      }
      pgmoneta_log_info("timeline=0 if: timeline: %u, low: %u, high: %u", timeline, low32, high32);
   }

```

If `READ REPLICATION` is successfull we will set `low32` and `high32` as per `restart_lsn`. But here suppose `restart_lsn = 0/5D7B7530`, the low32 would be set to `5D000000` (as per `wal_convert_xlogpos`). Isn't it wrong or redundant as any lsn between `5D000000` and `restart_lsn` in `5D` segment is already been replicated (as per the definition of restart_lsn which goes like this - **The address (LSN) of oldest WAL which still might be required by the consumer of this slot and thus won't be automatically removed during checkpoints unless this LSN gets behind more than [max_slot_wal_keep_size](https://www.postgresql.org/docs/current/runtime-config-replication.html#GUC-MAX-SLOT-WAL-KEEP-SIZE) from the current LSN. NULL if the LSN of this slot has never been reserved.** - this is equivalent to saying that every lsn older than `restart_lsn` is consumed by the replication consumer and therefore replicated), so low32 should be exactly equal to `restart_lsn` (according to me).

Jubilee101: If something was partially received then next time the streaming starts it streams the file from the beginning again. This is mostly for simplicity I think, so that server won't bother openning the file and pointing to some arbitrary position for you. And as long as the file is partially sent the entire segment is kept by replication slot, so the data won't be gone even if you have claimed receiving it.

PR #199, opened 2024-03-02
[#187] Parse .history and expose to prometheus

Jubilee101:
Hi,

Sorry about the delay. This has been a crazy week 😩 

This patch provides us the ability to parse the history file. To make my life a little easier, I excluded .history file from being compressed - they are tiny anyway.

Prometheus is not making my life easy either, I cannot display hex string, nor put high and low 32 bits together, because it's A STRING!! So I display them separately in decimal for now. By default timeline 1 will have an invalid parent timeline of 0 and a switch point of 0/0.

Also spent great deal of time trying to connect local prometheus client to pgmoneta on Azure server. Turns out there's a firewall that overwrote Azure's inbound rules :(

Let me know if there's problem~

### DIFF
--- a/src/libpgmoneta/prometheus.c
+++ b/src/libpgmoneta/prometheus.c
@@ -36,9 +36,11 @@
 #include <prometheus.h>
 #include <shmem.h>
 #include <utils.h>
+#include <wal.h>
 
 /* system */
 #include <ev.h>
+#include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
@@ -325,6 +327,51 @@ home_page(int client_fd)
    data = pgmoneta_append(data, "  <p>\n");
    data = pgmoneta_append(data, "  <h2>pgmoneta_wal_shipping_total_space</h2>\n");
    data = pgmoneta_append(data, "  The total disk space for the WAL shipping directory of a server\n");
+   data = pgmoneta_append(data, "  <p>\n");
+   data = pgmoneta_append(data, "  <h2>pgmoneta_server_timeline</h2>\n");
+   data = pgmoneta_append(data, "  The current timeline a server is on\n");
+   data = pgmoneta_append(data, "  <table border=\"1\">\n");
+   data = pgmoneta_append(data, "    <tbody>\n");
+   data = pgmoneta_append(data, "      <tr>\n");
+   data = pgmoneta_append(data, "        <td>name</td>\n");
+   data = pgmoneta_append(data, "        <td>The identifier for the server</td>\n");
+   data = pgmoneta_append(data, "      </tr>\n");
+   data = pgmoneta_append(data, "    </tbody>\n");
+   data = pgmoneta_append(data, "  </table>\n");
+   data = pgmoneta_append(data, "  <p>\n");
+   data = pgmoneta_append(data, "  <h2>pgmoneta_server_parent_tli</h2>\n");
+   data = pgmoneta_append(data, "  The parent timeline of a timeline on a server\n");
+   data = pgmoneta_append(data, "  <table border=\"1\">\n");
+   data = pgmoneta_append(data, "    <tbody>\n");
+   data = pgmoneta_append(data, "      <tr>\n");
+   data = pgmoneta_append(data, "        <td>name</td>\n");
+   data = pgmoneta_append(data, "        <td>The identifier for the server</td>\n");
+   data = pgmoneta_append(data, "      </tr>\n");
+   data = pgmoneta_append(data, "      <tr>\n");
+   data = pgmoneta_append(data, "        <td>tli</td>\n");
+   data = pgmoneta_append(data, "        <td>The current/previous timeline ID in the server history</td>\n");
+   data = pgmoneta_append(data, "      </tr>\n");
+   data = pgmoneta_append(data, "    </tbody>\n");
+   data = pgmoneta_append(data, "  </table>\n");
+   data = pgmoneta_append(data, "  <p>\n");
+   data = pgmoneta_append(data, "  <h2>pgmoneta_server_timeline_switchpos</h2>\n");
+   data = pgmoneta_append(data, "  The WAL switch position of a timeline on a server (showed in hex as a parameter)\n");
+   data = pgmoneta_append(data, "  <table border=\"1\">\n");
+   data = pgmoneta_append(data, "    <tbody>\n");
+   data = pgmoneta_append(data, "      <tr>\n");
+   data = pgmoneta_append(data, "        <td>name</td>\n");
+   data = pgmoneta_append(data, "        <td>The identifier for the server</td>\n");
+   data = pgmoneta_append(data, "      </tr>\n");
+   data = pgmoneta_append(data, "      <tr>\n");
+   data = pgmoneta_append(data, "        <td>tli</td>\n");
+   data = pgmoneta_append(data, "        <td>The current/previous timeline ID in the server history</td>\n");
+   data = pgmoneta_append(data, "      </tr>\n");
+   data = pgmoneta_append(data, "      <tr>\n");
+   data = pgmoneta_append(data, "        <td>walpos</td>\n");
+   data = pgmoneta_append(data, "        <td>The WAL switch position of this timeline</td>\n");
+   data = pgmoneta_append(data, "      </tr>\n");
+   data = pgmoneta_append(data, "    </tbody>\n");
+   data = pgmoneta_append(data, "  </table>\n");
    data = pgmoneta_append(data, "  <p>\n");
    data = pgmoneta_append(data, "  <h2>pgmoneta_backup_oldest</h2>\n");
    data = pgmoneta_append(data, "  The oldest backup for a server\n");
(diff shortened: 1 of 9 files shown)


src/libpgmoneta/prometheus.c
```diff
@@ -326,6 +327,62 @@ home_page(int client_fd)
...
+   data = pgmoneta_append(data, "  The high 32 bits in decimal of the switch position of a timeline on a server\n");
+   data = pgmoneta_append(data, "  <table border=\"1\">\n");
+   data = pgmoneta_append(data, "    <tbody>\n");
+   data = pgmoneta_append(data, "      <tr>\n");
+   data = pgmoneta_append(data, "        <td>name</td>\n");
+   data = pgmoneta_append(data, "        <td>The identifier for the server</td>\n");
+   data = pgmoneta_append(data, "      </tr>\n");
+   data = pgmoneta_append(data, "      <tr>\n");
+   data = pgmoneta_append(data, "        <td>tli</td>\n");
+   data = pgmoneta_append(data, "        <td>The current/previous timeline ID for the server</td>\n");
```
jesperpedersen, line 366: Missing `high32`

src/libpgmoneta/prometheus.c
```diff
@@ -326,6 +327,62 @@ home_page(int client_fd)
...
+   data = pgmoneta_append(data, "  The low 32 bits in decimal of the switch position of a timeline on a server\n");
+   data = pgmoneta_append(data, "  <table border=\"1\">\n");
+   data = pgmoneta_append(data, "    <tbody>\n");
+   data = pgmoneta_append(data, "      <tr>\n");
+   data = pgmoneta_append(data, "        <td>name</td>\n");
+   data = pgmoneta_append(data, "        <td>The identifier for the server</td>\n");
+   data = pgmoneta_append(data, "      </tr>\n");
+   data = pgmoneta_append(data, "      <tr>\n");
+   data = pgmoneta_append(data, "        <td>tli</td>\n");
+   data = pgmoneta_append(data, "        <td>The current/previous timeline ID for the server</td>\n");
```
jesperpedersen, line 381: Missing `low32`

src/libpgmoneta/prometheus.c
```diff
@@ -947,6 +1004,163 @@ general_information(int client_fd)
    }
    data = pgmoneta_append(data, "\n");
 
+   data = pgmoneta_append(data, "#HELP pgmoneta_server_timeline The current timeline a server is on\n");
+   data = pgmoneta_append(data, "#TYPE pgmoneta_server_timeline\n");
```
jesperpedersen, line 1008: Missing `counter`

src/libpgmoneta/prometheus.c
```diff
@@ -947,6 +1004,163 @@ general_information(int client_fd)
...
+      data = pgmoneta_append(data, "\"} ");
+
+      data = pgmoneta_append_int(data, config->servers[i].cur_timeline);
+
+      data = pgmoneta_append(data, "\n");
+   }
+   data = pgmoneta_append(data, "\n");
+
+   data = pgmoneta_append(data, "#HELP pgmoneta_server_parent_tli The parent timeline of a timeline on a server\n");
+   data = pgmoneta_append(data, "#TYPE pgmoneta_server_parent_tli\n");
```
jesperpedersen, line 1024: Missing `counter`

src/libpgmoneta/prometheus.c
```diff
@@ -947,6 +1004,163 @@ general_information(int client_fd)
...
+
+         curh = curh->next;
+         tli++;
+      }
+      pgmoneta_free_timeline_history(history);
+   }
+   data = pgmoneta_append(data, "\n");
+
+   data = pgmoneta_append(data, "#HELP pgmoneta_server_timeline_switchpos_high32 The high 32 bits in decimal of the switch position of a timeline on a serverr\n");
+   data = pgmoneta_append(data, "#TYPE pgmoneta_server_timeline_switchpos_high32\n");
```
jesperpedersen, line 1071: Missing `gauge`

src/libpgmoneta/prometheus.c
```diff
@@ -947,6 +1004,163 @@ general_information(int client_fd)
...
+
+         curh = curh->next;
+         tli++;
+      }
+      pgmoneta_free_timeline_history(history);
+   }
+   data = pgmoneta_append(data, "\n");
+
+   data = pgmoneta_append(data, "#HELP pgmoneta_server_timeline_switchpos_low32 The low 32 bits in decimal of the switch position of a timeline on a serverr\n");
+   data = pgmoneta_append(data, "#TYPE pgmoneta_server_timeline_switchpos_low32\n");
```
jesperpedersen, line 1118: Missing `gauge`

jesperpedersen (changes requested):

jesperpedersen:
So, putting it in a parameter didn't work for you - e.g. `{... switch="high32low32"} =1` ?

jesperpedersen:
Maybe https://github.com/jesperpedersen/podman-pgsql/tree/master/pgsql14-grafana-rocky8 can help...

Jubilee101:
> So, putting it in a parameter didn't work for you - e.g. {... switch="high32low32"} 1 ?

Oh yeah I see what you mean. Yeah it'll be better. I was a little dizzy after all the firewall debugging. Changes coming right up.

> Maybe https://github.com/jesperpedersen/podman-pgsql/tree/master/pgsql14-grafana-rocky8 can help...

Thanks! It's the firewall on the Fedora39. It's blocking every port except 22. I had to disable it for testing

src/libpgmoneta/prometheus.c
Jubilee101: I didn't know there's a counter type, still a lot to learn I guess.

Jubilee101: Though as for parent TLI I think it's not garanteed to be increase monotonically, which is its main difference between `guage`. You can always restore from an way earlier tli.

jesperpedersen:
Wouldn't it be better to have the whole thing in a parameter, and the value as 1... ?

Have to think a little about that one... having the whole as one, or using two metrics to figure it out - but having the values "at the front"...

Jubilee101:
> Wouldn't it be better to have the whole thing in a parameter, and the value as 1... ?

Are you thinking about something like pgmoneta_server_timeline_switchpos{server="primary",tli="1", "xlogpos"="1/8E356E38"} 1?

I thought about this too.. I feel it's a little hacking, plus I don't know how it'll turn out when plugged into grafana. But prometheus guys actually suggested that, in some sense, see: https://github.com/prometheus/prometheus/issues/2227

jesperpedersen:
Yeah, I think it is best to just have it as parameters... like you describe above.

You can extract the information, and decode it if necessary...

So, lets go with your suggestion

src/libpgmoneta/prometheus.c
jesperpedersen: True, use `gauge`

Jubilee101:
Voila~
ad2a491cee798ba2279f0be5ff3280f

Internally the xlogpos is still kept as high32 and low32 in uint32 separately, it might be more helpful if we want to use it for other purposes later on.

src/libpgmoneta/prometheus.c
```diff
@@ -326,6 +328,51 @@ home_page(int client_fd)
...
+   data = pgmoneta_append(data, "  The parent timeline of a timeline on a server\n");
+   data = pgmoneta_append(data, "  <table border=\"1\">\n");
+   data = pgmoneta_append(data, "    <tbody>\n");
+   data = pgmoneta_append(data, "      <tr>\n");
+   data = pgmoneta_append(data, "        <td>name</td>\n");
+   data = pgmoneta_append(data, "        <td>The identifier for the server</td>\n");
+   data = pgmoneta_append(data, "      </tr>\n");
+   data = pgmoneta_append(data, "      <tr>\n");
+   data = pgmoneta_append(data, "        <td>tli</td>\n");
+   data = pgmoneta_append(data, "        <td>The current/previous timeline ID for the server</td>\n");
```
jesperpedersen, line 352: Previous ?

src/libpgmoneta/prometheus.c
```diff
@@ -326,6 +328,51 @@ home_page(int client_fd)
...
+   data = pgmoneta_append(data, "  The xlog switch position in decimal of a timeline on a server\n");
+   data = pgmoneta_append(data, "  <table border=\"1\">\n");
+   data = pgmoneta_append(data, "    <tbody>\n");
+   data = pgmoneta_append(data, "      <tr>\n");
+   data = pgmoneta_append(data, "        <td>name</td>\n");
+   data = pgmoneta_append(data, "        <td>The identifier for the server</td>\n");
+   data = pgmoneta_append(data, "      </tr>\n");
+   data = pgmoneta_append(data, "      <tr>\n");
+   data = pgmoneta_append(data, "        <td>tli</td>\n");
+   data = pgmoneta_append(data, "        <td>The current/previous timeline ID for the server</td>\n");
```
jesperpedersen, line 367: It should always be `'current`, right ?

src/libpgmoneta/wal.c
```diff
@@ -504,6 +506,99 @@ pgmoneta_wal(int srv, char** argv)
...
+      return 0;
+   }
+
+   snprintf(filename, sizeof(filename), "%08X.history", tli);
+   path = pgmoneta_get_server_wal(srv);
+   path = pgmoneta_append(path, filename);
+   file = fopen(path, "r");
+   if (file == NULL)
+   {
+      pgmoneta_log_error("unable to open history file: %s", strerror(errno));
```
jesperpedersen, line 532: Capital `U`

jesperpedersen (changes requested):

jesperpedersen:
I wonder if all "public" facing things should be called `WAL` instead of `xlog`...

src/libpgmoneta/prometheus.c
Jubilee101: Yeah, it not only shows the parent timeline of current timeline, but also the parent timeline of previous timeline, tli is the id of the timeline whose parent we want to know about.

src/libpgmoneta/prometheus.c
Jubilee101: It depends, we also know the parents of previous timelines, i.e. what timeline each timeline branched off from, so I showed them all... If you feel `current` is enough I can change that.

Jubilee101:
> I wonder if all "public" facing things should be called WAL instead of xlog...

Yeah, good point. I'm just being lazy and use them interchangealy. I'll fix that.

src/libpgmoneta/prometheus.c
Jubilee101: Something like:
image

Jubilee101:
image

src/libpgmoneta/prometheus.c
```diff
@@ -947,6 +994,124 @@ general_information(int client_fd)
...
+
+      data = pgmoneta_append(data, "name=\"");
+      data = pgmoneta_append(data, config->servers[i].name);
+      data = pgmoneta_append(data, "\", ");
+
+      data = pgmoneta_append(data, "tli=\"");
+      data = pgmoneta_append_int(data, 1);
+      data = pgmoneta_append(data, "\"} ");
+
+      data = pgmoneta_append_int(data, 0);
```
jesperpedersen, line 1031: We should probably do `1` here

src/libpgmoneta/prometheus.c
```diff
@@ -947,6 +994,124 @@ general_information(int client_fd)
...
+
+         data = pgmoneta_append(data, "name=\"");
+         data = pgmoneta_append(data, config->servers[i].name);
+         data = pgmoneta_append(data, "\", ");
+
+         data = pgmoneta_append(data, "tli=\"");
+         data = pgmoneta_append_int(data, tli);
+         data = pgmoneta_append(data, "\", ");
+
+         data = pgmoneta_append(data, "WALpos=\"");
```
jesperpedersen, line 1100: Need to be lowercase

jesperpedersen (changes requested):

src/libpgmoneta/prometheus.c
Jubilee101: Hmm, 0 means invalid tli, since timeline starts with 1. I think it's a good way to say timeline 1 doesn't have a parent.

jesperpedersen: Ok

jesperpedersen:
Merged.

Issue #207, opened 2024-03-11
Expose WAL log sequence number (pgmoneta_current_wal_lsn) to Prometheus


lenhat509:
Display the current Log Sequence Number of the WAL on Prometheus metrics

```
pgmoneta_current_wal_1sn{server="primary", LSN="0/19FCACO"} 1
 ```

**Comment by jesperpedersen** [comment]
It should be

```
pgmoneta_current_wal_lsn{server="primary", lsn="0/19FCACO"} 1
```

jesperpedersen:
If LSN is unknown then it should say `lsn="0/0"`

lenhat509:
> It should be
> 
> ```
> pgmoneta_current_wal_lsn{server="primary", lsn="0/19FCACO"} 1
> ```
I fixed the comment. Thanks.

PR #208, opened 2024-03-11
[#207] Expose WAL log sequence number (pgmoneta_current_wal_lsn) to Prometheus

lenhat509:
My steps for implementing a pgmoneta_current_wal_lsn are:
1. Create a new attribute pgmoneta_current_wal_lsn, which will be used to store xlogpos, in the struct server. (in pgmoneta.h file)
2. Every time xlogpos gets updated by pgmoneta_query_response_get_data function, its value will also get stored in pgmoneta_current_wal_lsn (in wal.c)
3. Append the WAL LSN of every server to Prometheus (in prometheus.c)
4. In the copy_server function, allow pgmoneta_current_wal_lsn to be copied from one server to another. (in configuration.c)

### DIFF
--- a/src/include/pgmoneta.h
+++ b/src/include/pgmoneta.h
@@ -197,6 +197,7 @@ struct server
    char username[MAX_USERNAME_LENGTH]; /**< The user name */
    char wal_slot[MISC_LENGTH];         /**< The WAL slot name */
    char current_wal_filename[MISC_LENGTH]; /**< The current WAL filename*/
+   char current_wal_lsn[MISC_LENGTH]; /**< The current WAL log sequence number*/
    char follow[MISC_LENGTH];           /**< Follow a server */
    int retention_days;                 /**< The retention days for the server */
    int retention_weeks;                /**< The retention weeks for the server */
--- a/src/libpgmoneta/prometheus.c
+++ b/src/libpgmoneta/prometheus.c
@@ -615,6 +615,21 @@ home_page(int client_fd)
    data = pgmoneta_append(data, "    </tbody>\n");
    data = pgmoneta_append(data, "  </table>\n");
    data = pgmoneta_append(data, "  <p>\n");
+   data = pgmoneta_append(data, "  <h2>pgmoneta_current_wal_lsn</h2>\n");
+   data = pgmoneta_append(data, "  The current WAL log sequence number\n");
+   data = pgmoneta_append(data, "  <table border=\"1\">\n");
+   data = pgmoneta_append(data, "    <tbody>\n");
+   data = pgmoneta_append(data, "      <tr>\n");
+   data = pgmoneta_append(data, "        <td>server</td>\n");
+   data = pgmoneta_append(data, "        <td>The identifier for the server</td>\n");
+   data = pgmoneta_append(data, "      </tr>\n");
+   data = pgmoneta_append(data, "      <tr>\n");
+   data = pgmoneta_append(data, "        <td>lsn</td>\n");
+   data = pgmoneta_append(data, "        <td>The current WAL log sequence number</td>\n");
+   data = pgmoneta_append(data, "      </tr>\n");
+   data = pgmoneta_append(data, "    </tbody>\n");
+   data = pgmoneta_append(data, "  </table>\n");
+   data = pgmoneta_append(data, "  <p>\n");
    data = pgmoneta_append(data, "  <a href=\"https://pgmoneta.github.io/\">pgmoneta.github.io/</a>\n");
    data = pgmoneta_append(data, "</body>\n");
    data = pgmoneta_append(data, "</html>\n");
@@ -2369,6 +2384,34 @@ size_information(int client_fd)
 
       data = pgmoneta_append(data, "file=\"");
       data = pgmoneta_append(data, config->servers[i].current_wal_filename);
+      data = pgmoneta_append(data, "\"} ");
+
+      data = pgmoneta_append_bool(data, config->servers[i].wal_streaming);
+
+      data = pgmoneta_append(data, "\n");
+   }
+   data = pgmoneta_append(data, "\n");
+
+   // Append the WAL LSN of every server
+   data = pgmoneta_append(data, "#HELP pgmoneta_current_wal_lsn The current WAL log sequence number\n");
+   data = pgmoneta_append(data, "#TYPE pgmoneta_current_wal_lsn gauge\n");
+   for (int i = 0; i < config->number_of_servers; i++)
+   {
+      data = pgmoneta_append(data, "pgmoneta_current_wal_lsn{");
+
+      data = pgmoneta_append(data, "server=\"");
+      data = pgmoneta_append(data, config->servers[i].name);
+      data = pgmoneta_append(data, "\", ");
+
+      data = pgmoneta_append(data, "lsn=\"");
+      if (!strcmp(config->servers[i].current_wal_lsn, ""))
+      {
+         data = pgmoneta_append(data, "0/0");
+      }
+      else
+      {
+         data = pgmoneta_append(data, config->servers[i].current_wal_lsn);
+      }
       data = pgmoneta_append(data, "\"} ");
 
       data = pgmoneta_append_bool(data, config->servers[i].wal_streaming);
--- a/src/libpgmoneta/wal.c
+++ b/src/libpgmoneta/wal.c
@@ -62,6 +62,7 @@ static int wal_xlog_offset(size_t xlogptr, int segsize);
 static int wal_convert_xlogpos(char* xlogpos, uint32_t* high32, uint32_t* low32, int segsize);
 static int wal_find_streaming_start(char* basedir, uint32_t* timeline, uint32_t* high32, uint32_t* low32, int segsize);
 static int wal_shipping_setup(int srv, char** wal_shipping);
+static void update_wal_lsn(int srv, size_t xlogptr);
 
 void
 pgmoneta_wal(int srv, char** argv)
@@ -170,7 +171,6 @@ pgmoneta_wal(int srv, char** argv)
       xlogpos = (char*)malloc(xlogpos_size);
       memset(xlogpos, 0, xlogpos_size);
       memcpy(xlogpos, pgmoneta_query_response_get_data(identify_system_response, 2), xlogpos_size);
-
       if (wal_convert_xlogpos(xlogpos, &high32, &low32, segsize))
       {
          goto error;
@@ -200,6 +200,10 @@ pgmoneta_wal(int srv, char** argv)
          pgmoneta_log_error("Error during START_REPLICATION for server %s", config->servers[srv].name);
          goto error;
       }
+
+      // assign xlogpos at the beginning of the streaming to LSN
+      memset(config->servers[srv].current_wal_lsn, 0, MISC_LENGTH);
+      snprintf(config->servers[srv].current_wal_lsn, MISC_LENGTH, "%s", cmd);
 
       type = 0;
 
@@ -379,6 +383,9 @@ pgmoneta_wal(int srv, char** argv)
                         break;
                      }
                   }
+                  // update LSN after a message data is written to the segment
+                  update_wal_lsn(srv, xlogptr);
+
                   wal_send_status_report(ssl, socket, xlogptr, xlogptr, 0);
                   break;
                }
@@ -506,6 +513,16 @@ error:
    free(filename);
    free(xlogpos);
    exit(1);
+}
+
+static void
+update_wal_lsn(int srv, size_t xlogptr)
+{
+   struct configuration* config = (struct configuration*) shmem;
+   uint32_t low32 = xlogptr & 0xffffffff;
+   uint32_t high32 = xlogptr >> 32 & 0xffffffff;
+   memset(config->servers[srv].current_wal_lsn, 0, MISC_LENGTH);
+   snprintf(config->servers[srv].current_wal_lsn, MISC_LENGTH, "%X/%X", high32, low32);
 }
 
 int
--- a/AUTHORS
+++ b/AUTHORS
@@ -13,3 +13,4 @@ Haoran Zhang <andrewzhr9911@gmail.com>
 Hazem Alrawi <hazemalrawi7@gmail.com>
 Shahryar Soltanpour <shahryar.soltanpour@gmail.com>
 Shikhar Soni <shikharish05@gmail.com>
+Nguyen Cong Nhat Le <lenguyencongnhat2001@gmail.com>
--- a/src/libpgmoneta/configuration.c
+++ b/src/libpgmoneta/configuration.c
@@ -2923,6 +2923,7 @@ copy_server(struct server* dst, struct server* src)
    dst->wal_streaming = src->wal_streaming;
    /* dst->valid = src->valid; */
    memcpy(&dst->current_wal_filename[0], &src->current_wal_filename[0], MISC_LENGTH);
+   memcpy(&dst->current_wal_lsn[0], &src->current_wal_lsn[0], MISC_LENGTH);
 }
 
 static void


lenhat509:
Hi @jesperpedersen @Jubilee101 
I made the PR. Please check it out.

src/libpgmoneta/prometheus.c
```diff
@@ -2377,6 +2392,27 @@ size_information(int client_fd)
...
+   for (int i = 0; i < config->number_of_servers; i++)
+   {
+      data = pgmoneta_append(data, "pgmoneta_current_wal_lsn{");
+
+      data = pgmoneta_append(data, "server=\"");
+      data = pgmoneta_append(data, config->servers[i].name);
+      data = pgmoneta_append(data, "\", ");
+
+      data = pgmoneta_append(data, "LSN=\"");
+      data = pgmoneta_append(data, config->servers[i].pgmoneta_current_wal_lsn);
```
Jubilee101, line 2407: Consider the situation where servers[i].pgmoneta_current_wal_lsn is empty, because of server down or network error, you need to put "0/0" instead.

src/libpgmoneta/wal.c
```diff
@@ -171,6 +171,10 @@ pgmoneta_wal(int srv, char** argv)
       memset(xlogpos, 0, xlogpos_size);
       memcpy(xlogpos, pgmoneta_query_response_get_data(identify_system_response, 2), xlogpos_size);
 
+      // Assign the current xlogpos to LSN
+      memset(config->servers[srv].pgmoneta_current_wal_lsn, 0, MISC_LENGTH);
+      snprintf(config->servers[srv].pgmoneta_current_wal_lsn, MISC_LENGTH, "%s", xlogpos);
```
Jubilee101, line 176: I would prefer you move this line downwards, after we have validated it by wal_convert_xlogpos.

src/include/pgmoneta.h
```diff
@@ -198,6 +198,7 @@ struct server
    char wal_slot[MISC_LENGTH];         /**< The WAL slot name */
    char current_wal_filename[MISC_LENGTH]; /**< The current WAL filename*/
    char follow[MISC_LENGTH];           /**< Follow a server */
+   char pgmoneta_current_wal_lsn[MISC_LENGTH]; /**< The current WAL log sequence number*/
```
jesperpedersen, line 201: Move it next to `current_wal_filename` and rename to `current_wal_lsn`

src/libpgmoneta/prometheus.c
```diff
@@ -615,6 +615,21 @@ home_page(int client_fd)
...
+   data = pgmoneta_append(data, "  <h2>pgmoneta_current_wal_lsn</h2>\n");
+   data = pgmoneta_append(data, "  The current WAL log sequence number\n");
+   data = pgmoneta_append(data, "  <table border=\"1\">\n");
+   data = pgmoneta_append(data, "    <tbody>\n");
+   data = pgmoneta_append(data, "      <tr>\n");
+   data = pgmoneta_append(data, "        <td>server</td>\n");
+   data = pgmoneta_append(data, "        <td>The identifier for the server</td>\n");
+   data = pgmoneta_append(data, "      </tr>\n");
+   data = pgmoneta_append(data, "      <tr>\n");
+   data = pgmoneta_append(data, "        <td>LSN</td>\n");
```
jesperpedersen, line 627: Lowercase

src/libpgmoneta/prometheus.c
```diff
@@ -2377,6 +2392,27 @@ size_information(int client_fd)
...
+   data = pgmoneta_append(data, "#TYPE pgmoneta_current_wal_lsn gauge\n");
+   for (int i = 0; i < config->number_of_servers; i++)
+   {
+      data = pgmoneta_append(data, "pgmoneta_current_wal_lsn{");
+
+      data = pgmoneta_append(data, "server=\"");
+      data = pgmoneta_append(data, config->servers[i].name);
+      data = pgmoneta_append(data, "\", ");
+
+      data = pgmoneta_append(data, "LSN=\"");
```
jesperpedersen, line 2406: Lowercase

jesperpedersen (changes requested):

src/libpgmoneta/wal.c
```diff
@@ -420,6 +424,11 @@ pgmoneta_wal(int srv, char** argv)
       pgmoneta_consume_data_row_messages(ssl, socket, buffer, &end_of_timeline_response);
       timeline = atoi(pgmoneta_query_response_get_data(end_of_timeline_response, 0));
       xlogpos = pgmoneta_query_response_get_data(end_of_timeline_response, 1);
+
+      // Assign the current xlogpos to LSN
+      memset(config->servers[srv].pgmoneta_current_wal_lsn, 0, MISC_LENGTH);
+      snprintf(config->servers[srv].pgmoneta_current_wal_lsn, MISC_LENGTH, "%s", xlogpos);
+
```
lenhat509, line 431: @Jubilee101 Would you like me to move this block after the xlogpos validation too?

Jubilee101: Yup, that would be great

src/libpgmoneta/wal.c
lenhat509: I had this block of code after the wal_convert_xlogpos if-block originally. wal_convert_xlogpos use strtok on xlogpos by '/' character. From what I understand about strtok, after the wal_convert_xlogpos exits, xlogpos will become '0' instead of '0/19FCACO' for example. 0 was also displayed as LSN on Prometheus metrics when I ran the code. So I moved that block up.

src/include/pgmoneta.h
lenhat509: I'll push the code with the update

src/libpgmoneta/wal.c
Jubilee101: Nice catch! Yeah you are right. Then just leave it there, I'll submit a patch to improve the xlogpos parsing and move it down myself later.

src/libpgmoneta/prometheus.c
lenhat509: I fixed in the new commit

src/libpgmoneta/wal.c
lenhat509: Got it.

src/libpgmoneta/prometheus.c
lenhat509: I added the check in the new commit

lenhat509:
@Jubilee101 @jesperpedersen I have made the requested changes.

jesperpedersen:
You need to add yourself to the `AUTHORS` file, run `uncrustify.sh`, squash your commits and then force push

lenhat509:
@jesperpedersen Done. Is there anything else I need to add?

src/libpgmoneta/wal.c
```diff
@@ -171,6 +171,10 @@ pgmoneta_wal(int srv, char** argv)
       memset(xlogpos, 0, xlogpos_size);
       memcpy(xlogpos, pgmoneta_query_response_get_data(identify_system_response, 2), xlogpos_size);
 
+      // Assign the current xlogpos to LSN
+      memset(config->servers[srv].current_wal_lsn, 0, MISC_LENGTH);
+      snprintf(config->servers[srv].current_wal_lsn, MISC_LENGTH, "%s", xlogpos);
```
Jubilee101, line 176: Hey I took some closer look into it, and you shouldn't use this xlogpos here. We got that from IDENTIFY_SYSTEM command. So that's the current lsn on the server side, not on our side.

Jubilee101: A better place to update xlogpos on streaming starts would be when [sending the START_REPLICATION message](https://github.com/pgmoneta/pgmoneta/blob/1be259e66d61ef58cd828894e0e15ffc9723c42e/src/libpgmoneta/wal.c#L192).

lenhat509: I see. I'll remove this portion of the code.

src/libpgmoneta/wal.c
```diff
@@ -420,6 +424,11 @@ pgmoneta_wal(int srv, char** argv)
       pgmoneta_consume_data_row_messages(ssl, socket, buffer, &end_of_timeline_response);
       timeline = atoi(pgmoneta_query_response_get_data(end_of_timeline_response, 0));
       xlogpos = pgmoneta_query_response_get_data(end_of_timeline_response, 1);
+
+      // Assign the current xlogpos to LSN
+      memset(config->servers[srv].current_wal_lsn, 0, MISC_LENGTH);
+      snprintf(config->servers[srv].current_wal_lsn, MISC_LENGTH, "%s", xlogpos);
```
Jubilee101, line 430: You are merely updating xlogpos at the beginning and the end of the streaming. You should also update that during the streaming, i.e. [inside the inner while loop](https://github.com/pgmoneta/pgmoneta/blob/1be259e66d61ef58cd828894e0e15ffc9723c42e/src/libpgmoneta/wal.c#L267). Here xlogptr can be converted to xlogpos - with some bit manipulation to extract high and low 32 bits.

Jubilee101: I know it's a lot of back and forth, but I think you are close.

lenhat509: No problem. I'll look into it now. Thanks.

src/libpgmoneta/wal.c
Jubilee101: > A better place to update xlogpos on streaming starts would be when [sending the START_REPLICATION message](https://github.com/pgmoneta/pgmoneta/blob/1be259e66d61ef58cd828894e0e15ffc9723c42e/src/libpgmoneta/wal.c#L192). The naming is a little inconsistent, sorry about that.

Yeah also don't forget about this :)

lenhat509:
@Jubilee101  There is a bit delay in communication. Please ignore my latest push. I'll request your review in the next push.

lenhat509:
@Jubilee101 `lsn` is now updated at the streaming start, during the streaming, and at the steaming end. I ran this version on a VM, `pgmoneta_current_wal_lsn` on Prometheus is now in sync with `pg_current_wal_lsn()` from the database.

src/libpgmoneta/wal.c
```diff
@@ -379,6 +383,11 @@ pgmoneta_wal(int srv, char** argv)
                         break;
                      }
                   }
+                  // assign xlogpos to LSN after a message data is written to the segment
+                  xlogpos = wal_xlogptr_to_xlogpos(xlogptr);
```
jesperpedersen, line 387: We don't need `xlogpos` anymore since we have `config->servers[srv].current_wal_lsn`

src/libpgmoneta/wal.c
```diff
@@ -508,6 +522,16 @@ pgmoneta_wal(int srv, char** argv)
    exit(1);
 }
 
+char*
```
jesperpedersen, line 525: missing `static`

src/libpgmoneta/wal.c
```diff
@@ -508,6 +522,16 @@ pgmoneta_wal(int srv, char** argv)
...
+wal_xlogptr_to_xlogpos(size_t xlogptr)
```
jesperpedersen, line 526: Rename to `update_wal_lsn` and update the function signature

jesperpedersen (changes requested):

jesperpedersen:
Your patch is leaking memory - you can use https://valgrind.org/ to check for those.

However, it will be fixed once you make the above changes

src/libpgmoneta/wal.c
lenhat509: Thanks for letting me know. I fixed it in the latest push.

src/libpgmoneta/wal.c
lenhat509: Updated it. Instead of returning a value, I assigned a new value to `config->servers[srv].current_wal_lsn` inside the function. It can avoid memory leaks as well.

src/libpgmoneta/wal.c
lenhat509: Removed it.

lenhat509:
> Your patch is leaking memory - you can use https://valgrind.org/ to check for those.
> 
> However, it will be fixed once you make the above changes

Thanks for letting me know. I removed the use of `malloc` and used local variables instead. Please take a look.

src/libpgmoneta/prometheus.c
```diff
@@ -2377,6 +2392,34 @@ size_information(int client_fd)
...
+   for (int i = 0; i < config->number_of_servers; i++)
+   {
+      data = pgmoneta_append(data, "pgmoneta_current_wal_lsn{");
+
+      data = pgmoneta_append(data, "server=\"");
+      data = pgmoneta_append(data, config->servers[i].name);
+      data = pgmoneta_append(data, "\", ");
+
+      data = pgmoneta_append(data, "lsn=\"");
+      if (strcmp(config->servers[i].current_wal_lsn, "") == 0)
```
jesperpedersen, line 2407: Use `!strcmp()`

src/libpgmoneta/wal.c
```diff
@@ -62,6 +62,7 @@ static int wal_xlog_offset(size_t xlogptr, int segsize);
 static int wal_convert_xlogpos(char* xlogpos, uint32_t* high32, uint32_t* low32, int segsize);
 static int wal_find_streaming_start(char* basedir, uint32_t* timeline, uint32_t* high32, uint32_t* low32, int segsize);
 static int wal_shipping_setup(int srv, char** wal_shipping);
+static void update_wal_lsn(char* current_wal_lsn, size_t xlogptr);
```
jesperpedersen, line 65: You can resolve `struct configuration` within the function, so no need for a parameter

src/libpgmoneta/wal.c
```diff
@@ -201,6 +201,10 @@ pgmoneta_wal(int srv, char** argv)
          goto error;
       }
 
+      // assign xlogpos at the beginning of the streaming to LSN
+      memset(config->servers[srv].current_wal_lsn, 0, MISC_LENGTH);
+      snprintf(config->servers[srv].current_wal_lsn, MISC_LENGTH, "%s", cmd);
```
jesperpedersen, line 206: The `cmd` stuff is weird... just use the new function

src/libpgmoneta/wal.c
```diff
@@ -420,6 +427,11 @@ pgmoneta_wal(int srv, char** argv)
       pgmoneta_consume_data_row_messages(ssl, socket, buffer, &end_of_timeline_response);
       timeline = atoi(pgmoneta_query_response_get_data(end_of_timeline_response, 0));
       xlogpos = pgmoneta_query_response_get_data(end_of_timeline_response, 1);
+
+      // assign xlogpos at the end of the steaming to LSN
+      memset(config->servers[srv].current_wal_lsn, 0, MISC_LENGTH);
+      snprintf(config->servers[srv].current_wal_lsn, MISC_LENGTH, "%s", xlogpos);
```
jesperpedersen, line 433: Same here - new function

src/libpgmoneta/wal.c
```diff
@@ -508,6 +520,15 @@ pgmoneta_wal(int srv, char** argv)
    exit(1);
 }
 
+static void
+update_wal_lsn(char* current_wal_lsn, size_t xlogptr)
+{
+   uint32_t low32 = xlogptr & 0xffffffff;
```
jesperpedersen, line 526: Resolve `configuration` here

jesperpedersen (changes requested):

src/libpgmoneta/wal.c
lenhat509: `configuration` can be resolved from `shmem` and we still need `int srv` from `pgmoneta_wal` function. The function signature I can think of is `static void update_wal_lsn(int srv, size_t xlogptr);`. Is this okay?

lenhat509: or it should be `static void update_wal_lsn(size_t xlogptr);`?

Jubilee101: I think it is ok to pass in `int srv`.

src/libpgmoneta/wal.c
Jubilee101: The new function is used to convert xlogptr to xlogpos, and assign the value to `current_wal_lsn`. Here `cmd` is in fact already the xlogpos.

src/libpgmoneta/wal.c
Jubilee101: I don't think we even need these here on second thought. We are getting this from the server, which is not necessarily where we start streaming when the next timeline starts - in-segment offset should be removed first - so I think updating the lsn when sending START_REPLICATION in the next timeline is enough.

src/libpgmoneta/wal.c
jesperpedersen: Yes

src/libpgmoneta/wal.c
jesperpedersen: We need to update the LSN along the current timeline as well

Jubilee101: It is updated.. This xlogpos is just server's cue on where we should start on the next timeline.

lenhat509: The LSN is updated by the new function every time we receives a message data. So it is updated along the timeline, I believe.

lenhat509: I’ll remove this block of code.

src/libpgmoneta/wal.c
lenhat509: Got it. Thanks.

lenhat509:
I have updated the new function, updated prometheus.c, and deleted the LSN update with xlogpos from the server at the end of the timeline.

jesperpedersen:
Merged.

Welcome to the community !

PR #214, opened 2024-03-15
 [#218] Expose backup throughput to prometheus

GuChad369:


GuChad369:
It's a bit odd; the information is already available in Prometheus.

GuChad369:
It's a bit odd; the information is already available in Prometheus.
Screenshot 2024-03-15 at 9 57 25 AM

Jubilee101:
Hi, yes you are right, the [metrics](https://pgmoneta.github.io/metrics.html) haven't been updated for a while. Do you mind changing the metric to backup throughput instead? The backup time elapsed is kept [here](https://github.com/pgmoneta/pgmoneta/blob/19e165139f3082cbadc43a7fed439aa7d620b887/src/libpgmoneta/backup.c#L142).

Sorry about the confusion. This won't change your code too much.

GuChad369:
Sure!

I looked up the compression ratio on [Wikipedia](https://en.wikipedia.org/wiki/Data_compression_ratio#:~:text=5%20External%20links-,Definition,implicit%20ratio%2C%205%2F1), so I added one metric by simply switching the divisor and the dividend.

Alternatively, I could replace the original one with this, or I could delete the newly added one.

And I want to confirm that backup throughput is calculated as size/time. Is that correct?

Jubilee101:
> I looked up the compression ratio on [Wikipedia](https://en.wikipedia.org/wiki/Data_compression_ratio#:~:text=5%20External%20links-,Definition,implicit%20ratio%2C%205%2F1), so I added one metric by simply switching the divisor and the dividend.
> 
> Alternatively, I could replace the original one with this, or I could delete the newly added one.

I think one metric about the ratio is enough. So just replace that with throughput.

> And I want to confirm that backup throughput is calculated as size/time. Is that correct?

Yup.

GuChad369:
> I think one metric about the ratio is enough. So just replace that with throughput.
Okay, I will delete the newly added one, but do I also need to delete the original one?

GuChad369:
> I think one metric about the ratio is enough. So just replace that with throughput.

Okay, I will delete the newly added one, but do I also need to delete the original one?

Jubilee101:
> Okay, I will delete the newly added one, but do I also need to delete the original one?

Nope, keep that :)

GuChad369:
Do you have any advice on how to test the code?

Jubilee101:
We don't currently have any testing infrastructure... although we should some day. For testing just verify the result on prometheus - install prometheus locally, and configure it to connect to pgmoneta on port 5001, then open prometheus ui on port 9090, verify that the metrics are shown correctly. You can also check mem leak by running valgrind, here's a [doc](https://github.com/pgmoneta/pgmoneta/blob/main/contrib/valgrind/README.md) about this

src/libpgmoneta/prometheus.c
```diff
@@ -2195,6 +2195,89 @@ size_information(int client_fd)
       data = NULL;
    }
```
Jubilee101, line 2197: You need to define your metrics. See [here](https://github.com/pgmoneta/pgmoneta/blob/6a1ea10706c33bfdf34ed6c2b814f920100c34b4/src/libpgmoneta/prometheus.c#L570).

src/libpgmoneta/prometheus.c
```diff
@@ -2195,6 +2195,89 @@ size_information(int client_fd)
...
+               data = pgmoneta_append(data, backups[j]->label);
+               data = pgmoneta_append(data, "\"} ");
+
+               if (backups[j]->elapsed_time)
+               {
+                  unsigned long throughput = backups[j]->backup_size / backups[j]->elapsed_time;
+                  if (throughput < 1024)
+                  {
+                     data = pgmoneta_append_double(data, 1.0 * throughput);
+                     data = pgmoneta_append(data, " B/s");
```
Jubilee101, line 2229: I don't think prometheus allows string in the metrics. You need to run an prometheus instance to verify your result. See https://prometheus.io/docs/prometheus/latest/getting_started/

Jubilee101: Just settle for a fixed unit and docmuent it.

src/libpgmoneta/prometheus.c
```diff
@@ -2195,6 +2195,89 @@ size_information(int client_fd)
       data = NULL;
    }
 
+   data = pgmoneta_append(data, "#HELP pgmoneta_backup_newest_throughput The throughput of the newest backup for a server\n");
```
Jubilee101, line 2198: What is "newest" for? You are getting the througput for each and every backup, aren't you?

src/libpgmoneta/prometheus.c
Jubilee101: Or alternatively you can keep the unit as a parameter, that should be better.

Jubilee101:
The [grafana.md](https://github.com/pgmoneta/pgmoneta/blob/6a1ea10706c33bfdf34ed6c2b814f920100c34b4/doc/tutorial/10_grafana.md) also has some tutorials about prometheus.

src/libpgmoneta/prometheus.c
GuChad369: Yes, it doesn't permit it. I just made it bytes/seconds and will try to mention it somewhere.

GuChad369:
I ran it successfully and got the value from Prometheus.
Screenshot 2024-03-16 at 10 31 42 AM

GuChad369:
The problem now is that there are a bunch of leaks in this run. How can we fix it?
Screenshot 2024-03-16 at 10 26 30 AM

jesperpedersen:
Where are they coming from ? Attach the log instead of a screenshot

GuChad369:
[68679.log](https://github.com/pgmoneta/pgmoneta/files/14624632/68679.log)
I'm not sure why this happened. Did I run it incorrectly?

src/libpgmoneta/prometheus.c
```diff
@@ -421,6 +421,21 @@ home_page(int client_fd)
    data = pgmoneta_append(data, "    </tbody>\n");
    data = pgmoneta_append(data, "  </table>\n");
    data = pgmoneta_append(data, "  <p>\n");
+   data = pgmoneta_append(data, "  <h2>pgmoneta_backup_throughput</h2>\n");
+   data = pgmoneta_append(data, "  The throughput of the backup for a server\n");
```
Jubilee101, line 425: Perhaps you should mention `B/s` here.

Jubilee101:
This looks good to me now :) I don't think the mem leak is your problem. We'll look at it later.

Jubilee101 (approved):

src/libpgmoneta/prometheus.c
```diff
@@ -2195,6 +2210,73 @@ size_information(int client_fd)
       data = NULL;
    }
 
+   data = pgmoneta_append(data, "#HELP pgmoneta_backup_throughput The throughput of the backup for a server\n");
```
jesperpedersen, line 2213: Add (bytes / s) here as well

jesperpedersen (changes requested):

GuChad369:
@Jubilee101 Could you guide the initial step for accessing PostgreSQL server-side files, particularly with .c files or functions?

jesperpedersen:
Can you rebase this, please

GuChad369:
Done. 
I'm not sure if this approach is correct.

jesperpedersen:
You need to squash

GuChad369:
Done

jesperpedersen:
I don't know what you are doing, but it doesn't branch from `main`

GuChad369:
Done.
I've acquired new knowledge. Thank you for your patience and guidance.

jesperpedersen:
Merged.

Issue #197, opened 2024-03-01
Rate limit: Disk


jesperpedersen:
Provide a rate limit (global and server) to limit disk I/O traffic.

This is for backup

GuChad369:
For this issue, I have some basic ideas:
1. Set a rate limit globally and per server.
2. Create a function to measure the rate.
3. During the I/O process, fork a child process to check its value against the rate limit. If it exceeds the limit, use usleep() or send a signal to slow down the process.
Is this approach correct or naive? Could I get some help on which files might be involved?

jesperpedersen:
It should be handled in-process for the backup process. And, its `wf_backup.c` and friends.

GuChad369:
Is this [part](https://github.com/pgmoneta/pgmoneta/blob/f13ae08e5020e4bd36964060588b41b23d34a8bf/src/libpgmoneta/message.c#L2511-L2523) actually performing the write to disk? If we are implementing rate limiting, should we do it during this phase?

I plan to use the `token bucket algorithm` here to limit the rate. Is this possible?

jesperpedersen:
Correct

GuChad369:
Can we use GCC built-in functions `like __sync_bool_compare_and_swap` in the code? Since it involves multithreading, I want to make it lockless for efficiency. Is this approach allowed?

jesperpedersen:
Use `#include <stdatomic.h>`

PR #227, opened 2024-03-24
[#197] Rate limit: Disk

GuChad369:
This is just the first version of the rate limit. I have only implemented rate limiting for `pgmoneta_receive_archive_stream`, not for `pgmoneta_receive_archive_files`. 

What I want to make sure:
- This is the correct way to implement the limit.
- The code style is consistent with the rest of the file for integration.

Test Run:
- Successfully implemented backup rate limiting with no errors for now.

### DIFF
--- a/doc/CONFIGURATION.md
+++ b/doc/CONFIGURATION.md
@@ -65,6 +65,7 @@ See a [sample](./etc/pgmoneta.conf) configuration for running `pgmoneta` on `loc
 | tls_ca_file | | String | No | Certificate Authority (CA) file for TLS. This file must be owned by either the user running pgmoneta or root.  |
 | libev | `auto` | String | No | Select the [libev](http://software.schmorp.de/pkg/libev.html) backend to use. Valid options: `auto`, `select`, `poll`, `epoll`, `iouring`, `devpoll` and `port` |
 | buffer_size | 65535 | Int | No | The network buffer size (`SO_RCVBUF` and `SO_SNDBUF`) |
+| backup_max_rate | 0 | Int | No | The number of bytes of tokens added every one second to limit the backup rate|
 | keep_alive | on | Bool | No | Have `SO_KEEPALIVE` on sockets |
 | nodelay | on | Bool | No | Have `TCP_NODELAY` on sockets |
 | non_blocking | on | Bool | No | Have `O_NONBLOCK` on sockets |
--- a/src/include/pgmoneta.h
+++ b/src/include/pgmoneta.h
@@ -59,6 +59,9 @@ extern "C" {
 #define MAX_BUFFER_SIZE      65535
 #define DEFAULT_BUFFER_SIZE  65535
 
+#define DEFAULT_BURST 65535
+#define DEFAULT_EVERY 1
+
 #define MAX_USERNAME_LENGTH  128
 #define MAX_PASSWORD_LENGTH 1024
 
@@ -351,6 +354,8 @@ struct configuration
    int number_of_servers;        /**< The number of servers */
    int number_of_users;          /**< The number of users */
    int number_of_admins;         /**< The number of admins */
+
+   unsigned long backup_max_rate; /**< Number of tokens added to the bucket with each replenishment for backup. */
 
    struct server servers[NUMBER_OF_SERVERS];       /**< The servers */
    struct user users[NUMBER_OF_USERS];             /**< The users */
--- a/src/include/utils.h
+++ b/src/include/utils.h
@@ -919,6 +919,60 @@ pgmoneta_get_file_size(char* file_path);
 bool
 pgmoneta_is_file_archive(char* file_path);
 
+/** @struct
+ * Defines token bucket structure
+ */
+struct token_bucket
+{
+   unsigned long burst; /**< default value is 0, no limit */
+   atomic_ulong cur_tokens;
+   unsigned long max_rate;
+   int every;
+   atomic_ulong last_time;
+};
+
+/**
+ * Init a token bucket
+ * @param tb The token bucket
+ * @param max_rate The number of bytes of tokens added every one second
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_token_bucket_init(struct token_bucket* tb, unsigned long max_rate);
+
+/**
+ * Free the memory of the token bucket
+ * @param tb The token bucket
+ */
+void
+pgmoneta_token_bucket_destroy(struct token_bucket* tb);
+
+/**
+ * Add new token into the bucket
+ * @param tb The token bucket
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_token_bucket_add(struct token_bucket* tb);
+
+/**
+ * Get tokens from token bucket wrapper
+ * @param tb The token bucket
+ * @param tokens Needed tokens
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_token_bucket_consume(struct token_bucket* tb, unsigned long tokens);
+
+/**
+ * Get tokens from token bucket once
+ * @param tb The token bucket
+ * @param tokens Needed tokens
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_token_bucket_once(struct token_bucket* tb, unsigned long tokens);
+
 #ifdef DEBUG
 
 /**
--- a/src/libpgmoneta/configuration.c
+++ b/src/libpgmoneta/configuration.c
@@ -56,6 +56,7 @@
 
 static void extract_key_value(char* str, char** key, char** value);
 static int as_int(char* str, int* i);
+static int as_ulong(char* str, unsigned long* ul);
 static int as_bool(char* str, bool* b);
 static int as_logging_type(char* str);
 static int as_logging_level(char* str);
@@ -127,6 +128,8 @@ pgmoneta_init_configuration(void* shm)
    config->log_level = PGMONETA_LOGGING_LEVEL_INFO;
    config->log_mode = PGMONETA_LOGGING_MODE_APPEND;
    atomic_init(&config->log_lock, STATE_FREE);
+
+   config->backup_max_rate = 0;
 
    return 0;
 }
@@ -1169,6 +1172,20 @@ pgmoneta_read_configuration(void* shm, char* filename)
                      unknown = true;
                   }
                }
+               else if (!strcmp(key, "backup_max_rate"))
+               {
+                  if (!strcmp(section, "pgmoneta"))
+                  {
+                     if (as_ulong(value, &config->backup_max_rate))
+                     {
+                        unknown = true;
+                     }
+                  }
+                  else
+                  {
+                     unknown = true;
+                  }
+               }
                else
                {
                   unknown = true;
@@ -2008,6 +2025,46 @@ as_int(char* str, int* i)
    }
 
    *i = (int)val;
+
+   return 0;
+
+error:
+
+   errno = 0;
+
+   return 1;
+}
+
+static int
+as_ulong(char* str, unsigned long* ul)
+{
+   char* endptr;
+   long val;
+
+   errno = 0;
+   val = strtol(str, &endptr, 10);
+
+   if ((errno == ERANGE && (val == LONG_MAX || val == LONG_MIN)) || (errno != 0 && val == 0))
+   {
+      goto error;
+   }
+
+   if (str == endptr)
+   {
+      goto error;
+   }
+
+   if (*endptr != '\0')
+   {
+      goto error;
+   }
+
+   if (val < 0)
+   {
+      goto error;
+   }
+
+   *ul = (unsigned long)val;
 
    return 0;
 
--- a/src/libpgmoneta/message.c
+++ b/src/libpgmoneta/message.c
@@ -2145,7 +2145,7 @@ error:
 }
 
 int
-pgmoneta_receive_archive_files(SSL* ssl, int socket, struct stream_buffer* buffer, char* basedir, struct tablespace* tablespaces, int version)
+pgmoneta_receive_archive_files(SSL* ssl, int socket, struct stream_buffer* buffer, char* basedir, struct tablespace* tablespaces, int version, struct token_bucket* bucket)
 {
    struct query_response* response = NULL;
    struct message* msg = (struct message*)malloc(sizeof (struct message));
@@ -2240,6 +2240,22 @@ pgmoneta_receive_archive_files(SSL* ssl, int socket, struct stream_buffer* buffe
 
          if (msg->kind == 'd' && msg->length > 0)
          {
+
+            if (bucket)
+            {
+               while (1)
+               {
+                  if (!pgmoneta_token_bucket_consume(bucket, msg->length))
+                  {
+                     break;
+                  }
+                  else
+                  {
+                     SLEEP(500000000L)
+                  }
+               }
+            }
+
             // copy data
             if (fwrite(msg->data, msg->length, 1, file) != 1)
             {
@@ -2274,7 +2290,7 @@ pgmoneta_receive_archive_files(SSL* ssl, int socket, struct stream_buffer* buffe
    // If server version >= 13, receive manifest as well
    if (version >= 13)
    {
-      if (pgmoneta_receive_manifest_file(ssl, socket, buffer, basedir))
+      if (pgmoneta_receive_manifest_file(ssl, socket, buffer, basedir, bucket))
       {
          goto error;
       }
@@ -2340,7 +2356,7 @@ error:
 }
 
 int
-pgmoneta_receive_archive_stream(SSL* ssl, int socket, struct stream_buffer* buffer, char* basedir, struct tablespace* tablespaces)
+pgmoneta_receive_archive_stream(SSL* ssl, int socket, struct stream_buffer* buffer, char* basedir, struct tablespace* tablespaces, struct token_bucket* bucket)
 {
    struct query_response* response = NULL;
    struct message* msg = (struct message*)malloc(sizeof (struct message));
@@ -2515,6 +2531,22 @@ pgmoneta_receive_archive_stream(SSL* ssl, int socket, struct stream_buffer* buff
                {
                   break;
                }
+
+               if (bucket)
+               {
+                  while (1)
+                  {
+                     if (!pgmoneta_token_bucket_consume(bucket, msg->length))
+                     {
+                        break;
+                     }
+                     else
+                     {
+                        SLEEP(500000000L)
+                     }
+                  }
+               }
+
                if (fwrite(msg->data + 1, msg->length - 1, 1, file) != 1)
                {
                   pgmoneta_log_error("could not write to file %s", file_path);
@@ -2610,7 +2642,7 @@ error:
 }
 
 int
-pgmoneta_receive_manifest_file(SSL* ssl, int socket, struct stream_buffer* buffer, char* basedir)
+pgmoneta_receive_manifest_file(SSL* ssl, int socket, struct stream_buffer* buffer, char* basedir, struct token_bucket* bucket)
 {
    char tmp_file_path[MAX_PATH];
    char file_path[MAX_PATH];
@@ -2620,6 +2652,7 @@ pgmoneta_receive_manifest_file(SSL* ssl, int socket, struct stream_buffer* buffe
 
    memset(tmp_file_path, 0, sizeof(tmp_file_path));
    memset(file_path, 0, sizeof(file_path));
+
    // Name the manifest with .tmp suffix so that we know backup is invalid if replication is interrupted
    if (pgmoneta_ends_with(basedir, "/"))
    {
@@ -2644,6 +2677,7 @@ pgmoneta_receive_manifest_file(SSL* ssl, int socket, struct stream_buffer* buffe
       }
       pgmoneta_consume_copy_stream_end(buffer, msg);
    }
+
    while (msg->kind != 'c')
    {
       pgmoneta_consume_copy_stream_start(ssl, socket, buffer, msg);
@@ -2655,6 +2689,22 @@ pgmoneta_receive_manifest_file(SSL* ssl, int socket, struct stream_buffer* buffe
       }
       if (msg->kind == 'd' && msg->length > 0)
       {
+
+         if (bucket)
+         {
+            while (1)
+            {
+               if (!pgmoneta_token_bucket_consume(bucket, msg->length))
+               {
+                  break;
+               }
+               else
+               {
+                  SLEEP(500000000L)
+               }
+            }
+         }
+
          // copy data
          if (fwrite(msg->data, msg->length, 1, file) != 1)
          {
--- a/src/libpgmoneta/utils.c
+++ b/src/libpgmoneta/utils.c
@@ -43,6 +43,7 @@
 #include <inttypes.h>
 #include <libgen.h>
 #include <pwd.h>
+#include <stdatomic.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -2018,7 +2019,6 @@ pgmoneta_basename_file(char* s, char** basename)
    return 0;
 
 error:
-
    return 1;
 }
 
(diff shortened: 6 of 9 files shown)


GuChad369:
Why did that happen? I ran this code locally without the error.

Jubilee101:
Hi, thanks for working on this! 

I'm no expert on this part, but overall it seems quite nice. I assume you have verified the throughput changes? Also perhaps restore a backup and start a server on it to make sure there's no data corruption(unlikely, but just in case). 

Also note that you are merely controlling disk throughput in `pgmoneta_receive_archive_stream`, which is only for server version >= 15. For earlier version postgres does base backup differently, see `pgmoneta_receive_archive_files`, which also includes `pgmoneta_receive_manifest_file`. 

And, a fixed disk IO rate probably won't cut it. Perhaps you need to add configuration options for bucket burst and rate limit. You can use current parameter as default. Let me know if you need more info on this part.

Also just a thought, seems that the bucket only controls one streaming process. Would it be possible for the bucket to add token lazily? i.e. Only calculate how many tokens should've been added since last time each time when the process consumes the tokens. I think it's possible to use float tokens. That way you may be able to avoid using a separate thread.

GuChad369:
@Jubilee101 Thank you for your advice; it has been very helpful and has made this structure more robust.

> I'm no expert on this part, but overall it seems quite nice. I assume you have verified the throughput changes? Also perhaps restore a backup and start a server on it to make sure there's no data corruption(unlikely, but just in case).
> Also note that you are merely controlling disk throughput in pgmoneta_receive_archive_stream, which is only for server version >= 15. For earlier version postgres does base backup differently, see pgmoneta_receive_archive_files, which also includes pgmoneta_receive_manifest_file. (edit: sorry just saw you left that out on purpose. That's fine)

This has not been fully implemented yet. For this part, I think the token bucket structure is important. If we successfully create a good algorithm, we can use it whenever we want to implement rate limiting. This is a great way to improve our code, so I just want to make sure we are on the right path and make this algorithm perfect.

> Also just a thought, seems that the bucket only controls one streaming process. Would it be possible for the bucket to add token lazily? i.e. Only calculate how many tokens should've been added since last time each time when the process consumes the tokens. I think it's possible to use float tokens. That way you may be able to avoid using a separate thread.

This is a brilliant idea; I hadn't thought of it that way. So, I've changed the code to avoid threading and just consider the seconds. I'm not sure if we need to consider precise microseconds or nanoseconds.

jesperpedersen:
See `SLEEP()` in `pgmoneta.h`

GuChad369:
> See SLEEP() in pgmoneta.h
Is it necessary to specify measurements in nanoseconds? If we're simply interested in obtaining a rough average rate, wouldn't seconds be sufficient?

GuChad369:
> See SLEEP() in pgmoneta.h

Is it necessary to specify measurements in nanoseconds? If we're simply interested in obtaining a rough average rate, wouldn't seconds be sufficient?

jesperpedersen:
We can do the initial implementation in whole seconds if you think that is the best way

src/libpgmoneta/utils.c
```diff
@@ -3003,6 +3004,87 @@ pgmoneta_is_file_archive(char* file_path)
...
+
+   struct token_bucket* tb;
+   tb = (struct token_bucket*)malloc(sizeof(struct token_bucket));
+   if (tb == NULL)
+   {
+      return NULL;
+   }
+
+   tb->burst = burst;
+   atomic_init(&tb->cur_tokens, burst);
```
Jubilee101, line 3019: Why do we need atomic on tokens and timestamps?

GuChad369: Just in case, this algorithm may be utilized for multi-threading purposes in some instances.

src/include/utils.h
```diff
@@ -805,6 +805,48 @@ pgmoneta_get_file_size(char* file_path);
...
+/** @struct
+ * Defines token bucket structure
+ */
+struct token_bucket
+{
+   unsigned long burst;
+   atomic_ulong cur_tokens;
+   unsigned long max_rate;
+   unsigned long every;
+   _Atomic(time_t) last_time;
```
jesperpedersen, line 817: I would rather use an atomic type that isn't `_Atomic`

src/include/utils.h
```diff
@@ -805,6 +805,48 @@ pgmoneta_get_file_size(char* file_path);
...
+};
+
+/**
+ * Init a token bucket
+ * @param burst The max burst of the bucket
+ * @param max_rate The number of tokens added each time
+ * @param every Add tokens every specified number of seconds
+ * @return The token bucket
+ */
+struct token_bucket*pgmoneta_token_bucket_init(unsigned long burst, unsigned long max_rate, unsigned long every);
```
jesperpedersen, line 827: Formatting.

And use an output parameter and an `int` return value

src/include/utils.h
```diff
@@ -805,6 +805,48 @@ pgmoneta_get_file_size(char* file_path);
...
+
+/**
+ * Free the memory of the token bucket
+ * @param tb The token bucket
+ */
+void pgmoneta_token_bucket_destroy(struct token_bucket* tb);
```
jesperpedersen, line 833: Formatting

src/include/utils.h
```diff
@@ -805,6 +805,48 @@ pgmoneta_get_file_size(char* file_path);
...
+
+/**
+ * Add new token into the bucket
+ * @param tb The token bucket
+ * @return 0 upon success, otherwise 1
+ */
+int pgmoneta_token_bucket_add(struct token_bucket* tb);
```
jesperpedersen, line 840: Formatting

src/include/utils.h
```diff
@@ -805,6 +805,48 @@ pgmoneta_get_file_size(char* file_path);
...
+
+/**
+ * Get tokens from token bucket
+ * @param tb The token bucket
+ * @param tokens Needed tokens
+ * @return 0 upon success, otherwise 1
+ */
+int pgmoneta_token_bucket_consume(struct token_bucket* tb, unsigned long tokens);
```
jesperpedersen, line 848: Formatting

src/libpgmoneta/message.c
```diff
@@ -2380,6 +2380,10 @@ pgmoneta_receive_archive_stream(SSL* ssl, int socket, struct stream_buffer* buff
       pgmoneta_consume_copy_stream_end(buffer, msg);
    }
 
+   // Initialize the token bucket to limit the rate to 512 MB/s
+   // Capacity is set to 1 GB
+   struct token_bucket* bucket = pgmoneta_token_bucket_init(1024 * 1024 * 1024, 512 * 1024 * 1024, 1);
```
jesperpedersen, line 2385: No, default is full speed.

This need to come from a configuration parameter or multiple

src/libpgmoneta/utils.c
```diff
@@ -3003,6 +3004,87 @@ pgmoneta_is_file_archive(char* file_path)
...
+
+   time_t diff = time(NULL) - atomic_load(&tb->last_time);
+
+   if (diff <= 0)
+   {
+      atomic_store(&tb->last_time, time(NULL));
+      return 0;
+   }
+
+   unsigned long expected = atomic_load(&tb->cur_tokens);
```
jesperpedersen, line 3047: Variables need to defined at the start of the function

src/libpgmoneta/utils.c
```diff
@@ -3003,6 +3004,87 @@ pgmoneta_is_file_archive(char* file_path)
...
+   }
+
+   atomic_store(&tb->last_time, time(NULL));
+   return 0;
+}
+
+int
+pgmoneta_token_bucket_consume(struct token_bucket* tb, unsigned long tokens)
+{
+   if (pgmoneta_token_bucket_add(tb) == 0)
```
jesperpedersen, line 3073: Use `!` instead of `== 0`

src/libpgmoneta/utils.c
```diff
@@ -53,6 +53,7 @@
 #include <sys/time.h>
 #include <sys/stat.h>
 #include <sys/types.h>
+#include <stdatomic.h>
```
jesperpedersen, line 56: Sorting

jesperpedersen (changes requested):

GuChad369:
This is the completed version of the rate limit feature:

- I have implemented a nanosecond version for the token bucket, but we need to decide which version to keep. The nanosecond version may affect efficiency.
- I have successfully tested the stream version, but I'm unsure how to test the other two functions.

src/libpgmoneta/utils.c
```diff
@@ -3003,6 +3004,198 @@ pgmoneta_is_file_archive(char* file_path)
...
+
+      if (new_tokens > tb->burst)
+      {
+         new_tokens = tb->burst;
+      }
+   }
+
+   expected_time = atomic_load(&tb->last_time);
+   cur_time = (unsigned long) time(NULL);
+   while (!atomic_compare_exchange_weak(&tb->last_time, &expected_time, cur_time))
```
Jubilee101, line 3068: Correct me if I'm wrong, but think of this situation: suppose you add 1 token/sec, and last_time=1, at time=2 a thread comes asking the bucket, so that one token is added to the bucket. But before you update last_time, another thread at time=3 comes asking the bucket, since last_time is still 1, 2 tokens are added. So tokens at time=2 are added twice.

src/libpgmoneta/message.c
```diff
@@ -2655,6 +2749,28 @@ pgmoneta_receive_manifest_file(SSL* ssl, int socket, struct stream_buffer* buffe
...
+
+         if (bucket->burst)
+         {
+            if (msg->length > bucket->burst)
+            {
+               pgmoneta_log_error("the file size(%ld) exceeds the burst capacity of the bucket(%ld).", msg->length, bucket->burst);
+               goto error;
+            }
+
+            while (1)
```
Jubilee101, line 2761: I think user should have the option not to use rate limiter, if they don't configure it, or set it to `off`.

src/libpgmoneta/utils.c
GuChad369: Yes, you are right. I will make them update atomically or other ways.

src/libpgmoneta/message.c
GuChad369: The default value is 0, and it will not be used if a specific value is not provided.

src/libpgmoneta/utils.c
```diff
@@ -3003,6 +3004,198 @@ pgmoneta_is_file_archive(char* file_path)
...
+
+   diff = (unsigned long) time(NULL) - atomic_load(&tb->last_time);
+
+   if (diff <= 0)
+   {
+      return 0;
+   }
+
+   expected_tokens = atomic_load(&tb->cur_tokens);
+   new_tokens = expected_tokens + tb->max_rate * diff / tb->every;
```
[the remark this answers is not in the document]
Jubilee101: I fear it's probably necessary. If diff is smaller than every, then no tokens are added if tokens is in integer. But you still update last_time.., which means if you wait for time shorter than every, you will wait forever, since diff is always smaller than every.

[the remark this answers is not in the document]
GuChad369: Yes, that approach is better for controlling the limit, similar to the nanosecond version. However, I think it would be too frequent to manipulate the token. For us, a rough limit should suffice. This way, we can achieve an average limit with less manipulation, which I believe is more efficient.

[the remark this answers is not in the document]
GuChad369: No, if it is smaller, I will just return 0, quit this addition, and not update the time.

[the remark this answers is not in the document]
Jubilee101: Sounds good. Let's stick to your way then.

src/libpgmoneta/utils.c
```diff
@@ -3003,6 +3004,198 @@ pgmoneta_is_file_archive(char* file_path)
...
+{
+   unsigned long diff;
+   unsigned long expected_tokens;
+   unsigned long new_tokens;
+   unsigned long expected_time;
+   unsigned long cur_time;
+
+   diff = (unsigned long) time(NULL) - atomic_load(&tb->last_time);
+
+   if (diff <= 0)
```
Jubilee101, line 3042: Then probably this should be diff <= tb->every

GuChad369: Yes, you are right. I always use 1 second for testing and forgot about that.

src/libpgmoneta/utils.c
GuChad369: `
int 
pgmoneta_token_bucket_add(struct token_bucket* tb)
{
   unsigned long diff;
   unsigned long expected_tokens;
   unsigned long new_tokens;
   unsigned long expected_time;
   unsigned long cur_time;

   cur_time = (unsigned long) time(NULL);
   expected_time = atomic_load(&tb->last_time);

   while (!atomic_compare_exchange_weak(&tb->last_time, &expected_time, cur_time))
   {
      expected_time = atomic_load(&tb->last_time);
	  cur_time = (unsigned long) time(NULL);
   }

   diff = cur_time - expected_time;

   if (diff < tb->every)
   {
      return 0;
   }

   expected_tokens = atomic_load(&tb->cur_tokens);
   new_tokens = expected_tokens + tb->max_rate * diff / tb->every;

   if (new_tokens > tb->burst)
   {
      new_tokens = tb->burst;
   }

   while (!atomic_compare_exchange_weak(&tb->cur_tokens, &expected_tokens, new_tokens))
   {
      expected_tokens = atomic_load(&tb->cur_tokens);
      new_tokens = expected_tokens + tb->max_rate * diff / tb->every;

      if (new_tokens > tb->burst)
      {
         new_tokens = tb->burst;
      }
   }

   return 0;
}`

Do you think this approach is better? By switching the order, even if two threads try to update the token, they will still update their respective sections based on their time differences. This should not affect the result. I think this will address the race condition.

GuChad369: int
    pgmoneta_token_bucket_add(struct token_bucket* tb)
    {
        unsigned long diff;
        unsigned long expected_tokens;
        unsigned long new_tokens;
        unsigned long expected_time;
        unsigned long cur_time;

        cur_time = (unsigned long) time(NULL);
        expected_time = atomic_load(&tb->last_time);

        while (!atomic_compare_exchange_weak(&tb->last_time, &expected_time, cur_time))
        {
            expected_time = atomic_load(&tb->last_time);
            cur_time = (unsigned long) time(NULL);
        }

        diff = cur_time - expected_time;

        if (diff < tb->every)
        {
            return 0;
        }

        expected_tokens = atomic_load(&tb->cur_tokens);
        new_tokens = expected_tokens + tb->max_rate * diff / tb->every;

        if (new_tokens > tb->burst)
        {
            new_tokens = tb->burst;
        }

        while (!atomic_compare_exchange_weak(&tb->cur_tokens, &expected_tokens, new_tokens))
        {
            expected_tokens = atomic_load(&tb->cur_tokens);
            new_tokens = expected_tokens + tb->max_rate * diff / tb->every;

            if (new_tokens > tb->burst)
            {
                new_tokens = tb->burst;
            }
        }

        return 0;
    }

Do you think this approach is better? By switching the order, even if two threads try to update the token, they will still update their respective sections based on their time differences. This should not affect the result. I think this will address the race condition.

Jubilee101: Another good idea to avoid overflow is to get the time it takes to fill up a bucket, if diff is longer than that, just fill up the bucket directly - got this idea from Go's built-in rate limiter.

GuChad369: How about this way, we calculate the `diff` before the update. This way, the current time will not be updated until it is greater than the specified `every` interval. And, in the while update time loop, we continue to track the `diff`. If the `diff` is below the 'every', we exit the function.

Jubilee101: That might work~ Note that we have a built-in worker pool, and it's currently used in [wf_restore.c](https://github.com/pgmoneta/pgmoneta/blob/main/src/libpgmoneta/wf_restore.c), could be a good idea to configure workers and add your rate limiter there and test it out, i.e. restore a backup.

Another thing that worries me is the granularity - if your data buffer to write to disk is larger than the bucket capacity, you risk looping forever. It's probably worthwhile to add a wrapper function with rate limit around fwrite, where you can partially write a buffer each time until it has all been written to disk. But this is open for discussion.

GuChad369: > if your data buffer to write to disk is larger than the bucket capacity, you risk looping forever

I noticed this issue, so I added a line to exit the backup if this situation is encountered and to print the error. Normally, we need the burst to be larger than the buffer, as the burst will not harm the rate limit; it's just a one-time event. We limit it after the burst is consumed. Therefore, even if the burst is set to a large number, it will not affect our rate limit.

GuChad369: I should set a limit [here](https://github.com/pgmoneta/pgmoneta/blob/ca773ca4c088909db7ced2ded3305085a865d3b9/src/libpgmoneta/utils.c#L1875-L1895) to limit restore, right? What do you mean by 'configure workers'? If we simply set the limit here, shouldn't it be limited?

Jubilee101: > I should set a limit [here](https://github.com/pgmoneta/pgmoneta/blob/ca773ca4c088909db7ced2ded3305085a865d3b9/src/libpgmoneta/utils.c#L1875-L1895) to limit restore, right?

I think so.

> What do you mean by 'configure workers'? If we simply set the limit here, shouldn't it be limited?
Sorry about the confusion, it's simply for testing - if you want see if your bucket works well in multi-threading environment, you can test it out there, with [workers configured](https://github.com/pgmoneta/pgmoneta/blob/main/doc/CONFIGURATION.md). A worker pool will be created to copy files with a shared token bucket.

GuChad369: Can we simply place the token bucket in shared memory so that any process needing rate limiting can access and use it? However, this approach would allow only one rate limit configuration to be used universally.

Jubilee101: It's open for discussion. I would prefer one rate limiter per process, either it's for backup streaming or restoring or some other task. However, it can be shared among the worker threads within the process. Have you encountered any trouble with this?

GuChad369: I am currently testing the restore process, but when I run `pgmoneta-cli restore primary newest /tmp/restore`, the log shows `2024-03-28 19:53:02 INFO restore.c:106 Restore: primary/20240328151050 (Elapsed: 00:00:00)`. However, I can't find the restored files in `/tmp/restore`. This issue is unrelated to the `WAL segment error`, right?

Jubilee101: Hmm, odd, I just restored a backup, and it works fine... at least the data is there. And what do you mean by `WAL segment error`?

Jubilee101: Have you checked the data in your backup, is it there in the first place?

GuChad369: > And what do you mean by WAL segment error?

There is alway an error about the WAL 
`2024-03-30 08:34:08 INFO  configuration.c:1466 No WAL slot defined for primary`

GuChad369: image

Does this mean that the restore was successful? I just tested it again today, and it worked, which is so weird. I believe there must be something wrong somewhere.

jesperpedersen: Did you define `wal_slot = ...` and `create_slot = yes` ?

If there are questions during the startup process that isn't covered by the documentation then we need to fix those

GuChad369: I configured `wal_level = replica` in `postgresql.conf`, but what does `create_slot = yes` mean? Do I need to configure it in `pgmoneta.conf`?

Jubilee101: > but what does create_slot = yes

It's for creating a replication slot for WAL streaming. You see WAL segments are generated endlessly, the server cleans up outdated segments regularly. A slot lets server know how much segments have you saved on your end, so that it doesn't clean up WAL files you haven't yet received.

If you don't configure it, you are fine if you keep up. But if you fall behind, say you stopped pgmoneta for a while while the server is still running, you risk losing some segments for server has deleted them.

GuChad369: `pgmoneta_management_read_int32: read: 4 Invalid argument`
What does the management do? When I set the workers and token, the restore process finished at this line. However, when I don't use workers, this problem does not occur.

GuChad369: When I make the token shared and use it to limit the restore process in [copy_file](https://github.com/pgmoneta/pgmoneta/blob/9dfbbb9421577ecbc0e5c2952442b67d44efd26d/src/libpgmoneta/utils.c#L1878-L1953), the function `pgmoneta_management_read_int32` quits the process before the restoration begains. However, if I don't use the token, everything works well.

Jubilee101: `pgmoneta_management_read_int32` is for cli, when you issue restore command, cli sends it to pgmoneta, then tries to read the restoring outcome status from pgmoneta. So some error occurred in between, either pgmoneta fails to send the result, or the connection broke.

Jubilee101: Try adding more loggs to see what happens in between.

Jubilee101: > pgmoneta_management_read_int32 is for cli, when you issue restore command, cli sends it to pgmoneta, then tries to read the restoring outcome status from pgmoneta. So some error occurred in between, either pgmoneta fails to send the result, or the connection broke.

See [restore()](https://github.com/pgmoneta/pgmoneta/blob/9dfbbb9421577ecbc0e5c2952442b67d44efd26d/src/cli.c#L923)

Jubilee101: Chances are rate limit slows down the process, and [read_complete()](https://github.com/pgmoneta/pgmoneta/blob/9dfbbb9421577ecbc0e5c2952442b67d44efd26d/src/libpgmoneta/management.c#L1510) times out. Try increasing number of retries to see if problem persists.

GuChad369: I believe the token bucket can function in a multi-threaded environment and perform rate limiting. I tested it with `works=2` and the restore was successful. Below is the log, and I didn't find any issues there.
[pgmoneta.log](https://github.com/pgmoneta/pgmoneta/files/14813354/pgmoneta.log)

Jubilee101: Did you solve your previous problem? What's the issue there?

GuChad369: I made a mistake in the shared memory, so the management can't receive the response from the restore process. So, it resulted in an error.

GuChad369: How do we handle the token bucket now? Should we set different ones for different functions or just use a shared one?

Jubilee101: I would still prefer one bucket per process. Jesper may think otherwise. He'll join us later :)

jesperpedersen: One bucket per process

GuChad369:
This new version modifies the token bucket, adding rate limits to both backup and restore operations, and introduces separate configurations for each rate limit.

Test
- The effect of rate limiting is very easy to observe.
- With the same configuration, backup operations will be slower than restore operations.

jesperpedersen:
A quick review...

There is a `.DS_Store` file in there.

This should only be about backup I/O rate - so `backup_max_rate` should be enough as configuration - it should be documented in `doc/CONFIGURATION.md` at least - maybe a tutorial is in order too.

If there are functions shared between backup and other areas then their signatures need to be changed, and `backup_max_rate` should be passed in as a parameter. Other call-sites should use 0.

I think you are ok with `0` being the default - meaning off - but you should only allocate memory if enabled.

We can circle back to see if there are use-cases for restore in a later pull request.

Also, rebase, please

GuChad369:
- Could you please explain what a `.DS_Store` file is and its purpose? I didn't see it.
- There's no need to pass backup_max_rate as a parameter; it's already available in the configuration and can be accessed from there.

jesperpedersen:
`.DS_Store` is from a Mac...

I mean, if there are functions that are shared between backup and other area then they shouldn't use the configuration

jesperpedersen:
You can see your patch https://github.com/pgmoneta/pgmoneta/pull/227.patch here

GuChad369:
Can we initialize the bucket in the `basebackup_execute` function within `wf_backup.c`, which calls the I/O function? We can then pass it into `pgmoneta_receive_archive_files` and `pgmoneta_receive_archive_stream`, which might be more convenient.

If the value is 0, we simply pass a null ptr.

jesperpedersen:
Yeah, lets see how that looks

GuChad369:
- Update the method of utilizing the token bucket to reduce duplicated code. 
- Add a new function for the token bucket to handle situations where the burst size is small. 
- There is a bug: if we quit during the backup and restore process, although `pgmoneta` will terminate, the process will continue.

doc/CONFIGURATION.md
```diff
@@ -65,6 +65,8 @@ See a [sample](./etc/pgmoneta.conf) configuration for running `pgmoneta` on `loc
 | tls_ca_file | | String | No | Certificate Authority (CA) file for TLS. This file must be owned by either the user running pgmoneta or root.  |
 | libev | `auto` | String | No | Select the [libev](http://software.schmorp.de/pkg/libev.html) backend to use. Valid options: `auto`, `select`, `poll`, `epoll`, `iouring`, `devpoll` and `port` |
 | buffer_size | 65535 | Int | No | The network buffer size (`SO_RCVBUF` and `SO_SNDBUF`) |
+| backup_max_rate | 0 | Int | No | The number of bytes of tokens added every one second to limit the backup rate|
+| restore_max_rate | 0 | Int | No | The number of bytes of tokens added every one second to limit the restore rate|
```
jesperpedersen, line 69: Lets wait with `restore_max_rate` - we need a discussion about that

jesperpedersen (changes requested):

src/include/pgmoneta.h
```diff
@@ -352,6 +355,9 @@ struct configuration
    int number_of_users;          /**< The number of users */
    int number_of_admins;         /**< The number of admins */
 
+   unsigned long backup_max_rate; /**< Number of tokens added to the bucket with each replenishment for backup. */
+   unsigned long restore_max_rate; /**< Number of tokens added to the bucket with each replenishment for restore. */
```
jesperpedersen, line 359: Still here

src/libpgmoneta/configuration.c
```diff
@@ -128,6 +129,9 @@ pgmoneta_init_configuration(void* shm)
    config->log_mode = PGMONETA_LOGGING_MODE_APPEND;
    atomic_init(&config->log_lock, STATE_FREE);
 
+   config->backup_max_rate = 0;
+   config->restore_max_rate = 0;
```
jesperpedersen, line 133: Still here

src/libpgmoneta/configuration.c
```diff
@@ -1169,6 +1173,34 @@ pgmoneta_read_configuration(void* shm, char* filename)
...
+                     {
+                        unknown = true;
+                     }
+                  }
+                  else
+                  {
+                     unknown = true;
+                  }
+               }
+               else if (!strcmp(key, "restore_max_rate"))
```
jesperpedersen, line 1190: Still here

src/libpgmoneta/utils.c
```diff
@@ -1892,6 +1893,19 @@ copy_file(void* arg)
    int saved_errno = -1;
    int permissions = -1;
    struct worker_input* fi = NULL;
+   struct configuration* config = (struct configuration*)shmem;
+   struct token_bucket* bucket = NULL;
+
+   // default is 0
+   if (config->restore_max_rate)
```
jesperpedersen, line 1900: Still here

jesperpedersen (changes requested):

src/libpgmoneta/utils.c
GuChad369: Should we remove the restore functionality for now?

jesperpedersen: Yes, we need a discussion about it

jesperpedersen: Feel free to start this after this PR is in

jesperpedersen:
Merged.

Issue #205, opened 2024-03-05
Port new CLI & JSON functionality from pgagroal


Jubilee101:
Now that #187 is almost done - the remaining task requires some digging into postgres internal, as well as support from pgmoneta_ext side - I'd like to start porting Luca's CLI and JSON functionality from [pgagroal](https://github.com/agroal/pgagroal). This shouldn't be a hard task. I'll if see it needs any tweak to better suit pgmoneta.

- [x] JSON
- [x] CLI

PR #229, opened 2024-03-25
[#205] Porting command line parsing infrastructure

lenhat509:
Porting CLI infrastructure from pgagroal to pgmoneta

### DIFF
--- a/src/libpgmoneta/utils.c
+++ b/src/libpgmoneta/utils.c
@@ -35,6 +35,7 @@
 
 /* system */
 #include <dirent.h>
+#include <err.h>
 #include <errno.h>
 #include <ev.h>
 #include <execinfo.h>
@@ -905,6 +906,35 @@ pgmoneta_set_proc_title(int argc, char** argv, char* s1, char* s2)
                 s2 != NULL ? s2 : "");
 
 #endif
+}
+
+unsigned int
+pgmoneta_version_as_number(unsigned int major, unsigned int minor, unsigned int patch)
+{
+   return (patch % 100)
+          + (minor % 100) * 100
+          + (major % 100) * 10000;
+}
+
+unsigned int
+pgmoneta_version_number(void)
+{
+   return pgmoneta_version_as_number(PGMONETA_MAJOR_VERSION,
+                                     PGMONETA_MINOR_VERSION,
+                                     PGMONETA_PATCH_VERSION);
+}
+
+bool
+pgmoneta_version_ge(unsigned int major, unsigned int minor, unsigned int patch)
+{
+   if (pgmoneta_version_number() >= pgmoneta_version_as_number(major, minor, patch))
+   {
+      return true;
+   }
+   else
+   {
+      return false;
+   }
 }
 
 int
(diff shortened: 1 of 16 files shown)


lenhat509:
Hi @jesperpedersen @Jubilee101, I completed porting the command line parsing feature from pgagroal. Let me know if there is any concern.

src/libpgmoneta/utils.c
```diff
@@ -3003,6 +3004,128 @@ pgmoneta_is_file_archive(char* file_path)
...
+      }
+      return false;
+   }
+
+   parsed->cmd = &command_table[command_index];
+
+   /* Iterate until find an accepted_arg_count that is equal or greater than the typed command arg_count */
+   arg_count = argc - offset;
+   int j;
+   for (j = 0; j < MISC_LENGTH; j++)
```
jesperpedersen, line 3089: Define `int j` in the `for` loop

jesperpedersen: Or move it to the top of the function

src/libpgmoneta/utils.c
```diff
@@ -3003,6 +3004,128 @@ pgmoneta_is_file_archive(char* file_path)
...
+/* Parser for pgmoneta-cli amd pgmoneta-admin commands */
+bool
+parse_command(int argc,
+              char** argv,
+              int offset,
+              struct pgmoneta_parsed_command* parsed,
+              const struct pgmoneta_command command_table[],
+              size_t command_count)
+{
+#define EMPTY_STR(_s) (_s[0] == 0)
```
jesperpedersen, line 3016: Hmm, isn't

```
if (s != NULL && strlen(s) > 0)
```

better ?

src/libpgmoneta/utils.c
```diff
@@ -53,6 +53,7 @@
 #include <sys/time.h>
 #include <sys/stat.h>
 #include <sys/types.h>
+#include <err.h>
```
jesperpedersen, line 56: Sorting

jesperpedersen (changes requested):

src/libpgmoneta/utils.c
lenhat509: Got it. I'll move j to the top since it is used later.

src/libpgmoneta/utils.c
lenhat509: I'll fix that. Thanks.

src/libpgmoneta/utils.c
lenhat509: I think it makes the code a bit cleaner if we use macros. Subcommand is usually empty and we do the check frequently.

jesperpedersen:
You need to update `contrib/shell_comp`

src/libpgmoneta/utils.c
jesperpedersen, line 3016: I think you should move this to `pgmoneta.h`

jesperpedersen (changes requested):

lenhat509:
> You need to update `contrib/shell_comp`

Sorry about missing that. I'll work on it.

src/libpgmoneta/utils.c
lenhat509: Just the macro, not the whole parse_command function, right?

jesperpedersen: Just macro

lenhat509:
Hi @jesperpedersen, I updated `contrib/shell_comp`, fixed a bug in `admin.c`, modified `cli.c` to handle failed commands nicely. In `cli.c`, it can display a helper function for the corresponding failed command instead of showing the entire menu. `usage()` will only be executed if the primary command does not exist. As for `admin.c`, the entire menu will be displayed since it is relatively small.

jesperpedersen:
Please, rebase

lenhat509:
I added back version info for deprecated commands. Initially, I removed this info when I ported the code from pgagroal because I did not see version variables like `PGMONETA_MAJOR_VERSION` in pgmoneta. I just realized that I could define them in CMakeList.txt so I updated the code to show the version from which the commands were outdated (0.11).

jesperpedersen:
You need to update the documentation too...

Do a `git grep pgmoneta-cli` to see the files

lenhat509:
Thanks for letting me know. I updated the doc. Let me know if I miss something else.

jesperpedersen:
Please, rebase

lenhat509:
Hi @jesperpedersen , I have rebased it.

jesperpedersen:
Merged.

Issue #96, opened 2022-10-02
MacOS X support


omerfruk:
**Describe the bug**

Whene I install pgmoneta whit `make` command in macos. I am getting error like this 
>[  0%] Built target man
Consolidate compiler generated dependencies of target pgmoneta
[  2%] Building C object src/CMakeFiles/pgmoneta.dir/libpgmoneta/aes.c.o
/usr/local/pgmoneta/src/libpgmoneta/aes.c:60:28: error: use of undeclared identifier 'DT_DIR'
      if (entry->d_type == DT_DIR)
                           ^
/usr/local/pgmoneta/src/libpgmoneta/aes.c:144:28: error: use of undeclared identifier 'DT_REG'
      if (entry->d_type == DT_REG)
                           ^
/usr/local/pgmoneta/src/libpgmoneta/aes.c:223:28: error: use of undeclared identifier 'DT_DIR'
      if (entry->d_type == DT_DIR)
                           ^
3 errors generated.
make[2]: *** [src/CMakeFiles/pgmoneta.dir/libpgmoneta/aes.c.o] Error 1
make[1]: *** [src/CMakeFiles/pgmoneta.dir/all] Error 2
make: *** [all] Error 2

How can ı solve this problem. thenks

jesperpedersen:
MacOS isn't supported by the build system at the moment.

All necessary changes should be within `src/CMakeLists.txt` using the `Darwin` system identifier. Can you work on that?

jesperpedersen:
You could try the `CMAKE_OSX_SYSROOT` route first, see

https://stackoverflow.com/questions/10165335/can-cmake-specify-the-base-sdk-on-mac-os-x

if that helps... We need the system libraries from the MacOS SDK linked into the pgmoneta library in order for it to work.

2010YOUY01:
MacOS support has been done on a similar project `pgagroal`, you can use this PR as reference if you would like to build and run `pgmoneta` on MacOS X
https://github.com/agroal/pgagroal/pull/351

jesperpedersen:
@HazemRawi Go for it ! See @2010YOUY01's comment - it should be the same

HazemRawi:
I wanted to understand the whole process so I'm playing with CMakeLists.txt of pgagroal & pgmoneta as well as reading some specific pages of the documentation of CMake.
I'll be back soon.

PR #237, opened 2024-03-31
[#96] MAC OS support

GuChad369:
I modified two `CMakeLists.txt` files and the `cli.c` and `utils.c` files to make `pgmoneta` support macOS.

Test
- Running `pgmoneta -?` does not provide information about `pgmoneta`.
- The `add-user` command has been replaced by `user add`.
- The `update-user` command has been replaced by `user edit`.
- When editing the user password, it requires verification three times, which I believe is not normal behavior.
image
- Backup works well
- The restore process is not functioning normally. The log indicates success, but the data cannot be found in the directory.
image
- After stopping the process, it continues to print text in the shell.
image
- There is information printed directly in the shell, not in the log.


jesperpedersen:
Try and get MacOS support enabled in GitHub CI...

You can likely use things from our sister project pgagroal @ https://github.com/agroal/pgagroal/blob/master/.github/workflows/ci.yml

GuChad369:
- The file format will automatically use `uncrustify.sh`. What is the format of the file?
- How do I test the file? Do I have to push it to test it?

jesperpedersen:
Yeah, update ci.yml, and it should be tested

.github/workflows/ci.yml
```diff
@@ -3,58 +3,105 @@ name: CI
 on:
   push:
     branches:
-    - main
+      - main
```
jesperpedersen, line 6: Do we need the indentation ?

.github/workflows/ci.yml
```diff
@@ -3,58 +3,105 @@ name: CI
...
 
 jobs:
-  build:
-
+  build-linux:
     runs-on: ubuntu-latest
-    
+
+    steps:
+      - uses: actions/checkout@v3
```
jesperpedersen, line 16: Lets try with `checkout@4` while we are at it

.github/workflows/ci.yml
```diff
@@ -3,58 +3,105 @@ name: CI
...
-    - name: CLANG/mkdir
-      run: mkdir build
-      working-directory: /home/runner/work/pgmoneta/pgmoneta/
-    - name: CLANG/cmake
-      run: export CC=/usr/bin/clang && cmake -DCMAKE_BUILD_TYPE=Debug ..
-      working-directory: /home/runner/work/pgmoneta/pgmoneta/build/
-    - name: CLANG/make
-      run: make
-      working-directory: /home/runner/work/pgmoneta/pgmoneta/build/
+      - uses: actions/checkout@v3
```
jesperpedersen, line 64: More indentation changes

src/cli.c
```diff
@@ -111,8 +111,9 @@ static int reset(SSL* ssl, int socket);
 static int reload(SSL* ssl, int socket);
 static int retain(SSL* ssl, int socket, char* server, char* backup_id);
 static int expunge(SSL* ssl, int socket, char* server, char* backup_id);
-static int decrypt(SSL* ssl, int socket, char* path);
-static int encrypt(SSL* ssl, int socket, char* path);
+// The encrypt function is a standard library function declared in unistd.h on macOS
+static int decrypt_data(SSL* ssl, int socket, char* path);
+static int encrypt_data(SSL* ssl, int socket, char* path);
```
jesperpedersen, line 116: Why ?

src/libpgmoneta/utils.c
```diff
@@ -898,7 +898,27 @@ pgmoneta_set_proc_title(int argc, char** argv, char* s1, char* s2)
 
    // keep track of how long the title is now
    max_process_title_size = size;
+#elif defined(HAVE_OSX)
+   struct configuration* config;
```
jesperpedersen, line 902: Move to top of function

src/libpgmoneta/utils.c
```diff
@@ -898,7 +898,27 @@ pgmoneta_set_proc_title(int argc, char** argv, char* s1, char* s2)
...
+
+   config = (struct configuration*)shmem;
```
jesperpedersen, line 904: Ditto

src/libpgmoneta/utils.c
```diff
@@ -898,7 +898,27 @@ pgmoneta_set_proc_title(int argc, char** argv, char* s1, char* s2)
...
+
+   // sanity check: if the user does not want to
+   // update the process title, do nothing
+   if (config->update_process_title == UPDATE_PROCESS_TITLE_NEVER)
+   {
+      return;
+   }
+
+   // Compose the new title
+   char title[MAX_PROCESS_TITLE_LENGTH];
```
jesperpedersen, line 914: Ditto

jesperpedersen (changes requested):

src/cli.c
GuChad369: When compiling on macOS, a naming conflict arises with a function in `unistd.h`, resulting in a compilation error.

jesperpedersen: Ok, fair enough

src/libpgmoneta/utils.c
GuChad369: Should I move the declaration to the top of the function, or should I replace the existing condition with `HAVE_LINUX`?

jesperpedersen: Move the common structures outside of the `#ifdef`'s

GuChad369: However, there is also an `else` condition in which these elements are not utilized.

jesperpedersen: I think in this case it would be ok

GuChad369:
This is the completed version for macOS.

jesperpedersen:
Merged.

Issue #241, opened 2024-04-04
Pthread not found and linked automatically on RHEL8


Jubilee101:
We need to add a CMake find module for that.

PR #242, opened 2024-04-04
[#241] Add CMake rules to find and link libpthread

Jubilee101:


src/libpgmoneta/utils.c
```diff
@@ -826,7 +825,9 @@ pgmoneta_set_proc_title(int argc, char** argv, char* s1, char* s2)
    {
       return;
    }
+
 #ifdef HAVE_LINUX
+   char title[MAX_PROCESS_TITLE_LENGTH];
```
jesperpedersen, line 830: Lets move this back, and do a `memset()` on it after the `config` assignment

jesperpedersen (changes requested):

src/libpgmoneta/utils.c
Jubilee101: Oh didn't see that, right away

Issue #239, opened 2024-04-01
Lack of Feedback for Illegal Passwords in Setting Master Key for pgmoneta-admin


l32zhao:
# Problem

I discovered that when attempting to set an illegal password for the **master key** of pgmoneta-admin, there is no indication or feedback provided.

# Solution

To address this issue, it would be beneficial to include informative messaging, such as specifying the acceptable length of passwords, when setting the master key for pgmoneta-admin.

l32zhao:
Here is a screenshot of this problem:
![1711045362186](https://github.com/pgmoneta/pgmoneta/assets/29018475/fa78e2ee-ceff-424e-ae54-84e342ca4a61)

Jubilee101:
Hi there, would you like to work on this? Maybe just print out the reason why the validation fails. It should be quite useful, thanks for bringing this up.

PR #244, opened 2024-04-06
[#239] Add warning for illegal key while setting master-key manually

l32zhao:
This pull request enhances the master key validation process in the `master_key` function by adding warning messages.

Changes:

1. Modified the `is_valid_key` function to check for the following conditions:
   - Key length hint
   - Key cannot contain non-ASCII characters

Optional things could be done in the future:
   - Max length of key, like 64 characters
   - Key must contain at least one digit, one uppercase letter, one lowercase letter, and one symbol

### DIFF
--- a/src/admin.c
+++ b/src/admin.c
@@ -478,6 +478,7 @@ is_valid_key(char* key)
 
    if (strlen(key) < 8)
    {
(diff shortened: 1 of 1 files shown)


jesperpedersen:
Merged.

PR #252, opened 2024-04-08
[#187] Receive WAL end pos of backup from server

Jubilee101:
I was investigating base backup on the server side and I realized I was ignoring starting/ending WAL position server sends us the whole time. So no need to look into the backup_label file for those info now. 

Also exposed the ending pos to prometheus, and fixed some minor bugs there as well.


jesperpedersen:
Merged.

Issue #253, opened 2024-04-09
WAL segment lost due to replication slot not reserving segments on creation


Jubilee101:
If replication slot was created with no RESERVE_WAL flag set, then there's a tiny chance that before repl client connects to the server and activate the replication slot in wal process, early segments got deleted by the server.

PR #254, opened 2024-04-09
[#253] create replication slot with RESERVE_WAL set

Jubilee101:


jesperpedersen:
Merged.

Issue #265, opened 2024-04-13
Initial manual 


jesperpedersen:
This need to be under `doc/manual/` and it could likely be based on `pandoc` and `markdown` files to compile into `.pdf` and `.html`

GuChad369:
- Is this manual about setting up the environment?
- What is the structure of the directory?

jesperpedersen:
We could look at doing something similar to https://github.com/EnterpriseDB/barman/tree/master/doc/manual just with `cmake`.

Better ideas are welcomed !

GuChad369:
- Should we integrate CMake directly into our project, or should we keep it separate in the manual directory? 
- This way, if a user wants to generate it, they can do so; otherwise, we won't generate it by default.

jesperpedersen:
I don't understand - it should be part of the standard `cd build/ && cmake .. && make` setup

GuChad369:
Okay, thanks. I'll work on it.

PR #273, opened 2024-04-18
[#265] initial manual

GuChad369:
- Add a new initial `/manual` directory.
- Modify the `CMake` file to enable the use of `pandoc` and `LaTeX` for generating `.html` and `.pdf` files from `Markdown`.
- Update the Git `.workflows`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,6 +37,10 @@ jobs:
       run: sudo apt install -y clang
     - name: Install cJSON
       run: sudo apt install -y libcjson-dev
+    - name: Install pandoc
+      run: sudo apt install pandoc
+    - name: Install pdflatex
+      run: sudo apt install texlive-latex-extra
     - name: GCC/mkdir
       run: mkdir build
       working-directory: /home/runner/work/pgmoneta/pgmoneta/
@@ -148,6 +152,24 @@ jobs:
       run: brew install llvm
     - name: Install cJSON
       run: brew install cjson
+    - name: Install pandoc
+      run: brew install pandoc
+    - name: Install pdflatex
+      run: brew install --cask basictex
+    - name: Refresh env
+      run: eval "$(/usr/libexec/path_helper)"
+    - name: Locate pdflatex and update PATH
+      run: |
+        PDFLATEX_PATH=$(find /usr/local/texlive -name pdflatex -print -quit)
+        if [ -n "$PDFLATEX_PATH" ]; then
+          PDFLATEX_DIR=$(dirname "$PDFLATEX_PATH")
+          echo "PDFLATEX_DIR=$PDFLATEX_DIR" >> $GITHUB_ENV
+          echo "PATH=$PDFLATEX_DIR:$PATH" >> $GITHUB_ENV
+        else
+          echo "pdflatex could not be found"
+          exit 1
+        fi
+      shell: bash
     - name: Install PostgreSQL
       run: brew install postgresql
     - name: GCC/mkdir
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,6 +57,38 @@ if(NOT COMPILER_SUPPORTS_C17)
   message(FATAL_ERROR "The compiler ${CMAKE_C_COMPILER} has no C17 support. Please use a different C compiler.")
 endif()
 
+if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+    # Homebrew ships libarchive keg only, include dirs have to be set manually
+    execute_process(
+      COMMAND brew --prefix libarchive
+      OUTPUT_VARIABLE LIBARCHIVE_PREFIX
+      OUTPUT_STRIP_TRAILING_WHITESPACE
+      COMMAND_ERROR_IS_FATAL ANY
+    )
+    set(LibArchive_INCLUDE_DIR "${LIBARCHIVE_PREFIX}/include")
+
+    execute_process(
+        COMMAND brew --prefix pandoc
+        OUTPUT_VARIABLE PANDOC_PREFIX
+        OUTPUT_STRIP_TRAILING_WHITESPACE
+        COMMAND_ERROR_IS_FATAL ANY
+    )
+endif()
+
+find_package(Pandoc)
+if (PANDOC_FOUND)
+  message(STATUS "pandoc found")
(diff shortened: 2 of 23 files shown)


jesperpedersen:
```
-- System is Linux
-- pandoc found
CMake Error at cmake/FindPdflatex.cmake:14 (execute_process):
  execute_process failed command indexes:

    1: "Child return code: 1"

Call Stack (most recent call first):
  CMakeLists.txt:85 (find_package)
```

```
which pdflatex
/usr/bin/pdflatex
```

```
pdflatex --version
pdfTeX 3.141592653-2.6-1.40.25 (TeX Live 2023)
kpathsea version 6.3.5
Copyright 2023 Han The Thanh (pdfTeX) et al.
There is NO warranty.  Redistribution of this software is
covered by the terms of both the pdfTeX copyright and
the Lesser GNU General Public License.
For more information about these matters, see the file
named COPYING and the pdfTeX source.
Primary author of pdfTeX: Han The Thanh (pdfTeX) et al.
Compiled with libpng 1.6.37; using libpng 1.6.37
Compiled with zlib 1.2.13; using zlib 1.2.13
Compiled with xpdf version 4.04
```

jesperpedersen:
You need to update `README.md` and `pgmoneta.spec` with the files

jesperpedersen:
Make sure that you verify on Fedora / RHEL 8 / RHEL 9

GuChad369:
Could you provide some guidance on the pgmoneta.spec file? I am not familiar with it.

jesperpedersen:
You need to make

* A directory (see around https://github.com/pgmoneta/pgmoneta/blob/main/pgmoneta.spec#L33)
* Copy the .pdf and .html in (see around https://github.com/pgmoneta/pgmoneta/blob/main/pgmoneta.spec#L71)
* Define them (see around https://github.com/pgmoneta/pgmoneta/blob/main/pgmoneta.spec#L122)

Take a look at https://github.com/pgmoneta/pgmoneta/blob/main/doc/RPM.md

GuChad369:
I've updated the README and `pgmoneta.spec` files; the RPM was successfully tested on Fedora.

jesperpedersen:
cmake is ok, but

```
[ 98%] Generating HTML documentation
[100%] Generating PDF documentation
Error producing PDF.
! LaTeX Error: File `footnote.sty' not found.

Type X to quit or <RETURN> to proceed,
or enter new name. (Default extension: sty)

Enter file name: 
! Emergency stop.
<read *> 
         
l.86 \makesavenoteenv

make[2]: *** [src/CMakeFiles/all_docs.dir/build.make:100: doc/pgmoneta-manual.en.pdf] Error 43
make[1]: *** [CMakeFiles/Makefile2:254: src/CMakeFiles/all_docs.dir/all] Error 2
make: *** [Makefile:156: all] Error 2
```

jesperpedersen:
```
dnf install -y texlive-latex-extra
Last metadata expiration check: 0:20:02 ago on Fri 19 Apr 2024 10:14:28 PM EDT.
No match for argument: texlive-latex-extra
Error: Unable to find a match: texlive-latex-extra
```

```
rpm -qa | grep texlive-scheme-basic
texlive-scheme-basic-svn54191-69.fc39.noarch
```

jesperpedersen:
Its Fedora 39

GuChad369:
We're facing a problem with `pdflatex` as it doesn't have a specific library. To use `pdflatex`, we must download additional libraries like `TeX Live`, but each system offers different libraries for LaTeX. Our project also requires the `footnote.sty` package and different libraries may not offer the same method to verify its availability. Should we focus on solving this issue only for the Fedora system?

jesperpedersen:
The `README.md` should cover Fedora, the rest should be documented in https://github.com/pgmoneta/pgmoneta/blob/main/doc/DISTRIBUTIONS.md

jesperpedersen:
The key package for Fedora is `texlive-mdwtools`

GuChad369:
I used two other ways to install the related libraries. `texlive-mdwtools` does not include `pdflatex`.

jesperpedersen:
Yeah, I know - but we need it for the style

GuChad369:
Okay, let’s try it.

GuChad369:
- I've restructured the manual without splitting it into two separate documents, as I noticed duplicate content for both users and developers. Separating them would result in excessive repetition. 
- We can enhance the content further at a later stage.

jesperpedersen:
There are some formatting issues, especially in "Configuration" chapter.

And, we don't need "Architecture" under introduction that is an advanced developer chapter - so close to last.

Take a step back, and think about what a new user of pgmoneta would like to know about going from install to their first backup

doc/manual/00-head.en.md
```diff
@@ -0,0 +1,3 @@
+% pgmoneta Manual
```
jesperpedersen, line 1: We don't need "Manual"

doc/manual/00-head.en.md
```diff
@@ -0,0 +1,3 @@
...
+% PostgreSQL
```
jesperpedersen, line 2: Remove

doc/manual/00-head.en.md
```diff
@@ -0,0 +1,3 @@
...
+% April 18, 2024 (0.10.1)
```
jesperpedersen, line 3: Remove

doc/manual/01-introduction.en.md
```diff
@@ -0,0 +1,170 @@
+\newpage
+
+# Introduction
+
+[**pgmoneta**](https://github.com/pgmoneta/pgmoneta) is a backup / restore solution for [PostgreSQL](https://www.postgresql.org).
+
+**pgmoneta** is named after the Roman Goddess of Memory.
+
```
jesperpedersen, line 8: We need something something about RPO and RTO in the Introduction, but we could take that later

jesperpedersen (changes requested):

GuChad369:
I made some formatting changes. While `pandoc`'s default compilation can't handle all `Markdown` formatting, if we want a perfect HTML format, we must add a `CSS` file when `pandoc` compiles the HTML.

jesperpedersen:
We can likely find ways to improve the layout, colors, ...

jesperpedersen:
Rebased, and merged.

PR #240, opened 2024-04-03
Prometheus metric for backup version shows major and minor version

iuryr:
This pull request implements a change to the prometheus metric `pgmoneta_backup_server_version` in order to show major and minor versions separately.

- `src/libpgmoneta/info.c` changes populates `backup->minor_version` with pertinent information
- `src/libpgmoneta/prometheus.c` changes adjusts the display of the metric in order to show aforementioned data and also changes documentation displayed in `home_page` function

PR #275, opened 2024-04-18
Fix unclosed td tag in prometheus metric documentation

iuryr:
Fix unclosed td tag in prometheus metric documentation

For discussion see: https://github.com/pgmoneta/pgmoneta/pull/240


jesperpedersen:
Merged.

Issue #276, opened 2024-04-21
Support WAL shipping to remote storage engines


HazemRawi:
Currently WAL segments shipped to local directories only so, I want to expand this functionality to include shipping WAL segments to various storage engines.

- [ ]  SSH storage engine

- [ ]  Azure storage engine

- [ ]  S3 storage engine

PR #277, opened 2024-04-21
[#276]Support WAL shipping to ssh storage engine

HazemRawi:


Hazem-404:
> 

Very Good Code Hazem !

src/include/storage.h
```diff
@@ -63,6 +69,26 @@ pgmoneta_storage_create_s3(void);
...
 pgmoneta_storage_create_azure(void);
 
+/**
+ * Open WAL shipping file in remote ssh server
+ * @param srv The server index
+ * @param filename WAL file name
+ * @param segsize WAL segment size
+ * @return sftp_file WAL streaming file
+ */
+sftp_file
```
jesperpedersen, line 79: Use an out parameter instead, and `int` as the return type with the status of the operation

src/include/utils.h
```diff
@@ -861,6 +861,16 @@ pgmoneta_permission_recursive(char* d);
...
 pgmoneta_permission(char* e, int user, int group, int all);
 
+/**
+ * Calculate mode
+ * @param user The user (0; nothing, 4; read, 6; read/write)
+ * @param group The group (0; nothing, 4; read, 6; read/write)
+ * @param all All (0; nothing, 4; read, 6; read/write)
+ * @return The mode
+ */
+mode_t
```
jesperpedersen, line 871: Same here

jesperpedersen (changes requested):

jesperpedersen:
There are some formatting changes - use `Uncrustify_d-0.77.1_f`

jesperpedersen:
Generally, I think this is going in the right direction... but I need to take a closer look

GuChad369:
When GitHub runs the environment for macOS on this PR, it uses `/opt` as the global directory, whereas normally it should be `/usr`, which would already have the `ssh` library installed. I'm not sure why this change occurred. However, you can modify the `.github/workflows/ci.yml` file to install the library, or manually locate the library to ensure it works on macOS.

HazemRawi:
@GuChad369 Thanks for your help! 
Jesper handled this problem

HazemRawi:
I tested both workflows:

   -When the WAL segment file already exists.
   -When the WAL segment file needs to be created.

Everything works fine except that `sftp_wal_prepare()` takes about 12 minutes to execute. I suspect this delay might be due to the server(E2 server from Google cloud).
Do you have any ideas on how to improve this?

jesperpedersen:
Did you test it with a SSH server ? Or on your network ?

HazemRawi:
> Did you test it with a SSH server ? Or on your network ?

I test it using SSH server as described on `ssh.md`

>I suspect this delay might be due to the server

I mean the server doesn't come with a high system specification, As most of the time taken by `sftp_write() `.

jesperpedersen:
Ok, understood

jesperpedersen:
Rebased, and merged.

Issue #281, opened 2024-04-27
Manual: User guide


jesperpedersen:
Implement improvements to the user guide

jesperpedersen:
Ideally we should have some include functionality, like https://github.com/steindani/pandoc-include

Or, should we move to AsciiDoc ?

jesperpedersen:
We need a better template for the generated documentation,

https://lornajane.net/posts/2023/generating-a-nice-looking-pdf-with-pandoc
https://github.com/Wandmalfarbe/pandoc-latex-template

jesperpedersen:
@GuChad369 Cc:

GuChad369:
Do we need to specify the installation process for the Eisvogel package, or is it sufficient to mention that the package is required? Several steps are involved in adding the package for use.

GuChad369:
> Ideally we should have some include functionality, like https://github.com/steindani/pandoc-include
> 
> Or, should we move to AsciiDoc ?

I'm unsure which option is preferable. The first requires using a script to facilitate the inclusion of components. We would need to add the script to `doc/manual/` directory. 

On the other hand, if we opt for AsciiDoc, `pandoc` cannot convert AsciiDoc directly into PDF.

jesperpedersen:
Ok, I mentioned AsciiDoc because it has file inclusion built-in, but building a PDF is more important

GuChad369:
> We have to figure out how the needed components are installed - otherwise we will have to supply those in-tree under `doc/support`

I have successfully used the Eisvogel package to generate a better PDF, but it may require extra effort to install the necessary components.

jesperpedersen:
We need to test with RHEL8, RHEL9, FreeBSD and when building a RPM so just keep that in mind

GuChad369:
The problem is that installing the full TeX Live package directly might cause a timeout on GitHub due to the large size of the package. 

Alternatively, specifying only the needed packages could require installing them one by one. 

Which approach do you think would be better?

jesperpedersen:
If you see https://github.com/pgmoneta/pgmoneta/commit/ae9764c93f03adceff969ab7d32d8e371116bc73 - it is an example of a good "include" (`AUTHORS`)

jesperpedersen:
We have about a month to try - lets start with the simplest way...

GuChad369:
> We have about a month to try - lets start with the simplest way...

Indeed, for the manual, it's a long-term project. We will need to continuously improve it over time.

[...]

PR #283, opened 2024-04-29
[#281] Add pandoc-latex-template

GuChad369:
- Fix an uninitialized error in `se_ssh.c` at line 872.
- Rewrite the manual's reference for easier management.
- Add a new template to improve PDF formatting.


GuChad369:
Is there a more efficient method to test the workflow? Currently, the installation of packages requires many additional steps, making the process cumbersome. The current method of submitting tests is inefficient

jesperpedersen:
```
-- Found Pandoc: /usr/bin/pandoc  
CMake Error at cmake/FindPandoc.cmake:26 (message):
  eisvogel template not found at
  /home/jpedersen/.local/share/pandoc/templates/eisvogel.latex.  Please ensure the
  template is correctly installed.
Call Stack (most recent call first):
  CMakeLists.txt:78 (find_package)
```

```
pandoc-2.0.6-6.el8_6.x86_64
pandoc-2.14.0.3-17.el9.x86_64
pandoc-3.1.3-29.fc39.x86_64
```

jesperpedersen:
So, we are dealing with different versions... and a hard installation process.

Could be an idea to look through https://github.com/search?q=pandoc+template&type=repositories&s=stars&o=desc to see if there is something easier.

At least hard-coding pandoc to a specific version won't work, unless we disable manual generation for specific pandoc versions - has to work on Fedora...

GuChad369:
- To function properly, the Eisvogel template must install `eisvogel.latex` in the specific directory.
- I've added a check to ensure all dependencies are installed before using the tool. Instructions for installing the Eisvogel template can be found in DEVELOPER.md.
- I am trying to run the action locally to ensure the GitHub workflow functions perfectly before submission.

jesperpedersen:
I mean, maybe `pandoc ... --template=local/path/to/file ...` could work.

Otherwise, we make generating the manual optional such that build doesn't fail

GuChad369:
The [pandoc-include](https://github.com/steindani/pandoc-include) plugin isn't working; there are some errors when testing it. An alternative method to perform the include function would be to write a script that combines the Markdown files first, and then use `pandoc` to generate the output from the `combined md`. However, I think this approach isn't ideal as it complicates the process.

jesperpedersen:
Ok, but we have to find a way where we can inline the tutorials into the user guide based on everything from `doc/tutorial/`... maybe it is just something the commit should do...

Also, generating the manuals should be optional, so remove the FATAL, and add a check if they should be generated.

I think it is ok to split the `dnf` commands into two because some people won't generate them...

Alternative, look at how other projects are doing it, like https://github.com/EnterpriseDB/barman/tree/master/doc/manual

GuChad369:
> Ok, but we have to find a way where we can inline the tutorials into the user guide based on everything from `doc/tutorial/`... maybe it is just something the commit should do...

Is it a practical approach to include the `doc/tutorial/` files in our `UMDS` directory for direct integration into our user guide? 

> Also, generating the manuals should be optional, so remove the FATAL, and add a check if they should be generated.

We can define a global variable to conditionally make this feature optional.

> Alternative, look at how other projects are doing it, like https://github.com/EnterpriseDB/barman/tree/master/doc/manual

This project simply extracts the generation process from the overall project. If users need to generate, the project allows them to enter the directory and use `make` to generate the PDF. Should we also adopt this approach? 

Currently, my idea is to automatically check all the dependencies required for generation. If the check is successful, we generate the file; otherwise, we skip it and proceed with the normal operations.

jesperpedersen:
> Is it a practical approach to include the doc/tutorial/ files in our UMDS directory for direct integration into our user guide?

No, we shouldn't have 2 copies of each file - we need to figure out an include mechanism...

> We can define a global variable to conditionally make this feature optional.

Yes

> Should we also adopt this approach?

No, the global variable is enough - the instructions for building the manuals should be in DEVELOPERS.md if people want to use them

GuChad369:
- Integrate the tutorial into the generation process.
- Make generation optional.
- Provide an introduction for developers on how to implement the generation process.

jesperpedersen:
Rebased, and merged.

Issue #291, opened 2024-05-11
Security: Create functions for CRC32x checksums for a file


jesperpedersen:
Create functions

```
int
pgmoneta_create_crc32c_file(char* path, char** crc);

int
pgmoneta_create_crc32c_buffer(void* buffer, size_t size, char** crc);
```

This need to be done with AVX instructions

shikharish:
I would like to work on this.

jesperpedersen:
You likely need to stick to AVX2 - see https://en.wikipedia.org/wiki/Advanced_Vector_Extensions - or SSE...

I would just begin with a basic implementation, and then we can add compiler switches like `#ifdef HAVE_AVX2` to have a more optimized version.

There are a lot of articles around it, like

* https://news.ycombinator.com/item?id=32305900
* https://github.com/topics/crc32?l=c

Remember you have to develop this code yourself...

shikharish:
@jesperpedersen There is a CRC32C implementation already present in `hashmap.c`. Do we want to implement a similar function here? It uses SSE4.2 instructions.

PR #292, opened 2024-05-12
[#266] Add functions for SHA cheksums for a file

### LINKED ISSUE #266: Security: Create functions for SHA checksums for a file
Create functions

```
int
pgmoneta_create_sha256_file(char* path, char** sha);

int
pgmoneta_create_sha384_file(char* path, char** sha);

int
pgmoneta_create_sha512_file(char* path, char** sha);
```

using the OpenSSL API

shikharish:
Closes #266

### DIFF
--- a/src/include/security.h
+++ b/src/include/security.h
@@ -96,7 +96,25 @@ pgmoneta_tls_valid(void);
  * @return 0 upon success, otherwise 1
  */
 int
-pgmoneta_generate_file_sha256_hash(char* filename, char** sha256);
+pgmoneta_create_sha256_file(char* filename, char** sha256);
+
+/**
+ * Generate SHA384 for a file
+ * @param filename The file path
+ * @param sha384 The hash value
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_create_sha384_file(char* filename, char** sha384);
+
+/**
+ * Generate SHA512 for a file
+ * @param filename The file path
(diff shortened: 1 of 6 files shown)


src/include/security.h
```diff
@@ -89,14 +89,42 @@ pgmoneta_get_master_key(char** masterkey);
...
 
+/**
+ * Common function for generating hash for a file
+ * @param filename The file path
+ * @param hash The hash value
+ * @param md_name The md type(SHA256, SHA384, SHA512)
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_create_hash_file(char* filename, char** hash, const char* md);
```
jesperpedersen, line 100: I think this should be a `static` function inside `security.c` called `create_hash_file()`.

And, it should be `char* filename, char* algorithm, char** hash)` as the signature, first input then output after

jesperpedersen (changes requested):

jesperpedersen:
Only a few minor changes found in my first pass - good job

jesperpedersen:
Merged.

jesperpedersen:
@shikharish Do you want to tackle #291 next ? Its a lot harder...

shikharish:
@jesperpedersen Yes, I will give it a try! If you can share any related resources, that would be really helpful.

jesperpedersen:
Just make a comment on the issue that you will take it

Issue #278, opened 2024-04-24
Replace cJSON with an optimized native version


jesperpedersen:
It should be able to handle 10m entries for manifest, and support the CLI output

PR #293, opened 2024-05-12
[#278] Native JSON support

Jubilee101:
A native json libray, supports:
1. Multiple data types: int64, string, float, corresponding arrays, as well as nested json objects.
2. Basic operations to init, search, insert, update json fields
3. Shallow parsing, parse the json file to a json item. All nested json objects will be ignored for now (we don't need them anyway)
4. Navigating. Provided with an array of keys as path, navigate to where the json array or item starts. This uses bounded memory, no need to load the whole json file into memory at a time.
5. Iterative item array parsing. Parse items within one at a time. This uses bounded memory, no need to load the whole json file into memory at a time.
6. Format printing


jesperpedersen:
Merged.

Issue #267, opened 2024-04-13
CLI: Verify


jesperpedersen:
Create a `pgmoneta-cli verify <server> <id>` command that verifies the manifest against the files on-disk

jesperpedersen:
This could be expanded to run in the background as well with a default of `0` aka off

iuryr:
Hi @jesperpedersen. This message is just to say that I began working on this issue.

jesperpedersen:
@iuryr #266 have been merged

iuryr:
Jesper,

just to clarify, this feature aims to verify the checksum of each file in the manifest file generated by pgmoneta against the files 'generated' by pgmoneta backup, right?

jesperpedersen:
@iuryr Yes, you have to go through the `backup_manifest` file, and verify that the on-disk files still have the same SHA-256 value

jesperpedersen:
@iuryr Still ok on this ?

iuryr:
Hi @jesperpedersen ! Still ok. Took me more than I expected to understand the communication between pgmoneta and pgmoneta-cli processes. Now I'm working on the feature itself. Maybe I can submit a PR by the end of next week. Let me know if this issue is blocking somebody

jesperpedersen:
@iuryr No, you are ok - you can always update the issue or discussion with your thoughts and progress

iuryr:
> This could be expanded to run in the background as well with a default of `0` aka off

Hi @jesperpedersen , in my implementation I'm running the checking in a fork() of pgmoneta main process. Is this what you mean by "run in the background"? Also, can you explain what you mean my a default of `0`?

Finally, a sucessfull check (meaning the on-disk files checksums match the checksum reported in backup_manifest should make pgmoneta-cli print a message to the terminal? Or should it print something in case it does not match? Or another option?

jesperpedersen:
Yes, you should run it in a `fork()`, and you can use the worker pool as well.

Just do the explicit `verify` command first, then we can look at doing automatic verification in the background - you may want to follow along on https://github.com/pgmoneta/pgmoneta/discussions/312

Yes, there should be some sort of outout - also in JSON format. All files that doesn't match must be listed.

Hope this is helpful

Jubilee101:
[...]

PR #314, opened 2024-06-16
[#267] CLI Verify - Minimal viable implementation

iuryr:
Hello,

this draft pull request contains a minimal viable implementation of `pgmoneta-cli` feature called `verify`. The feature compares backup files checksums reported in backup_manifest to checksums computed from files on disk.

As of now, the command returns 0 in case reported checksums match with computed ones, 1 if an error occur, 2 if no errors occur but there is at least one file that computed checksum does not match the reported one. 

The command receives two inputs: a server name and a backup_id.
Usage: `pgmoneta-cli -c <config_file> verify <server_name> <backup_id>`

I have some questions for further development:
- Do I need to call `shutdown_ports()` after the `fork()` is created?
- Is it a good idea to code a pgmoneta_get_server_backup_identifier_backup_manifest in src/libpgmoneta/utils.c ?
- Should more log messages be issued?
- Should I already extend the function to consider other checksum-algorithms? For the MVP I'm assuming SHA-256.

@jesperpedersen Any guidance on references to JSON output and use of workers?

Appreciate any comments. Thank you

---
TODO

- [ ] Output to JSON
- [x] help_verify()
- [ ] Use of workers
- [ ] crustify

### DIFF
--- a/src/libpgmoneta/management.c
+++ b/src/libpgmoneta/management.c
@@ -34,6 +34,7 @@
 #include <network.h>
 #include <management.h>
 #include <stdint.h>
+#include <manifest.h>
 #include <utils.h>
 
 /* system */
@@ -82,6 +83,7 @@ static struct json* read_details_json(SSL* ssl, int socket);
 static struct json* read_list_backup_json(SSL* ssl, int socket, char* server);
 static struct json* read_delete_json(SSL* ssl, int socket, char* server);
 static struct json* read_info_json(SSL* ssl, int socket);
+static struct json* read_verify_json(SSL* ssl, int socket);
 
 int
 pgmoneta_management_read_header(int socket, signed char* id)
@@ -134,6 +136,11 @@ pgmoneta_management_read_payload(int socket, signed char id, char** payload_s1, 
       case MANAGEMENT_INFO:
          read_string("pgmoneta_management_read_payload", socket, payload_s1);
          read_string("pgmoneta_management_read_payload", socket, payload_s2);
+         break;
+      case MANAGEMENT_VERIFY:
+         read_string("pgmoneta_management_read_payload", socket, payload_s1);
+         read_string("pgmoneta_management_read_payload", socket, payload_s2);
+         read_string("pgmoneta_management_read_payload", socket, payload_s3);
          break;
       case MANAGEMENT_STOP:
       case MANAGEMENT_STATUS:
@@ -1282,6 +1289,35 @@ pgmoneta_management_encrypt(SSL* ssl, int socket, char* path)
    {
       goto error;
    }
+
+   return 0;
+
+error:
+
+   return 1;
+}
+
+int
+pgmoneta_management_verify_data(SSL* ssl, int socket, char* server, char *backup_id, char* option)
+{
+   if (write_header(ssl, socket, MANAGEMENT_VERIFY))
+   {
+      pgmoneta_log_warn("pgmoneta_management_verify_data: write: %d", socket);
+      errno = 0;
+      goto error;
+   }
+	if (write_string("pgmoneta_management_verify_data", socket, server))
+	{
+		goto error;
+	}
+	if (write_string("pgmoneta_management_verify_data", socket, backup_id))
+	{
+		goto error;
+	}
+	if (write_string("pgmoneta_management_verify_data", socket, option))
+	{
+		goto error;
+	}
 
    return 0;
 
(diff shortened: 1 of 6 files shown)


jesperpedersen:
> Do I need to call shutdown_ports() after the fork() is created?

You don't need, so yes.

> Is it a good idea to code a pgmoneta_get_server_backup_identifier_backup_manifest in src/libpgmoneta/utils.c ?

It is better in `manifest.h|c`

 > Should more log messages be issued?

Return value should only be 0 or 1 -- any error on a file should be reported as an error, and the files in question should be listed.

 > Should I already extend the function to consider other checksum-algorithms? For the MVP I'm assuming SHA-256.

Make sure that you rebase against `main` as we are changing how manifests works.

You can maybe ideas from the `info` command commit @ https://github.com/pgmoneta/pgmoneta/commit/170052e221df577c20fa1ea8ac0f0eea74d4eeb2

jesperpedersen:
See `src/libpgmoneta/wf_restore.c` for usage of workers, and you need to consider all the hash algorithms now

iuryr:
> See https://github.com/pgmoneta/pgmoneta/blob/main/doc/DEVELOPERS.md#basic-git-guide for help

Thanks, Jesper! Using the sync functionality in Github UI always gets me in trouble. I'll set the original pgmoneta repo as upstream and do the fetch and rebase via git cli, as advised in the document.

jesperpedersen:
You force push on your branch again

iuryr:
Sorry! Inadvertently closed the PR trying to fix settings for local and remote branches.

iuryr:
Hi @jesperpedersen 

In this last commit (still draft), I made `pgmoneta_verify_data` generate a JSON Object of the type array that can the printed using `pgmoneta_json_print` to `stdout`.

For example:
```json
[...,
{
  "Path": "global/pg_control",
  "Check": "Success",
  "Actual Checksum": "7b7e9232a4c044247f2e217c66494cc51d5383ba16fd83a9b3ed8e8f2e2b7765",
  "Manifest Checksum": "7b7e9232a4c044247f2e217c66494cc51d5383ba16fd83a9b3ed8e8f2e2b7765"
 },
 {
  "Path": "pg_wal/archive_status/000000010000000000000069.done",
  "Check": "Fail",
  "Actual Checksum": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
  "Manifest Checksum": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b856"
 }]
 ```

Is this a reasonable approach? Or should I generate a file in disk which would store the output of the `verify` command?

**Comment by jesperpedersen** [comment]
Yes, this is correct.

Maybe you want an option to only print "Fail" objects...

**Comment by jesperpedersen** [comment]
"Success" should be the "default" so we are not that interested in those...

So, the option should be `fail (default) | all` and the return code should reflect if there are failed entries

**Comment by iuryr** [comment]
Implemented `pgmoneta-cli verify <server> <id> [FAIL|all]`.

As of now, the forked process is the one printing to stdout, so I'm having trouble piping the output to a file, for example `pgmoneta-cli verify primary latest > output_file`;

I wonder if I have to make the forked process somehow pass the json object that stores the output to the pgmoneta-cli process via some of the `pgmoneta_write_*` functions... Any advice on how to proceed?

**Comment by jesperpedersen** [comment]
I think it is best to follow the `info` command through `main.c` via the `MANAGEMENT_INFO` and see how it is done. It should be very similar in setup. You have to send the exit code, and the failed / all files across the wire and print them on the CLI side of things.

Remember to rebase, and to run `uncrustify`...

**Comment by iuryr** [comment]
Please do not consider the last commit, it is wrong. I'm trying to upload my code as is right now to report problems I'm having

**Comment by iuryr** [comment]
@jesperpedersen and @Jubilee101 , I'll try to report the situation the PR is in right now.

# Not yet started:

- [ ] Use of workers
- [ ] Support for different checksum algorithms

# Started but not yet completed
Communication between processes pgmoneta/main and pgmoneta/verify: `pgmoneta_verify_data` in `manifest.c` is the function that runs inside pgmoneta/verify process and it aims to populate the array of strings `output`. 
`pgmoneta_management_write_verify_data` in `management.c` is the function responsible to write strings contained in `output` to the `pgmoneta-cli` process that will be responsible for "assembling" a json output and print it to stdout.

Right now I'm having trouble sending these strings from one process to the other. Since pgmoneta_veridy_data runs inside a loop, I need to debug it. My hunch is that freeing/not freeing some of the strings is causing the trouble.

So this is where I am now. I'm not even sure if I have a question to ask. Just need some more time to figure this out. Please let me know if the PR is blocking development for other contributors :/

**Comment by jesperpedersen** [comment]
Please, rebase your code, and run `uncrustify` such that it is easier to review

**Comment by iuryr** [comment]
> Please, rebase your code, and run `uncrustify` such that it is easier to review

I will try to do this still today. uncrustify was failing to format the code. I could not solve it before leaving for school

File: src/libpgmoneta/management.c [hunk 1]
```diff
@@ -3409,3 +3445,301 @@ print_info_json(struct json* json)
...
+	if (backups[backup_index]->valid == VALID_FALSE)
+	{
+		pgmoneta_log_error("Verify: Backup is not valid");
+		goto error;
+	}
+
+
+	//check if backup_manifest exists and is a regular file
+	backup_manifest = pgmoneta_get_server_backup_identifier_data(srv, backups[backup_index]->label);
+	backup_manifest = pgmoneta_append(backup_manifest, "backup_manifest");
```
Jubilee101, line 3668: Note that we now have our own manifest, which locates in the same dir where data/ and tablespaces are. The original manifest is now subject to encryption and compression so you cannot read it directly. Our version (the name is backup.manifest) is in csv format, which we also have support for, so you can read it relatively more easily. Check out [wf_manifest](https://github.com/pgmoneta/pgmoneta/blob/d9eaf72529d6c40c992208a9c61b9eb2a055b3e9/src/libpgmoneta/wf_manifest.c#L69) and [pgmoneta_compare_manifests](https://github.com/pgmoneta/pgmoneta/blob/d9eaf72529d6c40c992208a9c61b9eb2a055b3e9/src/libpgmoneta/manifest.c#L128) for details.

src/main.c
```diff
@@ -1348,7 +1349,72 @@ accept_mgt_cb(struct ev_loop* loop, struct ev_io* watcher, int revents)
...
+					pgmoneta_management_process_result(client_fd, srv, NULL, 1, true);
+					pgmoneta_log_error("Cannot create process");
+				}
+				else if (pid == 0)
+				{
+					pgmoneta_log_error("Verify: Fork sucessfull");
+					pgmoneta_set_proc_title(1, argv_ptr, "verify", config->servers[srv].name);
+					pgmoneta_management_write_int32(client_fd, 0);
+	
+					//shutdown_ports(); //I'm not sure if I have to call this function
```
Jubilee101, line 1392: I'm afraid you should. The child process inherits all of the sockets from its parent process, but it does not need to listen on any metric/management socket, so it should not keep them open. All it needs is just the client_fd to send info back to the cli process.

src/libpgmoneta/management.c
```diff
@@ -3409,3 +3445,301 @@ print_info_json(struct json* json)
...
+		pgmoneta_log_debug("Verify: inside pgmoneta_management_write_verify_data loop");
+		pgmoneta_log_debug("Output[0] is %s", output[0]);
+		pgmoneta_log_debug("Output[1] is %s", output[1]);
+		pgmoneta_log_debug("Output[2] is %s", output[2]);
+		pgmoneta_log_debug("Output[3] is %s", output[3]);
+		pgmoneta_management_write_int32(socket, 1);
+		write_string("pgmoneta_management_write_verify_data", socket,  output[0]);
+		write_string("pgmoneta_management_write_verify_data", socket,  output[1]);
+		write_string("pgmoneta_management_write_verify_data", socket,  output[2]);
+		write_string("pgmoneta_management_write_verify_data", socket,  output[3]);
```
Jubilee101, line 3728: I don't see you sending should_read at the end of each loop.

Jubilee101: Yet you read should_read at the end of each loop on the other side.

Jubilee101:
> Implemented pgmoneta-cli verify <server> <id> [FAIL|all].

Is `all` a really wise option? We already have 8k entries in the manifest now.. In larger db it could reach a million entry, printing them all doesn't sound very ideal.

jesperpedersen:
I'm ok with `... [fail|all]` since it is going to be tooling that pickup the JSON and process it

src/libpgmoneta/management.c
iuryr: I wondered how I should deal with encrypted and compressed manifests and files, but assumed, for simplicity sake, that users would have to run `pgmoneta-cli decrypt` and decompress a backup before running `pgmoneta-cli`.

After this comment I will look on the `backup.manifest`. Also, I will need to refactor to account for its format not being json.

src/libpgmoneta/management.c
iuryr: You're right about this, Haoran. At line 3722 I send an int to the cli-side, to allow it to read the "first batch" of strings. However, when debugging, I don't see it reading even at this first iteration of the loop. This is what I'm stuck on now.

But you're right, I will need to fix the end of the loop in order to prepare the cli-side to read the strings generated in the following iterations of the loop.

iuryr:
I will be going over Haoran's comments and continue to work on the issue.

> > Please, rebase your code, and run `uncrustify` such that it is easier to review
> 
> I will try to do this still today. uncrustify was failing to format the code. I could not solve it before leaving for school

In regards to uncrustify, it actually worked but I thought it didn't since it issued a lot of warnings regarding `.cfg` file. I will not rebase and force-push today so that I do not loose Haoran's comments for now.

The rest of this week I'm swamped with schools' demands, so, in order to manage expectations I will only manage to work on this issue again on saturday. I think a I'll manage to rebase and force-push a (properly formatted) a commit by the end of sunday.

src/libpgmoneta/management.c
iuryr: @jesperpedersen Can you clarify if I should work with the json `backup_manifest` or the pgmoneta native `backup.manifest` csv file? And should the verify feature also decrypt and decompress manifests when needed - or I should assume the user will do these operations before running `pgmoneta-cli verify <server> <id>`?

Jubilee101: > I wondered how I should deal with encrypted and compressed manifests and files,

Yeah, that's a good point. I wonder if we should decompress and decrypt for user, or at least make it an option for user when verify. 

> After this comment I will look on the backup.manifest. Also, I will need to refactor to account for its format not being json.

No worries. This won't be too much changes, csv is much easier format than json.

Jubilee101:
> I'm ok with ... [fail|all] since it is going to be tooling that pickup the JSON and process it

Our json is not designed for holding this many entries. We probably need to switch ART to hold them internally, later perhaps.

src/libpgmoneta/management.c
Jubilee101: OK, try making the message sequence align on both side. And see if at least the first int32 reaches the other side.

jesperpedersen:
@iuryr You need to run through the "restore" pipeline - so, decrypt and uncompress the files. Then verify the manifest, and then delete the temporary directory.

So, if you are unsure then look at the `restore` command first

jesperpedersen:
@iuryr Doing ok ?

iuryr:
Hi @jesperpedersen , sorry for the lack of updates. I did not manage to work on pgmoneta this past week nor will I manage to work on it this week due to some school deadlines. Whenever I have the time I'm working on a document, some sort of spec, that I intend to run through you before resuming working on the code. The aim is to organize myself better and clarify what are the desired behaviors of the feature.

jesperpedersen:
@iuryr That is ok. However, we need this for 0.13 so I'll take over, and we can find another task for you if you are interested

iuryr:
No problem. I'll reach out soon to see if I can contribute on another task. Thank you for the opportunity

Issue #316, opened 2024-06-20
New link function does not deal with compression suffix


Jubilee101:
When compression is configured, even if a file is in the changed file list, searching that file entry when traversing directory in the changed list won't give positive result. This is becaues we do compression before link, while manifest was generated before compression.  All files now with compression suffix are not in the manifest file. 

We cannot solve this by linking before compression, since after compression the file name changes and symlink is broken. So the only way is to trim the compression suffix from file name when searching it in the changed file list.

Issue #313, opened 2024-06-15
Backup manifest


jesperpedersen:
Create a backup manifest that includes all files that are part of the backup.

The manifest should support the hash algorithms similar to PostgreSQL itself

PR #317, opened 2024-06-20
[#316][#313] Link tablespaces with csv format manifest

Jubilee101:
1. Convert backup_manifest from json format to simplified csv format named as backup.manifest. This is originally for extending the manifest from recording data directory only to also include tablespaces. Upon deeper investigation we found that the manifest includes tablespaces in the first place. But having a csv format manifest is still important for us, since in the future extension side will most likely build manifest as well, having a csv infrastructure is much easier and memory efficient than dealing with json directly.
2. Use the new manifest to link data and tablespaces directory.
3. Fix the bug that files are not linked correctly because of lack of compression suffix handling

New backup.manifest sits in the parent directory of data and tablespaces, the same as backup.info. This brings some extra complexity -- all files in the data directory are prefixed with `data/` for instance, and only data directory should be traversed when linking, tablespaces are traversed through `data/pg_tblspc`. But I still don't think it's a good idea for us to add files directly into user's data directory. 

Note that we could replace `data/pg_tblspc/tblspc_oid` with `tablespace_name/`, but we won't guarantee our manifest works on the server side this way, since tablespaces could be anywhere on the server side. Thus we still use symlinks. Though this only works when pg_tblspc is linked to each table spaces correctly. So some extra work must be done on the hotstandby side, which this patch will not touch.

### DIFF
--- a/src/include/csv.h
+++ b/src/include/csv.h
@@ -26,48 +26,89 @@
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-#ifndef PGMONETA_LINK_H
-#define PGMONETA_LINK_H
+#ifndef PGMONETA_CSV_H
+#define PGMONETA_CSV_H
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
-#include <art.h>
-#include <workers.h>
-
+#include <stdbool.h>
+#include <stdio.h>
 #include <stdlib.h>
 
-/**
- * Create link between two directories with processed manifest info
- * @param base_from The base from directory (newer)
- * @param base_from_data The base from data directory
- * @param base_to The base to directory
- * @param from The current from directory
- * @param changed The changed files
- * @param added The added files
- * @param workers The optional workers
- */
-void
-pgmoneta_link_manifest(char* base_from, char* base_from_data, char* base_to, char* from, struct art* changed, struct art* added, struct workers* workers);
+struct csv_reader
+{
+   FILE* file;
+   char line[512];
+};
+
+struct csv_writer
+{
+   FILE* file;
+};
 
 /**
- * Relink link two directories
- * @param from The from directory
- * @param to The to directory
- * @param workers The optional workers
+ * Initialize a csv reader
+ * @param path The path to the csv file
+ * @param reader The reader
+ * @return 0 on success, 1 if otherwise
  */
-void
-pgmoneta_relink(char* from, char* to, struct workers* workers);
+int
+pgmoneta_csv_reader_init(char* path, struct csv_reader** reader);
 
 /**
- * Create link between two equal files
- * @param from The from directory
- * @param to The to directory
- * @param workers The optional workers
+ * Reset the reader pointer to the head of the file
+ * @param reader The reader
+ * @return 0 on success, 1 if otherwise
  */
-void
-pgmoneta_link_comparefiles(char* from, char* to, struct workers* workers);
+int
+pgmoneta_csv_reader_reset(struct csv_reader* reader);
+
+/**
+ * Initialize a csv writer
+ * @param path The path to the csv file
+ * @param writer The writer
+ * @return 0 on success, 1 if otherwise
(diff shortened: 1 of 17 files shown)


src/include/csv.h
```diff
@@ -0,0 +1,117 @@
...
+};
+
+/**
+ * Initialize a csv reader
+ * @param reader The reader
+ * @param path The path to the csv file
+ * @return 0 on success, 1 if otherwise
+ */
+int
+pgmoneta_csv_reader_init(struct csv_reader** reader, char* path);
```
jesperpedersen, line 58: Reverse parameters

src/include/csv.h
```diff
@@ -0,0 +1,117 @@
...
+pgmoneta_csv_reader_reset(struct csv_reader* reader);
+
+/**
+ * Initialize a csv writer
+ * @param writer The writer
+ * @param path The path to the csv file
+ * @return 0 on success, 1 if otherwise
+ */
+int
+pgmoneta_csv_writer_init(struct csv_writer** writer, char* path);
```
jesperpedersen, line 75: Reverse parameters

src/include/csv.h
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * Write a row to csv file
+ * @param writer The csv writer
+ * @param cols The columns of the row
+ * @param num_col The number of columns
+ * @return 0 on success, 1 if otherwise
+ */
+int
+pgmoneta_csv_write(struct csv_writer* writer, char** cols, int num_col);
```
jesperpedersen, line 85: num_col, cols, writer

src/include/csv.h
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * Get the next row in csv file
+ * @param reader The reader
+ * @param cols [out] The columns in the row
+ * @param num_col [out] The number of columns in the row
+ * @return true if has next row, false if otherwise
+ */
+bool
+pgmoneta_csv_next_row(struct csv_reader* reader, char*** cols, int* num_col);
```
jesperpedersen, line 95: num_col, cols, reader

src/include/pgmoneta.h
```diff
@@ -122,6 +122,11 @@ extern "C" {
 #define SLOT_NOT_FOUND        1
 #define INCORRECT_SLOT_TYPE       2
 
+// simple manifest csv structure definition in case we want to change later
+#define MANIFEST_COLUMN_COUNT 2
+#define MANIFEST_PATH_INDEX 0
+#define MANIFEST_CHECKSUM_INDEX 1
+
```
jesperpedersen, line 129: Should we have these in manifest.h ?

src/libpgmoneta/link.c
```diff
@@ -60,10 +63,9 @@ pgmoneta_link_manifest(char* base_from, char* base_to, char* from, struct art* c
    {
       goto done;
    }
-
    while ((entry = readdir(from_dir)))
    {
-      if (!strcmp(entry->d_name, ".") || !strcmp(entry->d_name, "..") || !strcmp(entry->d_name, "pg_tblspc"))
+      if (!strcmp(entry->d_name, ".") || !strcmp(entry->d_name, ".."))
```
jesperpedersen, line 68: Is this ok ?

src/libpgmoneta/wf_manifest.c
```diff
@@ -0,0 +1,129 @@
+#include <csv.h>
```
jesperpedersen, line 1: License header

src/libpgmoneta/wf_manifest.c
```diff
@@ -0,0 +1,129 @@
...
+#include <json.h>
```
jesperpedersen, line 2: include sorting

jesperpedersen (changes requested):

src/libpgmoneta/link.c
Jubilee101: Not sure what you mean, but it's tested, link function is able to penetrate through symlinks.

Issue #319, opened 2024-06-24
Integrate extension


GuChad369:
- Add the test suite from [pgmoneta_ext](https://github.com/pgmoneta/pgmoneta_ext/pull/11) to install the extension.
- Create `extension.h|.c` that will call the extension functions.

GuChad369:
To integrate the test suite, we can simply copy it from `pgmoneta_ext`.

For `extension.h|.c`, how can we make `pgmoneta` call the extension functions? 

Do we need to use a shell script command?

jesperpedersen:
No, you need to use the wire protocol to query the extension - see https://www.postgresql.org/docs/16/protocol-message-formats.html and `message.h|.c`

jesperpedersen:
You can send multiple pull requests for this in order to get started

GuChad369:
For the test suite, do I need to modify the tests to include the installation of `pgmoneta`, or just copy the test suite from `pgmoneta_ext`?

I would like to first send a pull request to handle the test suite. 

After that, I will send another pull request to handle the modifications to `extension.h|.c`.

GuChad369:
When using `pgmoneta-admin` to set the master key, if a master key already exists, it will cause a segmentation fault. 

I am not sure what the problem might be; could it be an issue with the source code? I tried to solve it, but I have no clue about this.

```sh
sudo -u pgmoneta /tmp/pgmoneta/build/src/pgmoneta-admin master-key -P pgmonetamasterkey
```

<img width="1189" alt="Screenshot 2024-06-25 at 12 00 24 PM" src="https://github.com/pgmoneta/pgmoneta/assets/113565433/a8c8f339-ca83-4b06-ad07-f4979ef07832">


With this problem, we can't set up the script to auto-configure `pgmoneta`.

jesperpedersen:
Does it happen locally for you as well ? In case with `-P` or `-g`

GuChad369:
> Does it happen locally for you as well ? In case with `-P` or `-g`

Yeah, you are right. Without the `-P` option, there is no segmentation fault. It seems to be a problem with the source code.

<img width="1194" alt="Screenshot 2024-06-25 at 12 58 50 PM" src="https://github.com/pgmoneta/pgmoneta/assets/113565433/87d9750e-2c60-4642-a889-a2f50500dc0c">

jesperpedersen:
Why is there a `master.key` - it should only be run once

jesperpedersen:
#320

[...]

PR #321, opened 2024-06-26
[#319] Test suite for pgmoneta

GuChad369:
This is the initial version of the test suite for `pgmoneta`. It is not yet complete and still has a long way to go.

Currently, I am facing a decision: should we separate the tests into several scripts, or use just one script for everything?

If we separate them, I envision the structure as follows:

- test
  - script
    - `installation.sh`: Installs all dependencies, `PostgreSQL`, `pgmoneta`, and `pgmoneta_ext`.
    - `test_pgmoneta.sh`: Tests `pgmoneta` backup and restore, and will include more tests once `extension.h|.c` is complete.
    - `test_pgmoneta_ext.sh`: Tests extension functions independently, such as those in our `pgmoneta_ext` project.
  - c code
    - `pgmoneta.c`: Tests `pgmoneta` by calling the extension.
    - `pgmoneta_ext.c`: Tests extension functions.

Is this a better approach, or should we combine everything into one script that automates the entire process from start to finish?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,9 @@ endif(CMAKE_SOURCE_DIR STREQUAL CMAKE_BINARY_DIR)
 
 project(pgmoneta VERSION ${VERSION_STRING} LANGUAGES  C)
 
+include(CTest)
+enable_testing()
+
 set(CPACK_PACKAGE_VERSION_MAJOR ${VERSION_MAJOR})
 set(CPACK_PACKAGE_VERSION_MINOR ${VERSION_MINOR})
 set(CPACK_PACKAGE_VERSION_PATCH ${VERSION_PATCH})
@@ -33,6 +36,8 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 message(STATUS "pgmoneta ${VERSION_STRING}")
 
 set(generation TRUE)
+set(check TRUE)
+set(container FALSE)
 
 include(CheckCCompilerFlag)
 include(CheckCSourceCompiles)
@@ -104,6 +109,27 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
         OUTPUT_STRIP_TRAILING_WHITESPACE
         COMMAND_ERROR_IS_FATAL ANY
     )
+endif()
+
+find_package(Check)
+if (CHECK_FOUND)
+  message(STATUS "check found")
+  add_library(Check::check SHARED IMPORTED)
+  set_target_properties(Check::check PROPERTIES
+    IMPORTED_LOCATION ${CHECK_LIBRARY}
+    INTERFACE_INCLUDE_DIRECTORIES ${CHECK_INCLUDE_DIR})
+else ()
+  set(check FALSE)
+  message(STATUS "check needed. The test suite process will be skipped.")
+endif()
+
+find_package(Docker)
+find_package(Podman)
+if (DOCKER_FOUND OR PODMAN_FOUND)
+  set(container TRUE)
+  message(STATUS "Docker or podman found")
+else ()
+  message(STATUS "Docker or podman needed. The test suite will be skipped.")
 endif()
 
 find_package(Pandoc)
@@ -225,3 +251,4 @@ file(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/src/")
 
 add_subdirectory(doc)
 add_subdirectory(src)
+add_subdirectory(test)
--- a/cmake/FindCheck.cmake
+++ b/cmake/FindCheck.cmake
@@ -1,0 +1,29 @@
+#
+# check support
+#
+
+find_package(PkgConfig)
+pkg_check_modules(PC_CHECK check)
+
+find_path(CHECK_INCLUDE_DIR
+  NAMES check.h
+  PATHS ${PC_CHECK_INCLUDE_DIRS}
+  PATH_SUFFIXES check
+)
+
+find_library(CHECK_LIBRARY
+  NAMES check
+  PATHS ${PC_CHECK_LIBRARY_DIRS}
+)
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(Check DEFAULT_MSG CHECK_INCLUDE_DIR CHECK_LIBRARY)
+
+if (CHECK_FOUND)
+  set(CHECK_LIBRARIES ${CHECK_LIBRARY})
+  set(CHECK_INCLUDE_DIRS ${CHECK_INCLUDE_DIR})
+else ()
+  set(check FALSE)
+endif ()
+
+mark_as_advanced(CHECK_INCLUDE_DIR CHECK_LIBRARY)
--- a/cmake/FindDocker.cmake
+++ b/cmake/FindDocker.cmake
@@ -1,0 +1,18 @@
+#
+# Check for Docker
+#
+
+find_program(DOCKER_EXECUTABLE
+  NAMES docker
+  PATHS /usr/local/bin /usr/bin /bin
+)
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(Docker DEFAULT_MSG DOCKER_EXECUTABLE)
+
+if (DOCKER_FOUND)
+  set(container TRUE)
+  set(DOCKER_EXECUTABLE ${DOCKER_EXECUTABLE})
+endif ()
+
+mark_as_advanced(DOCKER_EXECUTABLE)
--- a/cmake/FindPodman.cmake
+++ b/cmake/FindPodman.cmake
@@ -1,0 +1,18 @@
+#
+# Check for Podman
+#
+
+find_program(PODMAN_EXECUTABLE
+  NAMES podman
+  PATHS /usr/local/bin /usr/bin /bin
+)
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(Podman DEFAULT_MSG PODMAN_EXECUTABLE)
+
+if (PODMAN_FOUND)
+  set(container TRUE)
+  set(PODMAN_EXECUTABLE ${PODMAN_EXECUTABLE})
+endif ()
+
+mark_as_advanced(PODMAN_EXECUTABLE)
--- a/doc/TEST.md
+++ b/doc/TEST.md
@@ -1,0 +1,84 @@
+# Test
+
+## Container Environment
+
+To ensure the test suite works well, please make sure you have installed `Docker` or `podman` on your OS. If neither `Docker` nor `podman` is installed, the test suite compilation will be skipped.
+
+### Docker
+
+First, ensure your system is up to date.
+
+```sh
+dnf update
+```
+
+Install the necessary packages for Docker.
+
+```sh
+dnf -y install dnf-plugins-core
+```
+
+Add the Docker repository to your system.
+
+``` sh
+sudo dnf config-manager --add-repo https://download.docker.com/linux/fedora/docker-ce.repo
+```
+
+Install Docker Engine, Docker CLI, and Containerd.
+
+```sh
+sudo dnf install docker-ce docker-ce-cli containerd.io
+```
+
+Start the Docker service and enable it to start on boot.
+
+```sh
+sudo systemctl start docker
+sudo systemctl enable docker
+```
+
+Verify that Docker is installed correctly.
+
+```sh
+docker --version
+```
+
+If you see the Docker version, then you have successfully installed Docker on Fedora.
+
+### Podman
+
+Install Podman and the Docker alias package.
+
+```sh
+dnf install podman podman-docker.noarch
+```
+
+Verify that Podman is installed correctly.
+
+```sh
+podman --version
+```
+
+If you see the Podman version, then you have successfully installed Podman on Fedora.
+
+The `podman-docker.noarch` package simplifies the use of `Podman` for users accustomed to Docker.
+
+## Test suite
+
+Before you test, you need to install the `check` library. If there is no package for `check`, the `CMakeLists.txt` will not compile the test suite. Only after you have installed `check` will it compile the test suite.
+
+``` sh
+dnf install -y check check-devel check-static
+```
+
+You can simply use `CTest` to test all PostgreSQL versions from 13 to 16. It will automatically run `testsuite.sh` to test `pgmoneta` and `pgmoneta_ext` for each version. The script will automatically create the Docker container, run it, and then use the `check` framework to test their functions inside it. After that, it will automatically clean up everything for you.
+
+After you follow the [DEVELOPERS.md](https://github.com/pgmoneta/pgmoneta_ext/blob/main/doc/DEVELOPERS.md#developer-guide) to install `pgmoneta`, go to the directory `/pgmoneta/build` and run the test.
+
+``` sh
+make test
+```
+
+`CTest` will output logs into `/pgmoneta/build/Testing/Temporary/LastTest.log`. If you want to check the specific process, you can review that log file.
+
+`testsuite.sh` accepts three variables. The first one is `dir`, which specifies the `/test` directory location, with a default value of `./`. The second one is `dockerfile`, with a default value of `Dockerfile.rocky8`. The third one is the PostgreSQL `version`, with a default value of `13`.
--- a/test/lib/pgmoneta_ext_test.h
+++ b/test/lib/pgmoneta_ext_test.h
@@ -1,0 +1,55 @@
+/*
+ * Copyright (C) 2024 The pgmoneta community
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
+ * are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this list
+ * of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice, this
+ * list of conditions and the following disclaimer in the documentation and/or other
+ * materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its contributors may
+ * be used to endorse or promote products derived from this software without specific
+ * prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+ * OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+ * THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT
+ * OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR
+ * TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ */
+
+#ifndef PGMONETA_EXT_TEST_H
+#define PGMONETA_EXT_TEST_H
+
+#include <check.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+/**
+ * Execute a shell command and capture its output
+ * @param command The shell command to execute
+ * @param output A buffer to store the command's output
+ * @param output_size The size of the output buffer
+ * @return 0 on success, -1 on error
+ */
+static int
+execute_command(const char* command, char* output, size_t output_size);
+
+/**
+ * Set up a suite of test cases for pgmoneta_ext
+ * @return The result
+ */
+Suite*
+pgmoneta_ext_suite(void);
+
+#endif // PGMONETA_EXT_TEST_H
--- a/test/testsuite.sh
+++ b/test/testsuite.sh
@@ -1,0 +1,136 @@
+#!/bin/bash
+#
+# Copyright (C) 2024 The pgmoneta community
+#
+# Redistribution and use in source and binary forms, with or without modification,
+# are permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this list
+# of conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice, this
+# list of conditions and the following disclaimer in the documentation and/or other
+# materials provided with the distribution.
+#
+# 3. Neither the name of the copyright holder nor the names of its contributors may
+# be used to endorse or promote products derived from this software without specific
+# prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+# OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+# THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT
+# OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+# HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR
+# TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+set -e
+
+check_podman_installed() {
+  if command -v podman &> /dev/null; then
+    return 0
+  else
+    return 1
+  fi
+}
+
+run_podman(){
+  local dockerfile=${1:-'Dockerfile.rocky8'}
+  local version=${2:-13}
+
+  if ! podman build -f "$dockerfile" --build-arg PGVERSION="$version" -t postgres_pgmoneta_image .; then
+    echo "Podman build failed."
+    clean_podman
+    exit 1
+  fi
+
+  podman run -d --name postgres_pgmoneta_container postgres_pgmoneta_image
+
+  sleep 5
+
+  podman logs -f postgres_pgmoneta_container
+
+  clean_podman
+  
+}
+
+clean_podman(){
+  podman stop postgres_pgmoneta_container || true
+  podman rm postgres_pgmoneta_container || true
+  podman rmi postgres_pgmoneta_image || true
+  podman builder prune -f || true
+}
+
+check_docker_installed() {
+  if command -v docker &> /dev/null; then
+    return 0
+  else
+    return 1
+  fi
+}
+
+run_docker(){
+  local dockerfile=${1:-'Dockerfile.rocky8'}
+  local version=${2:-13}
+
+  if ! systemctl restart docker; then
+    echo "Failed to restart Docker. Please ensure Docker is installed and running."
+    clean_docker
+    exit 1
+  fi
+
+  if ! docker build -f "$dockerfile" --build-arg PGVERSION="$version" -t postgres_pgmoneta_image .; then
+    echo "Docker build failed."
+    clean_docker
+    exit 1
+  fi
+
+  docker run --name postgres_pgmoneta_container postgres_pgmoneta_image & container_id=$!
+
+  wait $container_id
+
+  clean_docker
+}
+
+clean_docker(){
+  docker stop postgres_pgmoneta_container || true
+  docker rm postgres_pgmoneta_container || true
+  docker rmi postgres_pgmoneta_image || true
+  docker builder prune -f || true
+}
+
+run_tests() {
+  local dir=$1
+  local dockerfile=$2
+  local version=$3
+
+  valid_versions=("13" "14" "15" "16")
+  if [[ ! " ${valid_versions[@]} " =~ " ${version} " ]]; then
+    echo "Invalid version. Please provide a version of 13, 14, 15, or 16."
+    exit 1
+  fi
+
+  cd "$dir"
+
+  if check_podman_installed; then
+    echo "Podman is installed."
+    run_podman "$dockerfile" "$version"
+  else
+    echo "Podman is not installed."
+    if check_docker_installed; then
+      echo "Docker is installed."
+      run_docker "$dockerfile" "$version"
+    else
+      echo "Please install Podman or Docker to proceed."
+      exit 1
+    fi
+  fi
+}
+
+dir=${1:-'./'}
+dockerfile=${2:-'Dockerfile.rocky8'}
+version=${3:-13}
+
+run_tests "$dir" "$dockerfile" "$version"
--- a/doc/manual/99-references.md
+++ b/doc/manual/99-references.md
@@ -37,10 +37,12 @@
   [aes_ni]: https://en.wikipedia.org/wiki/AES_instruction_set
   [ART_paper]: http://www-db.in.tum.de/~leis/papers/ART.pdf
   [libart]: https://github.com/armon/libart
+  [pgsqlrepo]: https://github.com/postgres/postgres
 
 <!-- doc/ -->
   [rpm]: https://github.com/pgmoneta/pgmoneta/blob/main/doc/RPM.md
   [configuration]: https://github.com/pgmoneta/pgmoneta/blob/main/doc/CONFIGURATION.md
+  [developers]: https://github.com/pgmoneta/pgmoneta_ext/blob/main/doc/DEVELOPERS.md#developer-guide
 <!-- doc/tutorial -->
   [t_install]: https://github.com/pgmoneta/pgmoneta/blob/main/doc/tutorial/01_install.md
   [t_remote_management]: https://github.com/pgmoneta/pgmoneta/blob/main/doc/tutorial/02_remote_management.md
--- a/doc/manual/dev-05-test.md
+++ b/doc/manual/dev-05-test.md
@@ -1,0 +1,84 @@
+\newpage
+
+# Test
+
+## Container Environment
+
+### Docker
+
+First, ensure your system is up to date.
+
+```sh
+dnf update
+```
+
+Install the necessary packages for Docker.
+
+```sh
+dnf -y install dnf-plugins-core
+```
+
+Add the Docker repository to your system.
+
+``` sh
+sudo dnf config-manager --add-repo https://download.docker.com/linux/fedora/docker-ce.repo
+```
+
+Install Docker Engine, Docker CLI, and Containerd.
+
+```sh
+sudo dnf install docker-ce docker-ce-cli containerd.io
+```
+
+Start the Docker service and enable it to start on boot.
+
+```sh
+sudo systemctl start docker
+sudo systemctl enable docker
+```
+
+Verify that Docker is installed correctly.
+
+```sh
+docker --version
+```
+
+If you see the Docker version, then you have successfully installed Docker on Fedora.
+
+### Podman
+
+Install Podman and the Docker alias package.
+
+```sh
+dnf install podman podman-docker.noarch
+```
+
+Verify that Podman is installed correctly.
+
+```sh
+podman --version
+```
+
+If you see the Podman version, then you have successfully installed Podman on Fedora.
+
+The `podman-docker.noarch` package simplifies the use of `Podman` for users accustomed to Docker.
+
+## Test suite
+
+You can simply use `CTest` to test all PostgreSQL versions from 13 to 16. It will automatically run `testsuite.sh` to test `pgmoneta` and `pgmoneta_ext` for each version. The script will automatically create the Docker container, run it, and then use the `check` framework to test their functions inside it. After that, it will automatically clean up everything for you.
+
+Go to the directory `/pgmoneta/test`, and give permission to `testsuite.sh` using:
+
+``` sh
+chmod +x testsuite.sh
+```
+
+After you follow the [DEVELOPERS.md][developers] to install `pgmoneta`, go to the directory `/pgmoneta/build` and run the test.
+
+``` sh
+make test
+```
+
+`CTest` will output logs into `/pgmoneta/build/Testing/Temporary/LastTest.log`. If you want to check the specific process, you can review that log file.
+
+`testsuite.sh` accepts three variables. The first one is `dir`, which specifies the `/test` directory location, with a default value of `./`. The second one is `dockerfile`, with a default value of `Dockerfile.rocky8`. The third one is the PostgreSQL `version`, with a default value of `13`.
--- a/src/libpgmoneta/security.c
+++ b/src/libpgmoneta/security.c
@@ -2890,7 +2890,11 @@ create_hash_file(char* filename, const char* algorithm, char** hash)
 
    md_ctx = EVP_MD_CTX_new();
 
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
    if (!EVP_DigestInit_ex2(md_ctx, md, NULL))
+#else
+   if (!EVP_DigestInit_ex(md_ctx, md, NULL))
+#endif
    {
       pgmoneta_log_error("Message digest initialization failed");
       EVP_MD_CTX_free(md_ctx);
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,0 +1,68 @@
+# 
+#  Copyright (C) 2024 The pgmoneta community
+# 
+#  Redistribution and use in source and binary forms, with or without modification,
+#  are permitted provided that the following conditions are met:
+# 
+#  1. Redistributions of source code must retain the above copyright notice, this list
+#  of conditions and the following disclaimer.
+# 
+#  2. Redistributions in binary form must reproduce the above copyright notice, this
+#  list of conditions and the following disclaimer in the documentation and/or other
+#  materials provided with the distribution.
+# 
+#  3. Neither the name of the copyright holder nor the names of its contributors may
+#  be used to endorse or promote products derived from this software without specific
+#  prior written permission.
+# 
+#  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+#  EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+#  OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+#  THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+#  SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT
+#  OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+#  HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR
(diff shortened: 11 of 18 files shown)


jesperpedersen:
For now, go in that direction !

We can have a `testsuite.sh` script that call all the other scripts - we can see if there is a better way when using the `check` project... We can always change it later once we have a feeling of what is best

GuChad369:
Use the `testsuite.sh` script to perform the tests. This script will install all necessary dependencies, `PostgreSQL`, `pgmoneta`, and `pgmoneta_ext`. It will then call the `check` framework to test the functions and clean up everything afterward.

jesperpedersen:
Please, rebase.

Take a look at https://cmake.org/cmake/help/book/mastering-cmake/chapter/Testing%20With%20CMake%20and%20CTest.html and see what we can do in order to support it

GuChad369:
> Take a look at https://cmake.org/cmake/help/book/mastering-cmake/chapter/Testing%20With%20CMake%20and%20CTest.html and see what we can do in order to support it

So, we will not use the check framework anymore and will instead figure out how to use CTest.

jesperpedersen:
Ok, take a quick look and see if it is worth it... we can always migrate later too

GuChad369:
I have read about it and done some research. The CMake testing is mainly done using the function:

``` CMake
add_test(NAME TestName COMMAND ExecutableToRun arg1 arg2 ...)
```

`ExecutableToRun` is any executable file. Each function runs one test, and if we add multiple functions, it will run multiple tests for us. It acts as a control system to manage the executable files and help us run them.

For our project, it may look something like this:

``` CMake
add_test(Test_version_13 "${CMAKE_CURRENT_SOURCE_DIR}/testsuite.sh" "n" 13)
add_test(Test_version_14 "${CMAKE_CURRENT_SOURCE_DIR}/testsuite.sh" "n" 14)
add_test(Test_version_15 "${CMAKE_CURRENT_SOURCE_DIR}/testsuite.sh" "n" 15)
add_test(Test_version_16 "${CMAKE_CURRENT_SOURCE_DIR}/testsuite.sh" "n" 16)
```

Or we can completely rewrite the tests, create a separate "script file" (or other executable file) for each function, and use `CMake` to control them instead of using the` check` framework.

``` CMake
add_test(Test_installation "${CMAKE_CURRENT_SOURCE_DIR}/installation.sh")
add_test(Test_pgmoneta_backup "${CMAKE_CURRENT_SOURCE_DIR}/pgmoneta_backup.sh")
add_test(Test_pgmoneta_restore "${CMAKE_CURRENT_SOURCE_DIR}/pgmoneta_retore.sh")
add_test(Test_pgmoneta_ext_function "${CMAKE_CURRENT_SOURCE_DIR}/pgmoneta_ext_version.sh")
add_test(Test_pgmoneta_ext_function "${CMAKE_CURRENT_SOURCE_DIR}/pgmoneta_ext_switch_wal.sh")
add_test(Test_pgmoneta_ext_function "${CMAKE_CURRENT_SOURCE_DIR}/pgmoneta_ext_checkpoint.sh")
```

Jubilee101:
> ExecutableToRun is any executable file. Each function runs one test, and if we add multiple functions, it will run multiple tests for us. It acts as a control system to manage the executable files and help us run them.

So we can just keep using check, right? 

> For our project, it may look something like this:
> add_test(Test_version_13 "${CMAKE_CURRENT_SOURCE_DIR}/testsuite.sh" "n" 13)
> add_test(Test_version_14 "${CMAKE_CURRENT_SOURCE_DIR}/testsuite.sh" "n" 14)
> add_test(Test_version_15 "${CMAKE_CURRENT_SOURCE_DIR}/testsuite.sh" "n" 15)
> add_test(Test_version_16 "${CMAKE_CURRENT_SOURCE_DIR}/testsuite.sh" "n" 16)

Yeah this looks good if it's no trouble for you. Make sure you add_test in test/CMakeLists.txt rather than the top dir cmake file. Also we should disable the test suite if check is not found.

> Or we can completely rewrite the tests, create a separate "script file" (or other executable file) for each function, and use CMake to control them instead of using the check framework.

Let's not go down this path for now to keep our life simpler.

GuChad369:
I used `CTest` to control our tests, and the result is something like this:

Screenshot 2024-06-28 at 10 54 36 AM


`CTest` will test each function one by one, unlike the `check` framework, which runs them concurrently.

There is a problem, however. 

`CTest` only verifies if the test runs without error. Since we use the `check` framework to validate functions, even if some functions fail (like if `pgmoneta` backup fails), `CTest` will not detect it as long as the script runs normally. 

The `check` framework, on the other hand, would report the backup failure.

jesperpedersen:
Yeah, we need the failure to be reported, so if the failure can't be extracted from `check` then we can't use the `ctest` framework

GuChad369:
> Yeah, we need the failure to be reported, so if the failure can't be extracted from `check` then we can't use the `ctest` framework

There is a `set_property()` function to set our failure condition. It will search for the provided regular expression, and if found, it will indicate a failure. 

So, we can add the check failure expression like

``` CMake
set (failRegex 
    "Failures: 1"
    "Failures: 2"
    "Failures: 3"
)
set_property (TEST test_version_13
              PROPERTY FAIL_REGULAR_EXPRESSION "${failRegex}")
```

 It will report a failure but won't show the `check` report.

Screenshot 2024-06-28 at 12 34 14 PM

jesperpedersen:
But, we still have the `check` failure log, right ?

GuChad369:
> But, we still have the `check` failure log, right ?

Yes, it will output a single log of the entire process. I have mentioned this in **TEST.md**.

jesperpedersen:
Ok, see if you can get the `test/` hierarchy called from the top-level `CMakeLists.txt` - it would be similar to `doc` and `src`

GuChad369:
> Ok, see if you can get the `test/` hierarchy called from the top-level `CMakeLists.txt` - it would be similar to `doc` and `src`

Currently, we use different `CMakeLists.txt` to compile the tests: one for `pgmoneta_test` in `/test/pgmoneta` and one for `pgmoneta_ext_test` in `/test/pgmoneta_ext`. These two CMake files are not subdirectories of the `/test/CMakeLists.txt`. 

Since we want to allow users to test only `pgmoneta` or `pgmoneta_ext` independently, is it a good choice to have the top-level `CMakeLists.txt` call the tests?

GuChad369:
If we call tests from the top-level `CMakeLists.txt`, the user would have to install the `check` library, which is not directly related to our `pgmoneta` project. The `check` library is only used for unit testing.

jesperpedersen:
Yeah, so it should be optional

jesperpedersen:
The `CMakeLists.txt` files should be as simple as possible, and it should be easy to run the test suite from `build/`

GuChad369:
> The `CMakeLists.txt` files should be as simple as possible, and it should be easy to run the test suite from `build/`

With this requirement, we will not support users testing `pgmoneta` or `pgmoneta_ext` independently. The only way to run the test suite is through the `/build` directory by calling `make test`.

I will restructure the current directory to integrate the tests. Currently, all tests are independent.

jesperpedersen:
We can always make explicit targets.

For now, it is best to make it easy for developers to run the test suite - even with existing installations

GuChad369:
> For now, it is best to make it easy for developers to run the test suite - even with existing installations

Our test suite requires users to install and configure `PostgreSQL` and `pgmoneta` using `installation.sh`. If users manually install `PostgreSQL` and `pgmoneta`, the test suite will not work, as the configuration needs to match what is set up by `installation.sh`.

jesperpedersen:
Precisely, so we need to figure out that part. Maybe not part of this PR...

Developers will compile and install PostgreSQL manually a lot of times, and manually install the extension too - so in certain cases it should just assume that PostgreSQL - in some version - is running on `:5432`

GuChad369:
> Precisely, so we need to figure out that part. Maybe not part of this PR...

I thought about this.

If we want to achieve this, we need to know where `PostgreSQL` is installed and its configuration files, as well as where `pgmoneta` is installed and its configuration files. Additionally, we need to know all the passwords for the role `repl`.

We also have to assume that the user configured `pgmoneta` following our instructions in `DEVELOPER.md`.

Everything not fixed in the configuration should be known. Therefore, I set this aside and focused on ensuring we can run the tests successfully.

Jubilee101:
Perhaps we can do setup and testing in a container, podman or docker, so that we don't contaminate the dev environment, and have a controllable testing environment... Let us know if it's too much workload, we can always do it in another PR later.

GuChad369:
> Perhaps we can do setup and testing in a container, podman or docker, so that we don't contaminate the dev environment, and have a controllable testing environment... Let us know if it's too much workload, we can always do it in another PR later.

Good idea! This is what I am attempting to do. For this PR, we should focus on making it work.

At least make it run first, then try to improve it.

jesperpedersen:
Having a container based environment is a good direction to go in

jesperpedersen:
https://hub.docker.com/_/postgres

jesperpedersen:
Just look into the different options such that we don't have to rely on `installation.sh` in the default case.

Having the tests run in a container is def isolated - we can look at how we transfer any failure to the CI...

Take your time on this, we want a good setup that is easy for all developers to use and doesn't mess with existing setups

GuChad369:
I did some research about Docker.

There are some options for how to test in Docker:

- Option 1: We initialize one basic OS image and then run our current tests on the image.
- Option 2: We initialize the OS and install PostgreSQL using the official image (this way, we may lose control of the installation since it will install the currently available image, which we may not need).
- Option 3: We initialize the OS and install all necessary dependencies, PostgreSQL, `pgmoneta`, and `pgmoneta_ext`, then run the `check` function on that image.

Which option is better? Or is there another choice for us to conduct the tests? I will try to go in that direction.

jesperpedersen:
Well, we want to run the pull request based on the latest main - for both core and ext - for all supported PostgreSQL versions.

Is it easier to begin with a `contrib/images` directory which has a `Dockerfile` for each supported PostgreSQL version ? Something like https://github.com/jesperpedersen/podman-pgsql/blob/master/pgsql14-primary-rocky8/Dockerfile ... Ideally for Rocky 8 / 9, and FreeBSD...

At least to me it sounds like 3) is the best way to go

GuChad369:
> Is it easier to begin with a `contrib/images` directory which has a `Dockerfile` for each supported PostgreSQL version ? 

I suppose it is simple. If we successfully create one image for one PostgreSQL version, we can just copy that image to create other versions. When doing tests, we can run each image separately.

jesperpedersen:
Yeah, it should be possible to have a template and just change the version numbers

GuChad369:
When using Docker to run `pgmoneta`, why does an error occur stating that `pgmoneta` can't start, even though it was installed successfully?

``` console
pgmoneta-cli: pgmoneta-cli: Host and port must be specified
```

Everything is okay. I can use `pgmoneta -?` to get the version, and all configurations are as recommended.

jesperpedersen:
Where is your `pgmoneta.conf` file ?

jesperpedersen:
For the containers I would install all configuration in `/etc/pgmoneta/`

GuChad369:
I run `pgmoneta` specifying the config using:

``` sh 
su - pgmoneta -c '/pgmoneta/build/src/pgmoneta -c /pgmoneta/pgmoneta.conf -u /pgmoneta/pgmoneta_users.conf -d
```

I suppose it will use the specified config file to start `pgmoneta`.

Everything else works well for now.

GuChad369:
The `Dockerfile.rocky` is currently correct. After successfully running it, we can convert it into FreeBSD.

contrib/images/Dockerfile.freebsd
```diff
@@ -0,0 +1,127 @@
...
+#  OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+#  THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+#  SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT
+#  OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+#  HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR
+#  TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+#  SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+# 
+
+FROM freebsd:13.0
```
jesperpedersen, line 29: Is there a `14` image, or better `:latest` ?

contrib/images/Dockerfile.rocky
```diff
@@ -0,0 +1,127 @@
...
+#  OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+#  THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+#  SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT
+#  OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+#  HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR
+#  TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+#  SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+# 
+
+FROM rockylinux:9
```
jesperpedersen, line 29: Can you split into 2 files - one with `8` and one with `9` ?

contrib/images/conf/postgresql.conf
```diff
@@ -0,0 +1,780 @@
+# -----------------------------
```
jesperpedersen, line 1: It is likely better to append to the `postgresql.conf` or do an `include` with just the changes that we need instead of maintaining complete files across all versions

contrib/images/test/CMakeLists.txt
```diff
@@ -0,0 +1,70 @@
+#
```
jesperpedersen, line 1: Why did the directory move ?

If `check` is installed, it should generate all the test images, and then compile and execute the test suite under `/test`

jesperpedersen (changes requested):

contrib/images/test/CMakeLists.txt
GuChad369: This way, it will be easy for the container to copy the test source code.

> If `check` is installed, it should generate all the test images, and then compile and execute the test suite under `/test`

Do you mean we need to integrate this test into the top-level `CMakeLists.txt` using `CTest`? 

If we want to succeed with this function, we need to fix all configurations; otherwise, the check won't work.

The test only works for container configurations. For user-custom configurations, it will not work.

GuChad369:
The current problem is that in the container, `pgmoneta` can't start and shows an error. I tried several ways, but it still doesn't work. I'm not sure why it requires the host and port to be specified.

contrib/images/test/CMakeLists.txt
GuChad369: After we ensure Docker tests run well, I will create a custom command in the top-level `CMakeLists.txt` to allow users to call the Docker test. This command will create the image, run the test, and print the results. This is what I am thinking.

jesperpedersen: I meant to have a `cmake/Findcheck.cmake` file that defines a `HAVE_CHECK` and the top-level `CMakeLists.txt` adds the `test/CMakeLists.txt` in order to compile the test suite if it is.

We can have the `Dockerfile`s in the `test/` directory to start with

jesperpedersen:
I don't know what you mean by that.

Does https://github.com/jesperpedersen/podman-pgsql/tree/master/pgsql14-pgmoneta-rocky8 help ?

jesperpedersen:
Lets start by move the current files to `test/` and go from there - `contrib/` is reserved for "extra" functionality, so it could `Dockerfile`s to generate a container image

contrib/images/test/CMakeLists.txt
GuChad369: OK, we can do this, it’s easy. But I meant that it’s meaningless.

Although users can compile the test suite, they can’t use it.

The test suite only works in the container.

jesperpedersen: Yeah, but we have to start somewhere

GuChad369: Okay

jesperpedersen: For now we have to create an infrastructure such that you can continue with the `extra` configuration property and test it...

GuChad369:
For the OS rockylinux:8, there is an error in the source code.

``` console
5.577 /pgmoneta/src/libpgmoneta/security.c: In function 'create_hash_file':
5.577 /pgmoneta/src/libpgmoneta/security.c:2894:9: error: implicit declaration of function 'EVP_DigestInit_ex2'; did you mean 'EVP_DigestInit_ex'? [-Werror=implicit-function-declaration]
5.577     if (!EVP_DigestInit_ex2(md_ctx, md, NULL))
5.577          ^~~~~~~~~~~~~~~~~~
5.577          EVP_DigestInit_ex
5.780 cc1: all warnings being treated as errors
5.782 make[2]: *** [src/CMakeFiles/pgmoneta.dir/build.make:524: src/CMakeFiles/pgmoneta.dir/libpgmoneta/security.c.o] Error 1
5.782 make[1]: *** [CMakeFiles/Makefile2:167: src/CMakeFiles/pgmoneta.dir/all] Error 2
5.782 make: *** [Makefile:156: all] Error 2
------
```

contrib/images/Dockerfile.freebsd
GuChad369: I tried to use `FreeBSD` but can't download the `FreeBSD` image from Docker. Additionally, for Docker, if you want to use `FreeBSD`, the host OS must be `FreeBSD`. There is no official image in Docker for `FreeBSD` at the moment.

https://stackoverflow.com/questions/33864142/can-freebsd-be-run-inside-docker

jesperpedersen:
Maybe we can use an `#if OPENSSL_VERSION_NUMBER`...

We can remove the FreeBSD part if that is easier for now

GuChad369:
When we use `CTest`, do we need to test different OSs, or just different PostgreSQL versions on one OS? Currently, I have chosen Rocky 8.

```CMake
add_test(test_version_13_rocky8 "${CMAKE_CURRENT_SOURCE_DIR}/../test/testsuite" "${CMAKE_CURRENT_SOURCE_DIR}/../test" "Dockerfile.rocky8" 13)
add_test(test_version_14_rocky8 "${CMAKE_CURRENT_SOURCE_DIR}/../test/testsuite" "${CMAKE_CURRENT_SOURCE_DIR}/../test" "Dockerfile.rocky8" 14)
add_test(test_version_15_rocky8 "${CMAKE_CURRENT_SOURCE_DIR}/../test/testsuite" "${CMAKE_CURRENT_SOURCE_DIR}/../test" "Dockerfile.rocky8" 15)
add_test(test_version_16_rocky8 "${CMAKE_CURRENT_SOURCE_DIR}/../test/testsuite" "${CMAKE_CURRENT_SOURCE_DIR}/../test" "Dockerfile.rocky8" 16)
```

jesperpedersen:
You can do Rocky 8 and Rocky 9 with PostgreSQL 13 to 16

Jubilee101:
There is https://github.com/pgmoneta/pgmoneta/blob/main/doc/DISTRIBUTIONS.md in case you have trouble setting things up @GuChad369

GuChad369:
An updated version of the test suite for `pgmoneta`.

Users can simply type `make test` after installing our project, and it will run `CTest` to test our project.

The process involves **8** tests, covering `Rocky 8` and `Rocky 9` with PostgreSQL versions 13 to 16. Each time, it will create a Docker image, run the container, and execute the check functions inside the container. After testing, it will automatically clean up everything and continue with the remaining tests.

`CTest` will ensure all tests run correctly. If there is a failure in the `check` process, it will show the problem and write the process into a log. Users can check the log to find the specific failure.

The entire test process will take approximately **40** minutes. The result will look something like this:

Screenshot 2024-07-04 at 7 38 54 PM

test/testsuite
```diff
@@ -0,0 +1,69 @@
...
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+# OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+# THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT
+# OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+# HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR
+# TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#!/bin/bash
```
jesperpedersen, line 27: Move this to the first line, and rename the file to `testsuite.sh`

doc/TEST.md
```diff
@@ -0,0 +1,62 @@
+# Test
+
+## Install Docker
```
jesperpedersen, line 3: We should have `## Container environment`, and then a section for Docker and a section for podman.

The package for the podman docker alias is `podman-docker.noarch`

jesperpedersen (changes requested):

jesperpedersen:
2 small changes, otherwise I think it looks good

doc/TEST.md
GuChad369: Do we need to change the content of `testsuite.sh` to use `Podman` to manage the process, or is the current setup using the command `docker ... ` okay?

jesperpedersen: The `podman-docker` package will just install a symlink for `docker -> podman`, and all commands are the same.

Just its part of the standard repositories, and you don't have to `systemctl enable` it, so the setup is a lot simpler

GuChad369:
Is the `Test.md` file correct?

test/lib/pgmoneta_ext_test.h
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Execute a shell command and capture its output
+ * @param command The shell command to execute
+ * @param output A buffer to store the command's output
+ * @param output_size The size of the output buffer
+ * @return 0 on success, -1 on error
+ */
+int
+execute_command(const char* command, char* output, size_t output_size);
```
Jubilee101, line 46: Perhaps this should be in utils.h, or just be a static function withing pgmoneta_ext_test.c

jesperpedersen: static function

jesperpedersen:
You have to document the `check` install requirement -- its `dnf install -y check check-devel check-static` on Fedora

Also, using `podman` doesn't work... uninstall docker, and install podman and work through the issues

GuChad369:
> You have to document the `check` install requirement -- its `dnf install -y check check-devel check-static` on Fedora

You only need to install `check` using:
```sh
dnf install -y check check-devel
```
You don't need to install `check-static`. This works for me on `Fedora 38`.

> Also, using `podman` doesn't work... uninstall docker, and install podman and work through the issues

I don't know what you mean about `Podman` not working.

GuChad369:
Do you mean we should not use `Docker` as the container, and instead prefer `Podman`?

jesperpedersen:
"Authentication is required to restart 'docker.service'"...

There is no such thing with podman

jesperpedersen:
Had to `chmod 755 testsuite.sh` too

jesperpedersen:
Nah, either docker or podman is ok - but both need to work

GuChad369:
> Nah, either docker or podman is ok - but both need to work

Sorry about the confusion. I am not familiar with Podman, and I have just learned Docker for this PR. Could you please explain more?

Do I need to rewrite a version for Podman now? I’m still confused about the next step.

jesperpedersen:
Or check for both during `CMakeLists.txt` - which is likely best - and then again during `testsuite.sh`

GuChad369:
Thanks for explaining. I will figure it out.

GuChad369:
Update the `testsuite.sh` to make it work for both `podman` and `Docker`.

The `CMakeLists.txt` will check for `Docker` or `podman`. If neither of these two is installed, the test will not be compiled.

If there are any other problems or areas that need improvement, please let me know. Thanks for your help!

GuChad369:
Currently, the test will not succeed since this `PR` hasn't been committed. If you want to test it, you can change the file `Dockerfile.rocky8` line 83 to the following:

```docker
RUN git clone --branch "test" --single-branch --depth 1 https://github.com/GuChad369/pgmoneta && \
```

CMakeLists.txt
```diff
@@ -106,6 +111,27 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
...
+find_package(Check)
+if (CHECK_FOUND)
+  message(STATUS "check found")
+  add_library(Check::check SHARED IMPORTED)
+  set_target_properties(Check::check PROPERTIES
+    IMPORTED_LOCATION ${CHECK_LIBRARY}
+    INTERFACE_INCLUDE_DIRECTORIES ${CHECK_INCLUDE_DIR})
+else ()
+  set(check FALSE)
+  message(STATUS "check needed. The test suite process will be skipped.")
```
jesperpedersen, line 123: Move this to the next `if`

cmake/FindCheck.cmake
```diff
@@ -0,0 +1,30 @@
...
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(Check DEFAULT_MSG CHECK_INCLUDE_DIR CHECK_LIBRARY)
+
+if (CHECK_FOUND)
+  set(CHECK_LIBRARIES ${CHECK_LIBRARY})
+  set(CHECK_INCLUDE_DIRS ${CHECK_INCLUDE_DIR})
+else ()
+  set(check FALSE)
+  message(STATUS "check needed. The test suite process will be skipped.")
```
jesperpedersen, line 27: You don't need this message

cmake/FindDocker.cmake
```diff
@@ -0,0 +1,20 @@
...
+)
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(Docker DEFAULT_MSG DOCKER_EXECUTABLE)
+
+if (DOCKER_FOUND)
+  set(container TRUE)
+  set(DOCKER_EXECUTABLE ${DOCKER_EXECUTABLE})
+else ()
+  message(STATUS "Docker needed.")
```
jesperpedersen, line 17: You don't need this message

cmake/FindPodman.cmake
```diff
@@ -0,0 +1,20 @@
...
+)
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(Podman DEFAULT_MSG PODMAN_EXECUTABLE)
+
+if (PODMAN_FOUND)
+  set(container TRUE)
+  set(PODMAN_EXECUTABLE ${PODMAN_EXECUTABLE})
+else ()
+  message(STATUS "Podman needed.")
```
jesperpedersen, line 17: You don't need this message

test/testsuite.sh
```diff
@@ -0,0 +1,132 @@
...
+    else
+        return 1
+    fi
+}
+
+run_podman(){
+  dockerfile=${1:-'Dockerfile.rocky8'}
+  version=${2:-13}
+
+  if ! systemctl restart podman; then
```
jesperpedersen, line 44: There is no such thing

jesperpedersen (changes requested):

jesperpedersen:
The test suite failed as well, so test the podman setup

jesperpedersen:
Remember to `chmod 755 testsuite.sh` as well...

GuChad369:
> The test suite failed as well, so test the podman setup

I said this.

Currently, it will not succeed since this PR hasn't been merged, and the source code is incorrect when installing `pgmoneta`.

CMakeLists.txt
GuChad369: Could you please clarify what you mean and where this should be moved to?

jesperpedersen: You have the message in the top-level `CMakeLists.txt` file, so you don't need it here

jesperpedersen:
Test it locally, then change what need to before force pushing

GuChad369:
I tested the current version using both `podman` and `docker`.

Only installed `podman`:

Screenshot 2024-07-06 at 7 03 44 PM

Only installed `docker`:

Screenshot 2024-07-06 at 8 07 32 PM

jesperpedersen:
```
make test
Running tests...
Test project /home/jews/PostgreSQL/pgmoneta/official/build
    Start 1: test_version_13_rocky8
1/8 Test #1: test_version_13_rocky8 ...........***Failed  362.37 sec
    Start 2: test_version_14_rocky8
2/8 Test #2: test_version_14_rocky8 ...........***Failed  353.67 sec
    Start 3: test_version_15_rocky8
3/8 Test #3: test_version_15_rocky8 ...........***Failed  357.56 sec
    Start 4: test_version_16_rocky8
4/8 Test #4: test_version_16_rocky8 ...........***Failed  965.59 sec
    Start 5: test_version_13_rocky9
5/8 Test #5: test_version_13_rocky9 ...........   Passed  369.20 sec
    Start 6: test_version_14_rocky9
6/8 Test #6: test_version_14_rocky9 ...........   Passed  377.56 sec
    Start 7: test_version_15_rocky9
7/8 Test #7: test_version_15_rocky9 ...........   Passed  373.63 sec
    Start 8: test_version_16_rocky9
8/8 Test #8: test_version_16_rocky9 ...........   Passed  400.70 sec

50% tests passed, 4 tests failed out of 8

Total Test time (real) = 3560.34 sec
```

`podman version 5.1.1`

jesperpedersen:
You don't have to compile PostgreSQL - you can just use the packages from yum.postgresql.org

https://github.com/pgmoneta/pgmoneta/blob/main/doc/DEVELOPERS.md#install-postgresql

jesperpedersen:
https://github.com/jesperpedersen/podman-pgsql/blob/master/pgsql14-primary-rocky8/Dockerfile#L35

GuChad369:
> ```
> make test
> Running tests...
> Test project /home/jews/PostgreSQL/pgmoneta/official/build
>     Start 1: test_version_13_rocky8
> 1/8 Test #1: test_version_13_rocky8 ...........***Failed  362.37 sec
>     Start 2: test_version_14_rocky8
> 2/8 Test #2: test_version_14_rocky8 ...........***Failed  353.67 sec
>     Start 3: test_version_15_rocky8
> 3/8 Test #3: test_version_15_rocky8 ...........***Failed  357.56 sec
>     Start 4: test_version_16_rocky8
> 4/8 Test #4: test_version_16_rocky8 ...........***Failed  965.59 sec
>     Start 5: test_version_13_rocky9
> 5/8 Test #5: test_version_13_rocky9 ...........   Passed  369.20 sec
>     Start 6: test_version_14_rocky9
> 6/8 Test #6: test_version_14_rocky9 ...........   Passed  377.56 sec
>     Start 7: test_version_15_rocky9
> 7/8 Test #7: test_version_15_rocky9 ...........   Passed  373.63 sec
>     Start 8: test_version_16_rocky9
> 8/8 Test #8: test_version_16_rocky9 ...........   Passed  400.70 sec
> 
> 50% tests passed, 4 tests failed out of 8
> 
> Total Test time (real) = 3560.34 sec
> ```
> 
> `podman version 5.1.1`

Did you check the log to see if the issue is with the source code?

GuChad369:
> You don't have to compile PostgreSQL - you can just use the packages from yum.postgresql.org

The current method works well and allows us to use the latest version exactly. Why should we change it?

jesperpedersen:
What should I look for in the log ? It is massive.

Using REL_XY_STABLE is of course the most ideal - but it is likely a lot faster to just install the latest packages. Did you check the time difference ?

GuChad369:
> What should I look for in the log ? It is massive.

It is located in /pgmoneta/build/Testing/Temporary/LastTest.log.

Could you please upload the log?

> Using REL_XY_STABLE is of course the most ideal - but it is likely a lot faster to just install the latest packages. Did you check the time difference ?

If you insist, I can change it to using `yum.postgresql.org`. However, we will add more operating systems in the future, and not every OS supports this package. 

We would need to use a different method to install PostgreSQL for other OS. I believe we should maintain consistency in the `Dockerfile`.

The current method is easier to maintain and avoids writing specific Dockerfiles for different operating systems. Of course, we can adjust this approach.

However, as we scale up, maintenance will become more difficult.

jesperpedersen:
[LastTest.log](https://github.com/user-attachments/files/16120525/LastTest.log)

jesperpedersen:
Lets see how long the run will take using the packages from yum.postgresql.org

Rocky 8 / 9 are the primary platforms, so maybe we won't bother with other ones like FreeBSD since it is best effort

jesperpedersen:
You need to do the `chmod 755 test/testsuite.sh` as well

GuChad369:
> You need to do the `chmod 755 test/testsuite.sh` as well - as part of the patch, so it doesn't change after the run

I already added it in the CMakeLists.txt

jesperpedersen:
No, the file itself.

Check `git status` after you do a run - there should be no changes

GuChad369:
> [LastTest.log](https://github.com/user-attachments/files/16120525/LastTest.log)

The reason you have failed is the same as before. Since this PR has not been merged, the source code is still incorrect.

```log
[ 51%] Building C object src/CMakeFiles/pgmoneta.dir/libpgmoneta/security.c.o
/pgmoneta/src/libpgmoneta/security.c: In function 'create_hash_file':
/pgmoneta/src/libpgmoneta/security.c:2894:9: error: implicit declaration of function 'EVP_DigestInit_ex2'; did you mean 'EVP_DigestInit_ex'? [-Werror=implicit-function-declaration]
    if (!EVP_DigestInit_ex2(md_ctx, md, NULL))
         ^~~~~~~~~~~~~~~~~~
         EVP_DigestInit_ex
cc1: all warnings being treated as errors
make[2]: *** [src/CMakeFiles/pgmoneta.dir/build.make:524: src/CMakeFiles/pgmoneta.dir/libpgmoneta/security.c.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:148: src/CMakeFiles/pgmoneta.dir/all] Error 2
make: *** [Makefile:156: all] Error 2
```

jesperpedersen:
Ok.

Lets go with the yum.postgresql.org based setup and get it merged such that we can continue.

We can always improve later with the `REL_XX_STABLE` branch setup

GuChad369:
Currently, the address does not work because they have updated the repository. For more details, please refer to the [PostgreSQL RPM repository GPG key update - 3 January 2024](https://yum.postgresql.org/news/pgdg-rpm-repo-gpg-key-update/).

jesperpedersen:
@devrimgunduz ?

GuChad369:
I am trying to solve the problem.

https://yum.postgresql.org/news/pgdg-rpm-repo-gpg-key-update/

``` consle
8.361 Error: Failed to download metadata for repo 'pgdg-common': repomd.xml GPG signature verification error: Bad GPG signature

```

devrimgunduz:
> I am trying to solve the problem.
> 
> https://yum.postgresql.org/news/pgdg-rpm-repo-gpg-key-update/
> 
> ```
> 8.361 Error: Failed to download metadata for repo 'pgdg-common': repomd.xml GPG signature verification error: Bad GPG signature
> ```

Which OS is this and what's the output of

ls /etc/yum.repos.d/*

GuChad369:
Dockefile
FROM rockylinux:8
[output.log](https://github.com/user-attachments/files/16128766/output.log)
I tried importing the updated GPG key from https://download.postgresql.org/pub/repos/yum/keys/PGDG-RPM-GPG-KEY-RHEL, but it didn't work.

jesperpedersen:
@GuChad369 @devrimgunduz Did you try it the same way as https://github.com/jesperpedersen/podman-pgsql/blob/master/pgsql14-primary-rocky8/Dockerfile#L35 ?

GuChad369:
> https://github.com/jesperpedersen/podman-pgsql/blob/master/pgsql14-primary-rocky8/Dockerfile#L35

Yes, I tried this first, but the error remains the same.

```console
8.361 Error: Failed to download metadata for repo 'pgdg-common': repomd.xml GPG signature verification error: Bad GPG signature
```

jesperpedersen:
@devrimgunduz Ideas ?

devrimgunduz:
I have no idea why people make these basic things harder and harder...

@GuChad369 the output that you pasted is not what I want. I just asked for the lsu output.

Also, why did you import the RPM key manually? This is not required as dnf is responsible for that.

Any chance you can start with a fresh instance,  and follow the very basic instructions here?

https://www.postgresql.org/download/linux/redhat/

Please start a fresh instance, and also please do *not* do anything manually. This is something that works for millions of users.

GuChad369:
> Also, why did you import the RPM key manually? This is not required as dnf is responsible for that.

I found a website that mentioned a change in the source: [PostgreSQL update](https://www.postgresql.org/message-id/4f803fbc884177e5ead5e54dee52db5048d21110.camel%40gunduz.org).

The error indicates a Bad GPG signature, so I tried to see if this was the problem.

devrimgunduz:
> > Also, why did you import the RPM key manually? This is not required as dnf is responsible for that.
> 
> I found a website that mentioned a change in the source: [PostgreSQL update](https://www.postgresql.org/message-id/4f803fbc884177e5ead5e54dee52db5048d21110.camel%40gunduz.org).
> 
> The error indicates a Bad GPG signature, so I tried to see if this was the problem.

Let's go step by step. AFAICS this is RHEL/Rocky 8. Please replace 16 in the second command with any major PostgreSQL version if you want to use something else:

```
dnf -y install https://download.postgresql.org/pub/repos/yum/reporpms/EL-8-x86_64/pgdg-redhat-repo-latest.noarch.rpm
dnf -y install postgresql16-server postgresql16-devel pgmoneta

```

GuChad369:
[output.log](https://github.com/user-attachments/files/16146015/output.log)

``` Dockerfile
RUN dnf -y install https://download.postgresql.org/pub/repos/yum/reporpms/EL-8-x86_64/pgdg-redhat-repo-latest.noarch.rpm && \
    dnf -y install postgresql16-server postgresql16-devel
```

The problem remains: `Bad GPG signature`.

jesperpedersen:
@GuChad369 I just tried https://github.com/jesperpedersen/podman-pgsql/blob/master/pgsql14-primary-rocky8 with `make`, and it worked.

Does it work for you ?

GuChad369:
No, it is the same error `Bad GPG signature`

``` console
pgdg-redhat-repo-42.0-43PGDG          ########################################
Rocky Linux 8 - AppStream                       4.0 MB/s | 9.4 MB     00:02    
Rocky Linux 8 - BaseOS                           66 kB/s | 4.1 MB     01:02    
Rocky Linux 8 - Extras                          810 kB/s | 534 kB     00:00    
Extra Packages for Enterprise Linux 8 - aarch64 4.1 MB/s |  14 MB     00:03    
PostgreSQL common RPMs for RHEL / Rocky / AlmaL 926  B/s | 659  B     00:00    
PostgreSQL common RPMs for RHEL / Rocky / AlmaL 2.4 MB/s | 2.4 kB     00:00    
Importing GPG key 0x08B40D20:
 Userid     : "PostgreSQL RPM Repository <pgsql-pkg-yum@lists.postgresql.org>"
 Fingerprint: D4BF 08AE 67A0 B4C7 A1DB CCD2 40BC A2B4 08B4 0D20
 From       : /etc/pki/rpm-gpg/PGDG-RPM-GPG-KEY-RHEL
PostgreSQL common RPMs for RHEL / Rocky / AlmaL 643  B/s | 659  B     00:01    
Error: Failed to download metadata for repo 'pgdg-common': repomd.xml GPG signature verification error: Bad GPG signature
```

jesperpedersen:
Where is https://github.com/jesperpedersen/podman-pgsql/blob/master/pgsql14-primary-rocky8/Dockerfile#L35 and onwards in your patch ?

GuChad369:
I did not copy that; I just downloaded the code and ”make build” it.

jesperpedersen:
But, I don't see the code in https://github.com/pgmoneta/pgmoneta/pull/321/files#diff-cb057c51e1f9e69db20f45b3fc38b0ab4c3f7b5f26fc4e4b5afa85a27edef5c4R38

GuChad369:
Since I didn’t commit it, I tested it locally. It did not work, so I didn’t commit. The current code still uses the compiled PostgreSQL source code for installation.

Jubilee101:
> I just tried https://github.com/jesperpedersen/podman-pgsql/blob/master/pgsql14-primary-rocky8 with make, and it worked.

Just tried it following the commands in README. Had some bumps but it's building and running normally. 

May I ask your hosting OS environment? I'm working on a Fedora39 VM I instantiated on Azure @GuChad369

Jubilee101:
The log looks somewhat like this:
```
...
pgdg-redhat-repo-42.0-43PGDG          ########################################
Rocky Linux 8 - AppStream                        12 MB/s |  11 MB     00:00
Rocky Linux 8 - BaseOS                           10 MB/s | 5.0 MB     00:00
Rocky Linux 8 - Extras                           28 kB/s |  14 kB     00:00
Extra Packages for Enterprise Linux 8 - x86_64   14 MB/s |  14 MB     00:00
PostgreSQL common RPMs for RHEL / Rocky / AlmaL 850  B/s | 659  B     00:00
PostgreSQL common RPMs for RHEL / Rocky / AlmaL 2.4 MB/s | 2.4 kB     00:00
Importing GPG key 0x08B40D20:
 Userid     : "PostgreSQL RPM Repository <pgsql-pkg-yum@lists.postgresql.org>"
 Fingerprint: D4BF 08AE 67A0 B4C7 A1DB CCD2 40BC A2B4 08B4 0D20
 From       : /etc/pki/rpm-gpg/PGDG-RPM-GPG-KEY-RHEL
PostgreSQL common RPMs for RHEL / Rocky / AlmaL 262 kB/s | 462 kB     00:01
PostgreSQL 16 for RHEL / Rocky / AlmaLinux 8 -  983  B/s | 659  B     00:00
PostgreSQL 16 for RHEL / Rocky / AlmaLinux 8 -  2.4 MB/s | 2.4 kB     00:00
Importing GPG key 0x08B40D20:
 Userid     : "PostgreSQL RPM Repository <pgsql-pkg-yum@lists.postgresql.org>"
 Fingerprint: D4BF 08AE 67A0 B4C7 A1DB CCD2 40BC A2B4 08B4 0D20
 From       : /etc/pki/rpm-gpg/PGDG-RPM-GPG-KEY-RHEL
PostgreSQL 16 for RHEL / Rocky / AlmaLinux 8 -  168 kB/s | 381 kB     00:02
PostgreSQL 15 for RHEL / Rocky / AlmaLinux 8 -  613  B/s | 659  B     00:01
PostgreSQL 15 for RHEL / Rocky / AlmaLinux 8 -  2.4 MB/s | 2.4 kB     00:00
...
Dependencies resolved.
...
```

GuChad369:
> May I ask your hosting OS environment? I'm working on a Fedora39 VM I instantiated on Azure @GuChad369

My computer is running `MacOS`, and I am using a VM with `Fedora 38` and `Ubuntu 22.04.4 LTS`. I followed the exact commands in the `README` without changing any source code or commands.

I tested the setup on all three operating systems, but all showed the same error.

``` console
Error: Failed to download metadata for repo 'pgdg-common': repomd.xml GPG signature verification error: Bad GPG signature
```

jesperpedersen:
https://stackoverflow.com/questions/77782695/failed-to-download-metadata-for-repository-pgdg-common-on-centos-7 ?

GuChad369:
> https://stackoverflow.com/questions/77782695/failed-to-download-metadata-for-repository-pgdg-common-on-centos-7 ?

I know this, but none of the solutions worked for me.

GuChad369:
``` sh
cd /etc/yum.repos.d/
mv pgdg-redhat-all.repo pgdg-redhat-all.repo.old
yum update -y
```

This method is the only one that has worked for the `Bad GPG signature` issue. However, after changing the repository to the old one, there is no match to install PostgreSQL.

jesperpedersen:
@devrimgunduz Any new ideas ?

GuChad369:
If we implement the extension, how do we use it in `pgmoneta`?

In `pgexporter`, the `SSH` and `file descriptor` (FD) are stored in the configuration, but in `pgmoneta`, we do not do this.

Each time, we use `pgmoneta_server_authenticate` to get the `SSH` and `socket` in `main.c`.

Should we just pass the `SSH` and `socket` for the `extension`, or should we get the connection inside the extension?

jesperpedersen:
@GuChad369 This is probably more for #328 - I don't think we need to cache the `SSL` and `FD` descriptors to begin with - just open a new one each time

jesperpedersen:
@GuChad369 Can you do the initial patch with just rocky9 using the `REL_XX_STABLE` branches, then we can look at doing rocky8 and yum.postgresql.org as an enhancement

GuChad369:
Currently, the only problem is the running time. Downloading the package and compiling PostgreSQL consumes a lot of time. However, it works well now and can run on any OS. Additionally, it will not be affected by updates in PostgreSQL since we use the source code mirror repository. I will continue to try to fix the bug with the `yum.postgresql.org` problem.

Jubilee101:
> Currently, the only problem is the running time. Downloading the package and compiling PostgreSQL consumes a lot of time. However, it works well now and can run on any OS. Additionally, it will not be affected by updates in PostgreSQL since we use the source code mirror repository. I will continue to try to fix the bug with the yum.postgresql.org problem.

Yeah we get how frustrating this gets.. especially when we've been going on and on for weeks. The running time is a problem but this should be a very good start. I have a strong feeling that the GPG problem has something to do with MacOS hosted VM.

GuChad369:
Yes, since they have updated the GPG address. Additionally, it is not easy to use VMs on macOS. I use Parallels and can only install restricted OSes.

jesperpedersen:
Rebased, and merged.

PR #336, opened 2024-07-11
[#319] Create extension.h|.c to call the extension functions

GuChad369:
Create a basic `extension.h|.c` to call our extension functions and store the result in the struct `query_response`.

This is a basic structure for the extension. We can discuss what we need to improve.


jesperpedersen:
Rebased, and merged.

Issue #328, opened 2024-06-28
Get extra files from the server


jesperpedersen:
Use the `pgmoneta_ext` extension to get extra files / directories (recursive) from the PostgreSQL server instance.

Of course there has to be read priviledge for it for the user.

A test case could be writing something to `/tmp` and then getting it.

Everything should go into an `/extras` directory under the backup, and be displayed as part of the `info` command.

Configuration should go under each server - separated by `,` like

```
extra = /tmp/myfile,/tmp/mydir/*,/tmp/mydirs/**
```

should translate into

```
cd extras/
/tmp/myfile
/tmp/mydir/*
/tmp/mydirs/**
```

jesperpedersen:
@GuChad369 Can you take this ?

GuChad369:
Of course, I will try, but it may take a long time to finish.

I suppose we need to first complete `extension.h|.c` which will be necessary.

jesperpedersen:
Yeah, and we need the test suite in first - it isn't a problem

PR #341, opened 2024-07-17
[#328] Get extra files

GuChad369:
Currently, we can retrieve the byte content of a file and list all file paths under a directory.

Is there already a function to transfer the file from the server to the backend in `pgmoneta`, or how do we handle file transfers in `pgmoneta`? 

I am trying to transfer the file to the backend now.

### DIFF
--- a/doc/CONFIGURATION.md
+++ b/doc/CONFIGURATION.md
@@ -81,6 +81,7 @@ See a [sample](./etc/pgmoneta.conf) configuration for running `pgmoneta` on `loc
 | host | | String | Yes | The address of the PostgreSQL instance |
 | port | | Int | Yes | The port of the PostgreSQL instance |
 | user | | String | Yes | The replication user name |
+| extra | | String | No | The source directory for retrieval on the server side (details are in the extra parameter)|
 | wal_slot | | String | Yes | The replication slot for WAL |
 | create_slot | no | Bool | No | Create a replication slot for this server. Valid values are: yes, no |
 | follow | | String | No | Failover to this server if follow server fails |
@@ -103,6 +104,20 @@ have access to the `postgres` database in order to get the necessary configurati
 Note, that PostgreSQL 13+ is required, as well as having `wal_level` at `replica` or `logical` level.
 
 Note, that if `host` starts with a `/` it represents a path and `pgmoneta` will connect using a Unix Domain Socket.
+
+### extra parameter
+
+The `extra` configuration is set in the server section. It is not required, but if you configure this parameter, when you perform a backup using the CLI `pgmoneta-cli -c pgmoneta.conf backup primary`, it will also copy all specified files on the server side and send them back to the client side.
+
+This `extra` feature requires the server side to install the [pgmoneta_ext](https://github.com/pgmoneta/pgmoneta_ext) extension and also make the user `repl` a `SUPERUSER` (this will be improved in the future). Currently, this feature is only available to the `SUPERUSER` role.
+
+You can set up `pgmoneta_ext` by following the [README](https://github.com/pgmoneta/pgmoneta_ext/blob/main/README.md) to easily install the extension. There are also more detailed instructions available in the [DEVELOPERS](https://github.com/pgmoneta/pgmoneta_ext/blob/main/doc/DEVELOPERS.md) documentation.
+
+The format for the `extra` parameter is a path to a file or directory. You can list more than one file or directory separated by commas. The format is as follows:
+
+```ini
+extra = /tmp/myfile1, /tmp/myfile2, /tmp/mydir1, /tmp/mydir2
+```
 
 # pgmoneta_users configuration
 
--- a/doc/manual/02-installation.md
+++ b/doc/manual/02-installation.md
@@ -198,3 +198,64 @@ where `$HOME` is your home directory.
 ### Build
 
 These packages will be detected during `cmake` and built as part of the main build.
+
+## Extension installation
+
+When you configure the `extra` parameter in the server section of `pgmoneta.conf`, it requires the server side to have the [pgmoneta_ext][pgmoneta_ext] extension installed to make it work.
+
+The following instructions can help you easily install `pgmoneta_ext`. If you encounter any problems, please refer to the more detailed instructions in the [DEVELOPERS][ext_developers] documentation.
+
+### Install pgmoneta_ext
+
+After you have successfully installed `pgmoneta`, the following commands will help you install `pgmoneta_ext`:
+
+```sh
+dnf install -y pgmoneta_ext
+```
+
+You need to add the `pgmoneta_ext` library for PostgreSQL in `postgrersql.conf` as well:
+
+```ini
+shared_preload_libraries = 'pgmoneta_ext'
+```
+
+And remember to restart PostgreSQL to make it work.
+
+### Verify success
+
+You can use the `postgres` role to test.
+
+1. Log into PostgreSQL
+
+    ``` sh
+    psql
+    ```
+
+2. Create a new test database
+
+    ``` sql
+    CREATE DATABASE testdb;
+    ```
+
+3. Enter the database
+
+    ``` sql
+    \c testdb
+    ```
+
+4. Follow the SQL commands below to check the function
+
+    ``` sql
+    DROP EXTENSION IF EXISTS pgmoneta_ext;
+    CREATE EXTENSION pgmoneta_ext;
+    SELECT pgmoneta_ext_version();
+    ```
+
+You should see
+
+``` console
+ pgmoneta_ext_version 
+----------------------
+ 0.1.0
+(1 row)
+```
--- a/doc/manual/user-02-configuration.md
+++ b/doc/manual/user-02-configuration.md
@@ -81,6 +81,7 @@ See a [sample][sample] configuration for running [**pgmoneta**][pgmoneta] on `lo
 | host | | String | Yes | The address of the PostgreSQL instance |
 | port | | Int | Yes | The port of the PostgreSQL instance |
 | user | | String | Yes | The replication user name |
+| extra | | String | No | The source directory for retrieval on the server side (details are in the extra parameter) |
 | wal_slot | | String | Yes | The replication slot for WAL |
 | create_slot | no | Bool | No | Create a replication slot for this server. Valid values are: yes, no |
 | follow | | String | No | Failover to this server if follow server fails |
@@ -102,6 +103,20 @@ The `user` specified must have the `REPLICATION` option in order to stream the W
 Note, that PostgreSQL 13+ is required, as well as having `wal_level` at `replica` or `logical` level.
 
 Note, that if `host` starts with a `/` it represents a path and [**pgmoneta**][pgmoneta] will connect using a Unix Domain Socket.
+
+### extra parameter
+
+The `extra` configuration is set in the server section. It is not required, but if you configure this parameter, when you perform a backup using the CLI `pgmoneta-cli -c pgmoneta.conf backup primary`, it will also copy all specified files on the server side and send them back to the client side.
+
+This `extra` feature requires the server side to install the [pgmoneta_ext][pgmoneta_ext] extension and also make the role `repl` a `SUPERUSER` (this will be improved in the future). Currently, this feature is only available to the `SUPERUSER` role.
+
+You can set up `pgmoneta_ext` by following the [README][ext_readme] to easily install the extension. There are also more detailed instructions available in the [DEVELOPERS][ext_developers] documentation.
+
+The format for the `extra` parameter is a path to a file or directory. You can list more than one file or directory separated by commas. The format is as follows:
+
+```ini
+extra = /tmp/myfile1, /tmp/myfile2, /tmp/mydir1, /tmp/mydir2
+```
 
 ## pgmoneta_users configuration
 
--- a/src/include/extension.h
+++ b/src/include/extension.h
@@ -39,32 +39,79 @@ extern "C" {
 #include <stdbool.h>
 #include <stdlib.h>
 
+#define MAX_QUERY_LENGTH 1024
+
 /**
- * Return the extension version number
- * @param server The server
+ * Check if the server has the extension installed
+ * @param ssl The SSL structure
+ * @param socket The socket
  * @param qr The query result
  * @return 0 upon success, otherwise 1
  */
 int
-pgmoneta_ext_version(int server, struct query_response** qr);
+pgmoneta_ext_is_installed(SSL* ssl, int socket, struct query_response** qr);
+
+/**
+ * Return the extension version number
+ * @param ssl The SSL structure
+ * @param socket The socket
+ * @param qr The query result
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_ext_version(SSL* ssl, int socket, struct query_response** qr);
 
 /**
  * Trigger WAL switch operation
- * @param server The server
+ * @param ssl The SSL structure
+ * @param socket The socket
  * @param qr The query result
  * @return 0 upon success, otherwise 1
  */
 int
-pgmoneta_ext_switch_wal(int server, struct query_response** qr);
+pgmoneta_ext_switch_wal(SSL* ssl, int socket, struct query_response** qr);
 
 /**
  * Force PostgreSQL to carry out an immediate checkpoint
- * @param server The server
+ * @param ssl The SSL structure
+ * @param socket The socket
  * @param qr The query result
  * @return 0 upon success, otherwise 1
  */
 int
-pgmoneta_ext_checkpoint(int server, struct query_response** qr);
+pgmoneta_ext_checkpoint(SSL* ssl, int socket, struct query_response** qr);
+
+/**
+ * Check if the current user is a superuser
+ * @param ssl The SSL structure
+ * @param socket The socket
+ * @param qr The query result
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_ext_priviledge(SSL* ssl, int socket, struct query_response** qr);
+
+/**
+ * Retrieve the bytes of the specified file
+ * @param ssl The SSL structure
+ * @param socket The socket
+ * @param file_path The path to the file
+ * @param qr The query result
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_ext_get_file(SSL* ssl, int socket, const char* file_path, struct query_response** qr);
+
+/**
+ * Retrieve all file paths under the specified directory
+ * @param ssl The SSL structure
+ * @param socket The socket
+ * @param file_path The path to the directory
+ * @param qr The query result
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_ext_get_files(SSL* ssl, int socket, const char* file_path, struct query_response** qr);
 
 #ifdef __cplusplus
 }
--- a/src/include/pgmoneta.h
+++ b/src/include/pgmoneta.h
@@ -68,7 +68,9 @@ extern "C" {
 #define MAX_PATH 1024
 #define MISC_LENGTH 128
 #define MAX_COMMENT 2048
+#define MAX_EXTRA_PATH 8192
 
+#define MAX_EXTRA 64
 #define NUMBER_OF_SERVERS 64
 #define NUMBER_OF_USERS   64
 #define NUMBER_OF_ADMINS   8
@@ -236,6 +238,8 @@ struct server
    int backup_max_rate;     /**< Number of tokens added to the bucket with each replenishment for backup. */
    int network_max_rate;    /**< Number of bytes of tokens added every one second to limit the netowrk backup rate */
    int manifest;  /**< The manifest hash algorithm */
+   int number_of_extra; /**< The number of source directory*/
+   char extra[MAX_EXTRA][MAX_EXTRA_PATH];  /**< Source directory*/
 } __attribute__ ((aligned (64)));
 
 /** @struct user
--- a/src/libpgmoneta/configuration.c
+++ b/src/libpgmoneta/configuration.c
@@ -82,6 +82,8 @@ static int restart_string(char* name, char* e, char* n);
 static bool is_empty_string(char* s);
 static int remove_leading_whitespace_and_comments(char* s, char** trimmed_line);
 
+static void split_extra(const char* extra, char res[MAX_EXTRA][MAX_EXTRA_PATH], int* count);
+
 /**
  *
  */
@@ -294,6 +296,30 @@ pgmoneta_read_configuration(void* shm, char* filename)
                         max = MAX_USERNAME_LENGTH - 1;
                      }
                      memcpy(&srv.username, value, max);
+                  }
+                  else
+                  {
+                     unknown = true;
+                  }
+               }
+               else if (!strcmp(key, "extra"))
+               {
+                  if (strlen(section) > 0)
+                  {
+                     max = strlen(section);
+                     if (max > MISC_LENGTH - 1)
+                     {
+                        max = MISC_LENGTH - 1;
+                     }
+                     memcpy(&srv.name, section, max);
+                     max = strlen(value);
+                     if (max > MAX_PATH - 1)
+                     {
+                        max = MAX_PATH - 1;
+                     }
+                     int count = 0;
+                     split_extra(value, srv.extra, &count);
+                     srv.number_of_extra = count;
                   }
                   else
                   {
@@ -3257,4 +3283,26 @@ error:
    free(result); // Free memory in case of error
    *trimmed_line = NULL;
    return 1;
+}
+
+static void
+split_extra(const char* extra, char res[MAX_EXTRA][MAX_EXTRA_PATH], int* count)
+{
+   int i = 0;
+   char temp[MAX_BUFFER_SIZE];
+   char* token;
+
+   strcpy(temp, extra);
+   token = strtok(temp, ",");
+
+   while (token != NULL)
+   {
+      // trim_spaces(token);
+      pgmoneta_remove_whitespace(token);
+      strcpy(res[i], token);
+      token = strtok(NULL, ",");
+      i++;
+   }
+
+   *count = i;
 }
--- a/src/libpgmoneta/message.c
+++ b/src/libpgmoneta/message.c
@@ -29,6 +29,7 @@
 /* pgmoneta */
 #include <pgmoneta.h>
 #include <achv.h>
+#include <extension.h>
 #include <logging.h>
 #include <manifest.h>
 #include <memory.h>
@@ -42,6 +43,7 @@
 #include <string.h>
 #include <unistd.h>
 #include <openssl/err.h>
+#include <openssl/evp.h>
 #include <openssl/ssl.h>
 #include <sys/time.h>
 #include <stdio.h>
@@ -57,6 +59,10 @@ static int get_number_of_columns(struct message* msg);
 static int get_column_name(struct message* msg, int index, char** name);
 
 static bool is_server_side_compression(void);
+
+static unsigned char* decode_base64(const char* base64_data, int* decoded_len);
+static char** get_paths(const char* data, int* count);
+static void extract_file_name(const char* path, char* file_name, char* file_path);
 
 int
 pgmoneta_read_block_message(SSL* ssl, int socket, struct message** msg)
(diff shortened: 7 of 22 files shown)


jesperpedersen:
No, you have to do a message for it, or use something like https://github.com/pgexporter/pgexporter/blob/main/src/include/queries.h to receive the data.

https://www.postgresql.org/docs/16/protocol-message-formats.html will give you the protocol - but there are plenty of examples in message.c

GuChad369:
`query_execute` can handle all extension functions except `pgmoneta_ext_get_file`, which returns `bytes`. The result is stored in a struct `query_response`, which stores data using `char**`.

Should I convert the `bytea` to `char` on the extension side and then convert `char` back to `bytea` on the core side? 

Or should I modify the tuple structure to accommodate the byte result, or create a new structure to handle the bytea result?

jesperpedersen:
Well, you can either type-cast the `byte` to `char`, or change the signature to use `void*`.

I would start with the former, and see how it goes

GuChad369:
The update can retrieve all text files and directories (recursively) from the PostgreSQL server instance into the `/extras` directory. 

However, it does not handle binary files. I am attempting to cast `bytes` to `char`. I have implemented custom encode and decode functions to achieve this, but I wonder if a built-in function can accomplish this.

To avoid proceeding incorrectly, I commit this version now and continue working on fixing the problem.

jesperpedersen:
Correct, we have to handle all file types - text and binary

GuChad369:
This update can now handle both text and binary files. The basic idea is to cast the binary data to characters on the ext side, and after receiving the result on the core side, cast the characters back to binary. This successfully allows for copying files from the server.

jesperpedersen:
You can use OpenSSL for base64, right ?

Get the configuration (only per server), and documentation in such that we can start to test

GuChad369:
Could you please clarify what you mean by "getting the configuration"? Why do we need to obtain the configuration?

And how should we document this single function?

jesperpedersen:
You need to add this functionality into `pgmoneta.conf`, and then make it part of the backup. So,

```
[pgmoneta]
...

[primary]
...
extra = /home/postgres/configuration/**
```

It should probably also be part of the CLI `info` command

GuChad369:
I write files into `base_dir/extras/`, which is under the backup directory. Did you mean we need to get the user's specified directory and write the files into that directory instead?

```C
base_dir = (char*)malloc((strlen(config->base_dir) + 1) * sizeof(char));
snprintf(extras_dir, strlen(base_dir) + strlen("/extras/") + 1, "%s/extras/", base_dir);
```

> It should probably also be part of the CLI `info` command

I'm not very familiar with this. Could you explain what you mean and what result you want to achieve?

jesperpedersen:
Yes, the files would be under `.../primary/backup/<timestamp>/`, and

```
backup.info  backup.manifest  data/ extra/
```

The `info` command list detailed information about the backup, so it would list all the files under `extra/` as part of that

GuChad369:
I need to add a line in `pgmoneta.conf` under the `[primary]` section to get the user's `extra` directory. After that, I will write the files into that specified directory, correct?

Let's implement this first.

jesperpedersen:
Yeah, but directories - a comma separated list. See the `struct server` and do something like `int number_of_extras` and `char[MAX_EXTRAS][MAX_PATH]`

GuChad369:
This is the current function, which retrieves files under `base_path/dir_file` from the server and writes them into the `extras/ directory`.

``` C
int
pgmoneta_get_extra_files(int server, char* base_path, char* dir_file);
```

jesperpedersen:
Yeah, you can use the full path under `extra/` - we need to make sure that there are no conflicts upon writing the files out

jesperpedersen:
And, of course only absolute paths - we don't really know what a relative path is...

GuChad369:
> Yeah, but directories - a comma separated list. See the `struct server` and do something like `int number_of_extras` and `char[MAX_EXTRAS][MAX_PATH]`

If there are several directories, which one should we choose to write into? Or should we change the current function to handle this?
```C
int
pgmoneta_get_extra_files(int server, char* base_path, char* dir_file, char* target_dir)
```
When we use this function, we pass the target directory into it.

This function will write files into `target_dir` under a directory named using a `timestamp` to avoid conflicts.

Jubilee101:
> If there are several directories, which one should we choose to write into?

I think the directories were source directory on the server side. So if you want to get files under `/home/postgres/config/` , add the path in the config, and the expected behavior is probably retrieving the files and write it into dir like `extra/home/postgres/config` and so on... The config for getting files outside of the `data/` directory.

> The info command list detailed information about the backup, so it would list all the files under extra/ as part of that

I think this can wait a bit? @jesperpedersen

GuChad369:
Okay, I see. Thanks for explaining.

jesperpedersen:
For the `info` command it should be sending `number_of_extras` and then each string over, so it shouldn't be a big problem

GuChad369:
- Use OpenSSL to decode Base64.
- Add a new configuration `extras`, that allows the user to list source directories separated by commas.
- Update the function `pgmoneta_get_extra_files` to retrieve all files from the directories specified in `extras` and then write them into `/tmp/pgmoneta/extras/`. This target directory depends on the `base_dir`.

jesperpedersen:
The `extra/` files is done per backup so they have to be part of that

GuChad369:
> The `extra/` files is done per backup so they have to be part of that

Should we call `pgmoneta_get_extra_files` each time we perform a backup?

GuChad369:
I'm not sure if this is correct, but I am trying to call `pgmoneta_get_extra_files` each time a backup is performed.

jesperpedersen:
Yes, you need to do that, and the files go under `<server>/backup/<timestamp>/extra` with their full path

Jubilee101:
> I'm not sure if this is correct, but I am trying to call pgmoneta_get_extra_files each time a backup is performed.

I believe that is the idea. Perhaps this should be a best effort process in the sense that the action may fail because the extension is not installed or the configured path is wrong, and it should not bring down the whole backup process. And also be aware that some versions of servers may not support the extension.

jesperpedersen:
Yes, if the extension isn't installed then it should just result a `WARN` statement, then same with incorrect paths

GuChad369:
> I believe that is the idea. Perhaps this should be a best effort process in the sense that the action may fail because the extension is not installed or the configured path is wrong, and it should not bring down the whole backup process. And also be aware that some versions of servers may not support the extension.

How can we avoid impacting the backup process, considering we need to write the file into the same `<timestamp>` directory as the backup? We have to call this function inside the backup to get the `timestamp`.

Would it be a good idea to add a try-catch block to prevent any potential issues from affecting the backup process?

Jubilee101:
> How can we avoid impacting the backup process, considering we need to write the file into the same <timestamp> directory as the backup? We have to call this function inside the backup to get the timestamp.

Not sure if I get the question, but the backup is written into `<timestamp>/data/` while the extra files go to `<timestamp>/extra`. They should not affect one another. The timestamp is settled before the backup process begins, it's named as `identifier` in the code. 

> Would it be a good idea to add a try-catch block to prevent any potential issues from affecting the backup process?

If `pgmoneta_get_extra_files` returns error just log warn and continue, do not go into error section. That should do the trick. Or if you'd like, this could be a standalone workflow that could be optionally mounted in the backup worflow.

jesperpedersen:
Yes, just do it as a workflow node in the pipeline - detect if the extension is installed or not.

So, here

https://github.com/pgmoneta/pgmoneta/blob/main/src/libpgmoneta/workflow.c#L115

jesperpedersen:
The `extra/` can be compressed and encrypted as well

GuChad369:
Okay, I will give it a try. Since I'm not familiar with the workflow, this might take a while to achieve. I need to first understand the process of the workflow.

jesperpedersen:
Take your time to understand it - this is a very important feature

Jubilee101:
> Okay, I will give it a try. Since I'm not familiar with the workflow, this might take a while to achieve. I need to first understand the process of the workflow.

No problem, take your time. For your reference, [this commit](https://github.com/pgmoneta/pgmoneta/commit/2c7cd607bbe41a70f6a1c33d4dd87b0754923708) added a all-new manifest workflow node and mounted it to the backup process. Focus on `workflow.h/c` and `wf_manifest.c`, you'll find the pattern.

GuChad369:
Thank you for your help.

GuChad369:
I have simulated the `manifest` to create an `extra` workflow mounted in the backup workflow. It works successfully for me. However, I need help understanding why all files have the suffix `.zstd` after being written into the directory.

I need to fix this issue to ensure the files open correctly, as they currently cannot be opened properly.

jesperpedersen:
I don't understand.

If compression and encryption is enabled then `extra/` should be "affected" by that

src/include/extension.h
```diff
@@ -39,32 +39,79 @@ extern "C" {
...
+
+/**
+ * Check if the server has the extension installed
+ * @param ssl The SSL structure
+ * @param socket The socket
+ * @param qr The query result
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_ext_install(SSL* ssl, int socket, struct query_response** qr);
```
jesperpedersen, line 52: `pgmoneta_ext_is_installed`

src/libpgmoneta/message.c
```diff
@@ -2856,3 +2862,267 @@ pgmoneta_receive_manifest_file(SSL* ssl, int socket, struct stream_buffer* buffe
...
+pgmoneta_receive_extra_files(SSL* ssl, int socket, char* username, char* source_dir, char* target_dir)
+{
+   int count = 0;
+   char** paths = NULL;
+   struct query_response* qr = NULL;
+
+   pgmoneta_ext_priviledge(ssl, socket, &qr);
+   if (qr != NULL && qr->tuples != NULL && qr->tuples->data != NULL && qr->tuples->data[0] != NULL && qr->tuples->data[0][0] == 'f')
+   {
+      pgmoneta_log_warn("retrieving extra files: User %s is not SUPERUSER", username);
```
jesperpedersen, line 2876: retrieving -> Retrieving

jesperpedersen: All of them

src/libpgmoneta/message.c
```diff
@@ -2856,3 +2862,267 @@ pgmoneta_receive_manifest_file(SSL* ssl, int socket, struct stream_buffer* buffe
...
+   {
+      return NULL;
+   }
+   cleaned_data = (char*)malloc(data_length - 1);
+   strncpy(cleaned_data, data + 1, data_length - 2);
+   cleaned_data[data_length - 2] = '\0';
+
+   if (strlen(cleaned_data) == 0)
+   {
+      *count = 0;
```
jesperpedersen, line 3022: Maybe do an `error:` section

src/libpgmoneta/wf_extra.c
```diff
@@ -0,0 +1,193 @@
...
+   if (usr == -1)
+   {
+      pgmoneta_log_info("User not found for server: %d", server);
+      goto error;
+   }
+
+   // establish a connection, with replication flag set
+   if (pgmoneta_server_authenticate(server, "postgres", config->users[usr].username, config->users[usr].password, false, &ssl, &socket) != AUTH_SUCCESS)
+   {
+      pgmoneta_log_info("Invalid credentials for %s", config->users[usr].username);
```
jesperpedersen, line 126: Print the server name too

src/libpgmoneta/wf_extra.c
```diff
@@ -0,0 +1,193 @@
...
+   {
+      pgmoneta_log_warn("extra failed: Server %s does not have the pgmoneta_ext extension installed.", config->users[usr].username);
+      goto error;
+   }
+   pgmoneta_free_query_response(qr);
+   qr = NULL;
+
+   for (int i = 0; i < config->servers[server].number_of_extra; i++)
+   {
+      printf("Start copy source: %s\n", config->servers[server].extra[i]);
```
jesperpedersen, line 141: Remove all `printf`

jesperpedersen (changes requested):

jesperpedersen:
#345

GuChad369:
I suspect the `.zstd` suffix problem is related to compression and encryption. I will address this issue in the next pull request for #345.

jesperpedersen:
I'm not following... we want compression and encryption on `extra/`

GuChad369:
Do we need to handle compression and encryption in this pull request?

jesperpedersen:
You can split it into multiple pull requests against the same issue

jesperpedersen:
Most of it should copy'n'paste from `data/` functionality

jesperpedersen:
Try and follow the https://github.com/pgmoneta/pgmoneta/blob/main/src/libpgmoneta/workflow.c#L95 workflow, so you will hook into the same files, and just use `extra/` instead

GuChad369:
I tested several files. After configuring `compression = gzip` and `encryption = aes-128-ctr`, the results were already compressed and encrypted.

Screenshot 2024-07-25 at 12 03 42 PM

jesperpedersen:
Then you are done with that part

jesperpedersen:
Send a PTAL when you are happy with the patch(es)

GuChad369:
Are there any other features that we need to add to this function?

jesperpedersen:
@Jubilee101 and I will look at this

doc/CONFIGURATION.md
```diff
@@ -81,6 +81,7 @@ See a [sample](./etc/pgmoneta.conf) configuration for running `pgmoneta` on `loc
 | host | | String | Yes | The address of the PostgreSQL instance |
 | port | | Int | Yes | The port of the PostgreSQL instance |
 | user | | String | Yes | The replication user name |
+| extra | | String | No | The source directory for retrieval on the server side |
```
jesperpedersen, line 84: You need to describe the requirements for using this - role change, brief setup of the extension and so on.

You also need describe the format of the string with `/path/to/file`, `/path/to/directory/*` and recursive `/path/to/directory/**`

src/libpgmoneta/configuration.c
```diff
@@ -82,6 +82,9 @@ static int restart_string(char* name, char* e, char* n);
 static bool is_empty_string(char* s);
 static int remove_leading_whitespace_and_comments(char* s, char** trimmed_line);
 
+static void trim_spaces(char* str);
```
jesperpedersen, line 85: `char* pgmoneta_remove_whitespace(char* orig);` ?

jesperpedersen (changes requested):

jesperpedersen:
* doc/GETTING_STARTED.md - brief description
* doc/CONFIGURATION.md - full description, plus a new section explaining the extension setup
* doc/tutorial - a new tutorial
* doc/man/pgmoneta.conf.5.rst - same as doc/CONFIGURATION.md
* doc/manual/user-01-quickstart.md - same as doc/GETTING_STARTED.md
* doc/manual/user-02-configuration.md - same as doc/CONFIGURATION.md
* pgmoneta.spec - new tutorial

jesperpedersen:
For the manual you can add a new section about the extension installation to doc/manual/02-installation.md

jesperpedersen:
You can look at https://github.com/pgmoneta/pgmoneta/commit/9a35f82ef3514ee1fa4dc5fb92d61fce378133f9 of how to add `extra` to the `info` command

src/libpgmoneta/configuration.c
GuChad369: `remove_leading_whitespace_and_comments` does not remove the trailing whitespace.

GuChad369: I didn't know about this function, so thanks for reminding me. I changed to using `pgmoneta_remove_whitespace`. While testing, I encountered an error in `bzip2_compression.c`. If I am not mistaken, I believe there is a logic error in the current functions within `bzip2_compression.c`, such as in the `bzip2_compress` function. This function will always return 1. I found this error while working on PR #345. Since I am unfamiliar with bzip2, I added a new function to make the test successful.

GuChad369:
> You can look at [9a35f82](https://github.com/pgmoneta/pgmoneta/commit/9a35f82ef3514ee1fa4dc5fb92d61fce378133f9) of how to add `extra` to the `info` command

Can you explain the `info` command? I am still confused about the desired behavior you want. I need to know this to test. Should I only focus on `info.h|.c`? Just add the number of extras and their file paths?

jesperpedersen:
So, if `extra` isn't enabled to write out an empty string, otherwise you will write out all the file names

src/libpgmoneta/configuration.c
jesperpedersen: If you found a bug then you can create an issue for it, and submit it separately

GuChad369: I don't know if it is a bug or if I was wrong.

GuChad369:
> So, if `extra` isn't enabled, you will write out an empty string; otherwise, you will write out all the file names.

So, your desired behavior is that when the user calls
```sh
pgmoneta-cli -c pgmoneta.conf status details
```

it will show:
```console
Mode             : Running
Used space       : 16 MB
Free space       : 582.7 MB
Total space      : 976.9 MB
Workers          : 0
Number of servers: 1
Server           : primary
  Retention      : 7 day(s) 
  Workers        : 0
  Backups        : 0
  Space          : 16 MB
  Hot standby    : 0
  Number of extras: 2
  Extras: /tmp/myfile,/tmp/mydir/*,/tmp/mydirs/**
```

jesperpedersen:
```
/pgmoneta-cli -c pgmoneta.conf info primary newest
Label                : 20240805165846
WAL                  : 000000010000000000000004
Commments            : mykey1|mycomment3,mykey2|mycomment2
Extra: myfile, ...
Backup size          : 8531968
Restore size         : 48799744
Elapsed time         : 2
Version              : 16
Minor version        : 3
Keep                 : 0
Valid                : 1
Number of tablespaces: 0
Start LSN Hi32       : 0
Start LSN Lo32       : 4000028
End LSN Hi32         : 0
End LSN Lo32         : 4000138
Checkpoint LSN Hi32  : 0
Checkpoint LSN Lo32  : 4000098
Start timeline       : 1
End timeline         : 1
```

GuChad369:
I have updated all the documentation as required.

The `info` command works well now, and it only shows the files that were copied successfully. Any files that failed to copy will not be shown in the `info`. If no files were copied, it will show as empty.

**Empty:**
Screenshot1

**Not Empty:**
Screenshot2

jesperpedersen:
Why are there 3 `log.txt` ? Anyway, dispkay them with relative path from `extra/`

GuChad369:
To display the relative path, I may need to change the source code of `pgmoneta_update_info_string`. Is this okay? Currently, it can't handle large strings.

```ini
STATUS=1
LABEL=20240808094227
TABLESPACES=0
PGMONETA_VERSION=0.14.0
COMMENTS=
WAL=000000010000000000000062
RESTORE=48771072
VERSION=15
MINOR_VERSION=1
KEEP=0
START_WALPOS=0/62000028
END_WALPOS=0/62000100
START_TIMELINE=1
END_TIMELINE=1
HASH_ALGORITM=3
CHKPT_WALPOS=0/62000060
EXTRA=/home/postgres/pgsql/pg_replslot/repl/state, /home/postgres/pgsql/pg_replslot/log.txt, /home/postgres/pgsql/pg_replslot/tBACKUP=5251072
ELAPSED=1
```

jesperpedersen:
Yeah, you can just use `pgmoneta_append()` for the string

GuChad369:
I first get all the extra content and then write it into `backup.info` together using the function `pgmoneta_update_info_string`, but this is not working.

jesperpedersen:
Did you change `buffer[128]`, `line[128]`, k[MISC_LENGTH] and v[MISC_LENGTH] to `char*` ?

... you may have to change the other functions too to use `char*` instead of a static size ...

GuChad369:
> Did you change `buffer[128]`, `line[128]`, `k[MISC_LENGTH]` and `v[MISC_LENGTH]` to `char*` ?
> 
> ... you may have to change the other functions too to use `char*` instead of a static size ...

Yeah, so I am afraid of breaking the existing functionality if I change the source code.

jesperpedersen:
Ok, just update the size then - to 8192 f.ex

jesperpedersen:
We have to deal with #344 so everything will change anyway

jesperpedersen:
I have increased the buffer size to 8k

GuChad369:
This updated version

```console
[pgmoneta@fedora-linux-38 ~]$ pgmoneta-cli -c pgmoneta.conf info primary newest
Label                : 20240808103224
WAL                  : 000000010000000000000072
Commments            : 
Extra                : /home/postgres/pgsql/pg_replslot/repl/state, /home/postgres/pgsql/pg_replslot/log.txt, /home/postgres/pgsql/pg_replslot/tmp/data/log.txt, /home/postgres/pgsql/pg_replslot/tmp/log.txt
Backup size          : 5251072
Restore size         : 48779264
Elapsed time         : 0
Version              : 15
Minor version        : 1
Keep                 : 0
Valid                : 1
Number of tablespaces: 0
Start LSN Hi32       : 0
Start LSN Lo32       : 72000028
End LSN Hi32         : 0
End LSN Lo32         : 72000100
Checkpoint LSN Hi32  : 0
Checkpoint LSN Lo32  : 72000060
Start timeline       : 1
End timeline         : 1
```

doc/CONFIGURATION.md
```diff
@@ -81,6 +81,7 @@ See a [sample](./etc/pgmoneta.conf) configuration for running `pgmoneta` on `loc
 | host | | String | Yes | The address of the PostgreSQL instance |
 | port | | Int | Yes | The port of the PostgreSQL instance |
 | user | | String | Yes | The replication user name |
+| extra | | String | No | The source directory for retrieval on the server side (Details are in the extra section)|
```
jesperpedersen, line 84: Details -> details

extra section -> extra parameter

doc/CONFIGURATION.md
```diff
@@ -104,6 +105,20 @@ Note, that PostgreSQL 13+ is required, as well as having `wal_level` at `replica
 
 Note, that if `host` starts with a `/` it represents a path and `pgmoneta` will connect using a Unix Domain Socket.
 
+### extra section
```
jesperpedersen, line 108: section -> parameter

doc/GETTING_STARTED.md
```diff
@@ -51,6 +51,7 @@ host = localhost
 port = 5432
 user = repl
 wal_slot = repl
+extra = /tmp/myfile, /tmp/mydir, /tmp/mydirs
```
jesperpedersen, line 54: Lets just remove this

doc/GETTING_STARTED.md
```diff
@@ -66,6 +67,8 @@ Next we create a section called `[primary]` which has the information about our
 on `localhost` on port `5432` and we will use the `repl` user account to connect, and the
 Write+Ahead slot will be named `repl` as well.
 
+If you set the `extra` parameter, when performing a backup, it will also copy all specified files on the server side to the client side. See more details in the [CONFIGURATION](https://github.com/pgmoneta/pgmoneta/blob/main/doc/CONFIGURATION.md).
```
jesperpedersen, line 70: Lets just remove this

doc/manual/02-installation.md
```diff
@@ -198,3 +198,62 @@ where `$HOME` is your home directory.
...
 
 These packages will be detected during `cmake` and built as part of the main build.
+
+## Extension installation
+
+When you configure the `extra` parameter in the server section of `pgmoneta.conf`, it requires the server side to have the [pgmoneta_ext][pgmoneta_ext] extension installed to make it work.
+
+The following instructions can help you easily install `pgmoneta_ext`. If you encounter any problems, please refer to the more detailed instructions in the [DEVELOPERS][ext_developers] documentation.
+
+### Install pgmoneta_ext
```
jesperpedersen, line 208: You can assume that the extension is deployed on yum.postgresql.org

jesperpedersen: So, split in into `dnf install` and compiling source

doc/manual/user-01-quickstart.md
```diff
@@ -49,12 +49,15 @@ host = localhost
 port = 5432
 user = repl
 wal_slot = repl
+extra = /tmp/myfile, /tmp/mydir, /tmp/mydirs
```
jesperpedersen, line 52: Remove

doc/manual/user-01-quickstart.md
```diff
@@ -49,12 +49,15 @@ host = localhost
...
 ```
 
 In our main section called `[pgmoneta]` we setup [**pgmoneta**][pgmoneta] to listen on all network addresses. We will enable Prometheus metrics on port 5001 and have the backups live in the `/home/pgmoneta` directory. All backups are being compressed with zstd and kept for 7 days. Logging will be performed at `info` level and put in a file called `/tmp/pgmoneta.log`. Last we specify the location of the `unix_socket_dir` used for management operations and the path for the PostgreSQL command line tools.
 
 Next we create a section called `[primary]` which has the information about our [PostgreSQL][postgresql] instance. In this case it is running on `localhost` on port `5432` and we will use the `repl` user account to connect, and the Write+Ahead slot will be named `repl` as well.
 
+If you set the `extra` parameter, when performing a backup, it will also copy all specified files on the server side to the client side. See more details in the [CONFIGURATION][configuration].
```
jesperpedersen, line 59: Remove

doc/manual/user-02-configuration.md
```diff
@@ -81,6 +81,7 @@ See a [sample][sample] configuration for running [**pgmoneta**][pgmoneta] on `lo
 | host | | String | Yes | The address of the PostgreSQL instance |
 | port | | Int | Yes | The port of the PostgreSQL instance |
 | user | | String | Yes | The replication user name |
+| extra | | String | No | The source directory for retrieval on the server side |
```
jesperpedersen, line 84: Same comments as the non-manual files

src/include/pgmoneta.h
```diff
@@ -234,6 +236,8 @@ struct server
    int backup_max_rate;     /**< Number of tokens added to the bucket with each replenishment for backup. */
    int network_max_rate;    /**< Number of bytes of tokens added every one second to limit the netowrk backup rate */
    int manifest;  /**< The manifest hash algorithm */
+   int number_of_extra; /**< The number of source directory*/
+   char extra[MAX_EXTRA][MAX_PATH];  /**< Source directory*/
```
jesperpedersen, line 240: it should be `[MAX_EXTRA][MAX_EXTRA_PATH];`

src/libpgmoneta/configuration.c
```diff
@@ -82,6 +82,8 @@ static int restart_string(char* name, char* e, char* n);
 static bool is_empty_string(char* s);
 static int remove_leading_whitespace_and_comments(char* s, char** trimmed_line);
 
+static void split_extra(const char* extra, char res[MAX_EXTRA][MAX_PATH], int* count);
```
jesperpedersen, line 85: MAX_EXTRA_PATH

src/libpgmoneta/configuration.c
```diff
@@ -3258,3 +3284,25 @@ remove_leading_whitespace_and_comments(char* s, char** trimmed_line)
    *trimmed_line = NULL;
    return 1;
 }
+
+static void
+split_extra(const char* extra, char res[MAX_EXTRA][MAX_PATH], int* count)
```
jesperpedersen, line 3289: MAX_EXTRA_PATH

jesperpedersen (changes requested):

GuChad369:
For the `doc/manual/02-installation.md`, I am unsure if the `pgmoneta_ext` installation will work as described.

doc/manual/02-installation.md
```diff
@@ -198,3 +198,56 @@ where `$HOME` is your home directory.
...
+
+After you have successfully installed `pgmoneta`, the following commands will help you install `pgmoneta_ext`:
+
+```sh
+dnf install -y pgmoneta_ext
+```
+
```
jesperpedersen, line 215: You need the postgrersql.conf `shared_preload_libraries = 'pgmoneta_ext' part as well.

And, describe that it requires a restart of PostgreSQL

jesperpedersen (changes requested):

doc/manual/02-installation.md
GuChad369: So, it should be something like this:

```sh
dnf install -y pgmoneta_ext
```

You need to add the `pgmoneta_ext` library for PostgreSQL in `postgrersql.conf` as well:

```ini
shared_preload_libraries = 'pgmoneta_ext'
```

And remember to restart PostgreSQL to make it work.

jesperpedersen: yeah

GuChad369: ok, thanks

jesperpedersen:
You can validate if the extension is installed in `pgmoneta_validate_configuration(void* shm)` if `strlen(extra) > 0` and give a WARN if it isn't

GuChad369:
We already have a function `pgmoneta_ext_is_installed` to check if the extension is installed.

jesperpedersen:
Hmm, maybe it is enough during the backup since the extension could have been uninstalled...

Do an INFO if `strlen(extra) > 0` that the extension is required

GuChad369:
Is there a new commit related to the `info` functionality?

jesperpedersen:
What do you mean ?

https://github.com/pgmoneta/pgmoneta/commit/144f8f013c759ba211a35eb8f596fe2318c49d8e went in

jesperpedersen:
@Jubilee101 Looks like `key` isn't assigned...

GuChad369:
Add a new check to ensure the `pgmoneta_ext` extension is installed on the server side when the `extra` parameter is set.

jesperpedersen:
Merged.

Issue #350, opened 2024-07-31
Universal typed struct


Jubilee101:
We've been bugged for a long time that we don't have a type system to manage payload life cycles for data structures such as json, deque or ART. This seems like a good time to implement one.

PR #354, opened 2024-08-08
[#350] Universal value types

Jubilee101:
Add a universal typed `value` struct that can be shared across `ART`, `deque` and `json`. It has two callbacks - `to_string` for printing and `destroy_data` for memory reclaim. With all types (almost) equipped with to_string() method, logging is more clearer than ever before. The system is also easy to extend to include other types.

Redesign `deque` and `ART` on basis of `value` struct, which are now able to hold ANY type of values without the memory leak nightmare.

Redesign `json` on basis of `deque` and `json`, now able to put and look up a kv pair in O(k) time where k is the length of the key, and append an array entry in O(1) time, and, you guessed it, no worries about memory leak :)

Add iterator for `json` and `deque`.

### DIFF
--- a/src/include/art.h
+++ b/src/include/art.h
@@ -34,12 +34,13 @@ extern "C" {
 #endif
 
 #include <deque.h>
+#include <value.h>
 
 #include <stdint.h>
 
 #define MAX_PREFIX_LEN 10
 
-typedef int (*art_callback)(void* data, const unsigned char* key, uint32_t key_len, void* value);
+typedef int (*art_callback)(void* data, const unsigned char* key, uint32_t key_len, struct value* value);
 
 typedef void (*value_destroy_callback)(void* value);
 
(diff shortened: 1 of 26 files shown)


src/include/art.h
```diff
@@ -85,38 +82,46 @@ pgmoneta_art_destroy(struct art* tree);
...
  * Deletes a value from the ART tree
  * @param t The tree
  * @param key The key
  * @param key_len The length of the key
- * @return NULL if the item was not found, otherwise
- * the value pointer is returned
+ * @return 0 if success or value not found, 1 if otherwise
  */
-void*
+int
```
jesperpedersen, line 102: shouldn't value not found give a 1 ?

src/include/deque.h
```diff
@@ -102,9 +102,9 @@ pgmoneta_deque_add(struct deque* deque, char* tag, void* data);
  * This function is thread safe, but the returned value is not protected
  * @param deque The deque
  * @param tag [out] Optional, tag will be returned through if not NULL
- * @return The value if deque's not empty, otherwise NULL
+ * @return The value data if deque's not empty, otherwise 0
  */
-void*
+uintptr_t
```
jesperpedersen, line 107: and if the data value is 0 ?

jesperpedersen (changes requested):

Issue #147, opened 2023-08-25
Incremental Backup


Jubilee101:
We would like to add the incremental backup option for pgmoneta. Our eventual hope is to save time and space when doing backup by only receiving changed files and replacing the rest with symlinks. See more details at the discussion forum #146.

jesperpedersen:
This is a potential big task.

We know the LSN of the last backup, so we only need to get the changes since then. How, and how to apply is another matter.

Delta backups are important in order to support bigger installations, so this is a 0.8 task for sure.

Jubilee101:
Yes this is a big task. I'll take my time. I'll start with something easy and let's discuss in the forum about design choices and details.

Jubilee101:
As a starter, we'll work on support for the official delta/incr in postgres 17

PR #361, opened 2024-08-22
[#147] Send manifest file

GuChad369:
This is a basic idea for achieving the function of sending our manifest file to the server side. 

I prefer using a table to send this file, as it will be easier for the extension to read.

### DIFF
--- a/src/include/extension.h
+++ b/src/include/extension.h
@@ -113,6 +113,27 @@ pgmoneta_ext_get_file(SSL* ssl, int socket, const char* file_path, struct query_
 int
 pgmoneta_ext_get_files(SSL* ssl, int socket, const char* file_path, struct query_response** qr);
 
+/**
+ * Creates a table and inserts data from a manifest file
+ * @param ssl The SSL structure used for secure communication
+ * @param socket The socket file descriptor for the connection
+ * @param file_path The path to the manifest file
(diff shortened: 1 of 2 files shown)


GuChad369:
@Jubilee101

I encountered a problem with this function. The `CREATE` command works on the server side, but the return result fails. I think the issue might be with the `pgmoneta_query_execute` function, which may not be handling the `CREATE` command result properly, leading to a null return.

I will continue working on this, but I want to make sure:  Should I modify the `pgmoneta_query_execute` function, or should I create a new function to avoid breaking existing code?

Jubilee101:
Yeah it is a easy solution. But I'm not sure it is ideal, setting aside the fact that users may not be happy that we write data into their database, doing the insert may mess up with the later backup as well as WAL. It's basically just sending a file, and there are number of ways to do it, scp or ftp for example, and we already have ssh support, poke @jesperpedersen for details. Let's not rush it too much. Think about all the plans and do the pros and cons. And take a look at pgbackrest in case they have better ideas. 

Let's take the discussion to the issue or the discussion forum. We are closing this PR.

GuChad369:
Creating a new table and inserting data might not be the best approach, as it could complicate future backups. You're right about that. Instead, I'll focus on understanding how pgBackRest handles this process first.

Issue #269, opened 2024-04-13
Direct IO


jesperpedersen:
Investigate the impact of using `O_DIRECT` for disk I/O - especially in restore scenarios

jesperpedersen:
We need a disk benchmark to test the current setup vs using `O_DIRECT`. Latter should be faster, but we need to know that the extra complexity is worth it

jesperpedersen:
This task is also about having a uniform disk access method - so maybe an `io.h` that operates on `FILE*`

jesperpedersen:
https://commitfest.postgresql.org/46/4532/ could be a good read as well...

palak-chaturvedi:
Hey, @jesperpedersen 
I did some research on DIRECT_IO.

1. We replace the fopen tag with open() - that gets attribute O_DIRECT and create a file descriptor - fd (int). `int open(const char *pathname, int flags, mode_t mode);`
2. Create a fileFlag variable and fileMode variable for permissions.
3. The O_DIRECT flag can be added to fileFlag according to needs.
4. Open the fd to return a pointer using `FILE* fp = fdopen(fd, "w");` 
5. Change file open method in all places.
6. And also add a PG_MONETA_DIRECT_IO flag in configuration if that flag is true we will call the O_DIRECT method otherwise we won't.
https://www.codequoi.com/en/handling-a-file-by-its-descriptor-in-c/

And do we also need to check that the shared memory is aligned to the disk in pgmoneta also?
This is based on the work that is done in postgres for direct io.

palak-chaturvedi:
This is the memory usage that I have found normally 

start
```
Every 2.0s: free -m    fedora: Wed Apr 24 11:47:02 2024

               total        used        free      shared  buff/cache   available
Mem:            5769         773        4241         1           985        4996
Swap:           5768           0        5768
```
after init db
```
Every 2.0s: free -m    fedora: Wed Apr 24 11:48:45 2024

               total        used        free      shared  buff/cache   available
Mem:            5769         918        3528          40        1593        4850
Swap:           5768           0        5768
```

Create User command
```
Every 2.0s: free -m    fedora: Wed Apr 24 11:54:04 2024

               total        used        free      shared  buff/cache   available
Mem:            5769         955        3483          68        1628        4813
Swap:           5768           0        5768
```

Pgmoneta start
```
Every 2.0s: free -m    fedora: Wed Apr 24 11:56:16 2024

               total        used        free      shared  buff/cache   available
[...]

PR #388, opened 2024-10-04
[#269] Direct IO

shikharish:
@jesperpedersen

### DIFF
--- a/src/include/io.h
+++ b/src/include/io.h
@@ -46,6 +46,15 @@ extern "C" {
 #endif
 
 /**
+ * Returns an aligned buffer with same data as unaligned buffer
+ * @param buffer The unaligned buffer
+ * @param size The buffer size
+ * @return The aligned buffer
+ */
+void*
+pgmoneta_unaligned_to_aligned_buffer(void* buffer, size_t size);
+
+/**
  * Wrapper around write() libc call
  * @param fd The file descriptor
  * @param buffer The buffer to be written
--- a/src/include/io.h
+++ b/src/include/io.h
@@ -1,0 +1,69 @@
+/*
+ * Copyright (C) 2024 The pgmoneta community
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ * this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its contributors
+ * may be used to endorse or promote products derived from this software without
+ * specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef PGMONETA_IO_H
+#define PGMONETA_IO_H
+
+#ifdef __cpluscplus
+extern "C" {
+#endif
+
+#include <fcntl.h>
+#include <stdio.h>
+#include <sys/types.h>
+
+#define BLOCK_SIZE 512
+
+#if !defined (HAVE_LINUX) && !defined (HAVE_FREEBSD)
+#define O_DIRECT 0
+#endif
+
+/**
+ * Wrapper around write() libc call
+ * @param fd The file descriptor
+ * @param buffer The buffer to be written
+ * @param bytes The number of bytes to be written
+ * @return The number of bytes written, -1 on error
+ */
+ssize_t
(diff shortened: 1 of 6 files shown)


jesperpedersen:
You need to investigate why `pgmoneta-cli backup primary` takes so long - its 2 sec vs 17 sec

shikharish:
I don't see where it took 17sec. On my system the results with and without the patch are similar.

2 backups With O_DIRECT:
```
vscode ➜ /workspaces/pgmoneta/build/src (direct-io) $ ./pgmoneta-cli -c pgmoneta.conf backup primary     
Header: 
  ClientVersion: 0.15.0
  Command: 1
  Output: 0
  Timestamp: 20241006193213
Outcome: 
  Status: true
  Time: 00:00:03
Request: 
  Server: primary
Response: 
  Backup: 20241006193213
  BackupSize: 6258688
  Compression: 2
  Encryption: 0
  MajorVersion: 17
  MinorVersion: 0
  RestoreSize: 40222720
  Server: primary
  ServerVersion: 0.15.0
vscode ➜ /workspaces/pgmoneta/build/src (direct-io) $ ./pgmoneta-cli -c pgmoneta.conf backup primary
Header: 
  ClientVersion: 0.15.0
  Command: 1
  Output: 0
  Timestamp: 20241006193218
Outcome: 
  Status: true
  Time: 00:00:02
Request: 
  Server: primary
Response: 
  Backup: 20241006193218
  BackupSize: 3969024
  Compression: 2
  Encryption: 0
  MajorVersion: 17
  MinorVersion: 0
  RestoreSize: 40222720
  Server: primary
  ServerVersion: 0.15.0
```

2 backups without O_DIRECT (main branch):
```
vscode ➜ /workspaces/pgmoneta/build/src (main) $ ./pgmoneta-cli -c pgmoneta.conf backup primary
Header: 
  ClientVersion: 0.15.0
  Command: 1
  Output: 0
  Timestamp: 20241006193347
Outcome: 
  Status: true
  Time: 00:00:01
Request: 
  Server: primary
Response: 
  Backup: 20241006193347
  BackupSize: 16146432
  Compression: 2
  Encryption: 0
  MajorVersion: 16
  MinorVersion: 3
  RestoreSize: 220741632
  Server: primary
  ServerVersion: 0.15.0
vscode ➜ /workspaces/pgmoneta/build/src (main) $ ./pgmoneta-cli -c pgmoneta.conf backup primary
Header: 
  ClientVersion: 0.15.0
  Command: 1
  Output: 0
  Timestamp: 20241006193350
Outcome: 
  Status: true
  Time: 00:00:02
Request: 
  Server: primary
Response: 
  Backup: 20241006193350
  BackupSize: 5312512
  Compression: 2
  Encryption: 0
  MajorVersion: 16
  MinorVersion: 3
  RestoreSize: 220741632
  Server: primary
  ServerVersion: 0.15.0
```

shikharish:
updated the comment  - was mistakenly using different postgresql version on different branches

Jubilee101:
Tried it as well, yeah, the result was similar to what @shikharish had.

jesperpedersen:
So, `O_DIRECT` is slower...

jesperpedersen:
We need another design... we can't have all these `fcntl` / `malloc` calls per `pgmoneta_write_file` .

There can only be 1 call per `fd`...

jesperpedersen:
`pgmoneta_write_file()` should basically be just the `write()` call.

The memory buffer for the call should be a memory aligned buffer based on the disk buffer size.

If `O_DIRECT` isn't supported then it should be transparent to the `pgmoneta_write_file()`  call -- if we even need that function

shikharish:
@jesperpedersen The O_SYNC flag is the reason for the comparatively slower speed. Using only O_DIRECT gives faster results:

```
vscode ➜ /workspaces/pgmoneta/build/src (direct-io) $ ./pgmoneta-cli -c pgmoneta.conf backup primary           
Header: 
  ClientVersion: 0.15.0
  Command: 1
  Output: 0
  Timestamp: 20241008065106
Outcome: 
  Status: true
  Time: 00:00:01
Request: 
  Server: primary
Response: 
  Backup: 20241008065106
  BackupSize: 6258688
  Compression: 2
  Encryption: 0
  MajorVersion: 17
  MinorVersion: 0
  RestoreSize: 40222720
  Server: primary
  ServerVersion: 0.15.0
vscode ➜ /workspaces/pgmoneta/build/src (direct-io) $ ./pgmoneta-cli -c pgmoneta.conf backup primary
Header: 
  ClientVersion: 0.15.0
  Command: 1
  Output: 0
  Timestamp: 20241008065109
Outcome: 
  Status: true
  Time: 00:00:01
Request: 
  Server: primary
Response: 
  Backup: 20241008065109
  BackupSize: 3969024
  Compression: 2
  Encryption: 0
  MajorVersion: 17
  MinorVersion: 0
  RestoreSize: 40222720
  Server: primary
  ServerVersion: 0.15.0
```

Is it necessary to use the O_SYNC flag? I am sure the speed for buffered-io will also decrease if we use the O_SYNC flag.

jesperpedersen:
We can flush the file before closing it, but we also need to change `pgmoneta_write_file()` such that we don't have all these system calls, then it will be faster

jesperpedersen:
You can try `posix_fadvise()` with `POSIX_FADV_SEQUENTIAL` once the file is opened if `#ifdef HAVE_LINUX`

shikharish:
If we make `pgmoneta_write_function()` completely transparent to direct and buffered writes, we would have to do the same work outside of the function for every write operation. For ex., any odd-byte sized write operation *has* to be done with O_DIRECT disabled, so we will have to add this check. 

I have made some updates to the PR:
- `pgmoneta_write_function()` will only check for misalignment - if zero, proceeds to write with O_DIRECT - if non-zero, disables O_DIRECT, writes file, renable O_DIRECT
- remove O_SYNC
- add `fsync()` calls

Tested backup with this patch on a larger database - 1:03sec vs  1:09sec without patch(main branch)

shikharish:
To give an idea, out of 11509 `write()` calls,
- 6399 were direct-io (from `wal.c`) 
  - out of which 259 were with misalignment(so new buffer was malloced and data copied for each call)
- 5110 were buffered-io (from `message.c`)
  - almost all were with misalignment which was not fixed

jesperpedersen:
@shikharish Still doing ok ?

jesperpedersen:
@shikharish Still want to work on this ?

shikharish:
Yes, give me a few more days please.

jesperpedersen:
Don't worry about it, it is a matter of keeping the pull request rebased so it can be applied easily, and communicate about what issues you are having, or things you are working on

shikharish:
The final solution I think is to change the way we parse any `message` we get from the database. Currently, we store all the data in `message->data` and that is the issue. I am working on changing the api in message.c so that `message->data` only contains the actual data of the message and `message->length` contains only the length of the actual message. And when we need to write the message, we can re-construct it to follow pgprotocol.

shikharish:
It is a bit complicated to cover all cases of the message types such that any write will follow the alignment rules of O_DIRECT. For ex., when writing wal data('w') to file we skip the first 25 bytes so will have to handle that case differently.

jesperpedersen:
@shikharish I think this is way too complex for what we will gain kernel space usage, and disk I/O. Furthermore, the whole message infrastructure will be complicated and harder to maintain.

I think for now we should just shelve the project, and only reopen if we have a very good and simple idea that will work. We have a lot of tasks to work on which I would rather see completed.

I think it is great that you have investigated this area so we now know what pitfalls there are.

Thanks !

shikharish:
@jesperpedersen Alright. Working on this was really interesting. I will keep this in the back of my mind for now. And work on other tasks. Thanks.

Issue #376, opened 2024-09-20
Management: Support compression


jesperpedersen:
Support NONE, and all the different compression methods.

The CLI can use `-C 0|none...` and the same method will be used for the return payload

GuChad369:
I remember I implemented `pgmoneta-cli compress file_path`. So for this issue, the desired behavior is to make it work when the user calls `pgmoneta-cli -C zstd file_path` to perform the same task. If the user calls `pgmoneta-cli -C none file_path`, it will use the default method for compression. Is my understanding correct?

jesperpedersen:
No, this is for the protocol, so

```
pgmoneta-cli -C zstd status
```

will give the output - but on the wire it has been compressed

GuChad369:
Apologies, I'm not clear on this. Could you explain in more detail or provide an example?

jesperpedersen:
You need to compress the JSON document, so add a `-C zstd` option -- all of them -- then compress the JSON document, and send it as the string.

Of course on the other side you need to uncompress to get the JSON document back

GuChad369:
What I need to do is modify the function `pgmoneta_management_write_json` and also add the option in `cli.c`.

```c
int
pgmoneta_management_write_json(SSL* ssl, int socket, struct json* json)
{
   ... 

   s = pgmoneta_json_to_string(json, FORMAT_JSON, NULL, 0);

   ...

   if(-C){
      compress_buffer = compress_s_output(...)
      base64 = encode(compress_buffer)
   }

   if (write_string("pgmoneta-cli", ssl, socket, base64))
   {
      goto error;
   }

   ...
}
```

is this correct?

Jubilee101:
> Apologies, I'm not clear on this. Could you explain in more detail or provide an example?

Sorry we are not so clear about the context. Since #344 we've upgraded our protocol between cli and management. Recall how we used to send parameters and options from cli input to management on main, it's one parameter at a time. Now they are all wrapped inside a json as the payload, which will be converted to string and sent to main. And on the other side the json string gets parsed and the parameters and command are extracted.

This has clear benefits as it simplified the management and made it easier to extend. Also if you try commands now you'll see they all support json output.

But clearly it brings bandwidth overhead of sending the json string. That's where you come in to help, by compressing the string before sending it, and decompressing after receiving it. I think you are already getting the gist of it.

GuChad369:
[...]

PR #391, opened 2024-10-07
[#376] Management: Support compression

GuChad369:
- Expose a new variable, `compress_method,` to support compression and reduce bandwidth stress when sending the JSON string.
- Implement four types of compression method functions to convert a string to a buffer and convert a buffer back to a string.


src/include/achv.h
```diff
@@ -43,9 +43,10 @@ extern "C" {
  * @param client_fd The client
  * @param server The server
  * @param request The request
+ * @param compress_method The compress method for json format
  */
 void
-pgmoneta_archive(SSL* ssl, int client_fd, int server, struct json* request);
+pgmoneta_archive(SSL* ssl, int client_fd, int server, struct json* request, uint8_t compress_method);
```
jesperpedersen, line 49: Do `..., uint8_t compression, struct json* request`

src/include/management.h
```diff
@@ -46,6 +46,10 @@ extern "C" {
  * Management header
  */
 #define MANAGEMENT_COMPRESSION_NONE 0
+#define MANAGEMENT_COMPRESSION_GZIP 1
+#define MANAGEMENT_COMPRESSION_ZSTD 2
+#define MANAGEMENT_COMPRESSION_LZ4 3
+#define MANAGEMENT_COMPRESSION_BZIP2 4
```
jesperpedersen, line 52: Align the numbers

src/include/management.h
```diff
@@ -278,21 +282,23 @@ extern "C" {
  * @param socket The socket descriptor
  * @param server The server
  * @param output_format The output format
+ * @param compress_method The compress method for json format
  * @return 0 upon success, otherwise 1
  */
 int
-pgmoneta_management_request_backup(SSL* ssl, int socket, char* server, int32_t output_format);
+pgmoneta_management_request_backup(SSL* ssl, int socket, char* server, int32_t output_format, uint8_t compress_method);
```
jesperpedersen, line 289: Have it before output_format, so `uinbt8_t compression, int32_t output_format`

src/include/utils.h
```diff
@@ -1182,6 +1182,24 @@ pgmoneta_atoi(const char* input);
 char*
 pgmoneta_indent(char* str, char* tag, int indent);
 
+/**
```
jesperpedersen, line 1185:
```
src/include/utils.h:pgmoneta_base64_encode(char* raw, size_t raw_length, char** encoded, size_t* encoded_length);
src/include/utils.h:pgmoneta_base64_decode(char* encoded, size_t encoded_length, char** raw, size_t* raw_length);
```

?

jesperpedersen (changes requested):

jesperpedersen:
Minor comments

src/cli.c
```diff
@@ -431,6 +434,33 @@ main(int argc, char** argv)
...
+            }
+            else if (!strncmp(optarg, "bz2", MISC_LENGTH))
+            {
+               compress_method = MANAGEMENT_COMPRESSION_BZIP2;
+            }
+            else if (!strncmp(optarg, "none", MISC_LENGTH))
+            {
+               compress_method = MANAGEMENT_COMPRESSION_NONE;
+            }
+            else
```
jesperpedersen, line 458: ?

src/cli.c
```diff
@@ -614,106 +638,126 @@ main(int argc, char** argv)
       }
    }
 
+   if (compress_method == MANAGEMENT_COMPRESSION_NONE)
```
jesperpedersen, line 641: Why ?

src/cli.c
```diff
@@ -95,26 +95,26 @@ static void help_info(void);
...
-static int reload(SSL* ssl, int socket, int32_t output_format);
-static int retain(SSL* ssl, int socket, char* server, char* backup_id, int32_t output_format);
-static int expunge(SSL* ssl, int socket, char* server, char* backup_id, int32_t output_format);
-static int decrypt_data(SSL* ssl, int socket, char* path, int32_t output_format);
-static int encrypt_data(SSL* ssl, int socket, char* path, int32_t output_format);
-static int decompress_data(SSL* ssl, int socket, char* path, int32_t output_format);
-static int compress_data(SSL* ssl, int socket, char* path, int32_t output_format);
-static int info(SSL* ssl, int socket, char* server, char* backup, int32_t output_format);
-static int annotate(SSL* ssl, int socket, char* server, char* backup, char* command, char* key, char* comment, int32_t output_format);
+static int backup(SSL* ssl, int socket, char* server, int32_t output_format, uint8_t compress_method);
```
jesperpedersen, line 98: `output_format` is last

jesperpedersen (changes requested):

src/include/utils.h
GuChad369: what is your meaning here?

jesperpedersen: We have BASE64 functions

src/cli.c
GuChad369: do we leave it blank?

src/cli.c
GuChad369: We handle the default compression method at this point. If none is specified, we use the default method to compress the data. By managing it here, we avoid needing to retrieve the default method later. Since the `config` variable is already available in `main()`, it’s more convenient to get the default method here.

jesperpedersen: `none` is default - so no `-C` means no compression

GuChad369: If "none" means no compression, why do we provide this option?

GuChad369: we can limit user can only input
`-C gz`
`-C gz`
`-C gz`

jesperpedersen: Yes, and `-C none` because it is the default.

If people want to use a protocol sniffer, then they shouldn't have to deal with compression and encryption on the wire

jesperpedersen:
I don't understand - no `-C none` means no compression, as well as `-C none` does, and it is the default.

If there is no `-C` then there is no compression, or if there is `-C none` is means there is no compression

GuChad369:
Alright, I’ll update the logic.

jesperpedersen:
Rebase

GuChad369:
Since all of my compression functions convert the string into an `unsigned char*`, which represents the binary data, our function `pgmoneta_base64_encode()` and `pgmoneta_base64_decode()` requires a `char*`. We have to convert the `unsigned char*` into a `char*`, which may lead to a failure in properly encoding and decoding the original binary buffer.

jesperpedersen:
@GuChad369 You need to look at the function signatures - a quick look shows that `BIO_write()` takes a `void*` instead of `char*` so likely the function signatures are wrong, and everything need to be changed.

So, in the end we will have a `void*` so you will need to write that in management.c

GuChad369:
Should I modify the current functions, or would it be better to create a new one?

jesperpedersen:
Modify the current ones

jesperpedersen:
We should only have one function across the project for each "area"

GuChad369:
I plan to update the functions to the following signatures, along with some minor code adjustments. This change will affect all areas that currently use these two functions:

```c
int pgmoneta_base64_encode(void* raw, size_t raw_length, char** encoded, size_t* encoded_length);
int pgmoneta_base64_decode(char* encoded, size_t encoded_length, void** raw, size_t* raw_length);
```

I'm concerned that this change might impact the functionality of our project.

jesperpedersen:
It looks correct - of course it will require changes for their usage, but it shouldn't be a problem.

You can do the BASE64 in a separate pull request if you want

jesperpedersen:
The important part is to look at the underlying APIs - in this case from OpenSSL - and align usage against that

GuChad369:
This update resolves the previous issue. Please review it, and if you find any problems, let me know.

jesperpedersen:
Yes, you are just missing the documentation -- explain `-C` in full.

You can use `git grep list-backup` to find all the files where it maybe should be explained

src/cli.c
```diff
@@ -162,7 +162,8 @@ usage(void)
    printf("  -v, --verbose                Output text string of result\n");
    printf("  -V, --version                Display version information\n");
    printf("  -F, --format text|json|raw   Set the output format\n");
-   printf("  -?, --help                   Display help\n");
+   printf("  -C, --compress COMPRESS_METHOD|none   Compress the json\n");
```
jesperpedersen, line 165: Here you need `none|gz|...`

doc/CLI.md
```diff
@@ -23,6 +23,7 @@ Options:
   -v, --verbose                 Output text string of result
   -V, --version                 Display version information
   -F, --format text|json|raw    Set the output format
+  -C, --compress none|gz|...    Compress the json format
```
jesperpedersen, line 26: I meant that you need to list all of them

doc/man/pgmoneta-cli.1.rst
```diff
@@ -45,6 +45,12 @@ OPTIONS
 -V, --version
   Display version information
 
+-F, --format text|json|raw    
+  Set the output format
+
+-C, --compress none|gz|...
```
jesperpedersen, line 51: Same here

doc/manual/user-01-quickstart.md
```diff
@@ -12,13 +12,17 @@ Usage:
...
+  -c, --config CONFIG_FILE    Set the path to the pgmoneta.conf file
+  -h, --host HOST             Set the host name
+  -p, --port PORT             Set the port number
+  -U, --user USERNAME         Set the user name
+  -P, --password PASSWORD     Set the password
+  -L, --logfile FILE          Set the log file
+  -v, --verbose               Output text string of result
+  -V, --version               Display version information
+  -F, --format text|json|raw  Set the output format
+  -C, --compress none|gz|...  Compress the json format
```
jesperpedersen, line 24: Again

doc/manual/user-01-quickstart.md
```diff
@@ -119,15 +123,17 @@ Usage:
...
```
jesperpedersen, line 135: Same

doc/manual/user-01-quickstart.md
```diff
@@ -191,12 +197,12 @@ Usage:
   pgmoneta-admin [ -f FILE ] [ COMMAND ]
 
 Options:
-  -f, --file FILE         Set the path to a user file
-  -U, --user USER         Set the user name
-  -P, --password PASSWORD Set the password for the user
-  -g, --generate          Generate a password
-  -l, --length            Password length
-  -V, --version           Display version information
+  -f, --file FILE             Set the path to a user file
```
jesperpedersen, line 200: Indent changes

src/cli.c
```diff
@@ -162,7 +162,8 @@ usage(void)
    printf("  -v, --verbose                Output text string of result\n");
    printf("  -V, --version                Display version information\n");
    printf("  -F, --format text|json|raw   Set the output format\n");
-   printf("  -?, --help                   Display help\n");
+   printf("  -C, --compress none|gz|...   Compress the json format\n");
```
jesperpedersen, line 165: Same

src/cli.c
```diff
@@ -462,12 +492,6 @@ main(int argc, char** argv)
       exit(1);
    }
 
-   if (configuration_path != NULL && (host != NULL || port != NULL))
```
jesperpedersen, line 465: Why ?

jesperpedersen (changes requested):

src/cli.c
GuChad369: accident changes

GuChad369:
updated

doc/CLI.md
```diff
@@ -14,16 +14,17 @@ Usage:
...
+  -c, --config CONFIG_FILE              Set the path to the pgmoneta.conf file
+  -h, --host HOST                       Set the host name
+  -p, --port PORT                       Set the port number
+  -U, --user USERNAME                   Set the user name
+  -P, --password PASSWORD               Set the password
+  -L, --logfile FILE                    Set the log file
+  -v, --verbose                         Output text string of result
+  -V, --version                         Display version information
+  -F, --format text|json|raw            Set the output format
+  -C, --compress none|gz|zstd|lz4|bz2   Compress the json format
```
jesperpedersen, line 26: "Compress the wire protocol"

jesperpedersen: Likewise for the rest

src/cli.c
```diff
@@ -447,6 +450,33 @@ main(int argc, char** argv)
...
+            {
+               compression = MANAGEMENT_COMPRESSION_BZIP2;
+            }
+            else if (!strncmp(optarg, "none", MISC_LENGTH))
+            {
+               break;
+            }
+            else
+            {
+               warnx("pgmoneta-cli: Use either -c or -h/-p to define endpoint");
```
jesperpedersen, line 476: This is wrong

jesperpedersen (changes requested):

src/cli.c
GuChad369: There have been some changes to the `-F` option, which caused a conflict with my commit, leading to unintended changes.

jesperpedersen:
```
Direct leak of 355 byte(s) in 1 object(s) allocated from:
    #1 0x7ffbfe376ecd in pgmoneta_base64_decode src/libpgmoneta/utils.c:940:14
    #2 0x7ffbfe303370 in pgmoneta_management_read_json src/libpgmoneta/management.c:891:11
    #3 0x555d32e55576 in process_result src/cli.c:1441:8

Direct leak of 127 byte(s) in 1 object(s) allocated from:
    #1 0x7ffbfe2db1db in pgmoneta_gzip_string src/libpgmoneta/gzip_compression.c:633:35
    #2 0x7ffbfe2feae7 in pgmoneta_management_write_json src/libpgmoneta/management.c:999:17
    #3 0x7ffbfe2fff04 in pgmoneta_management_request_status src/libpgmoneta/management.c:322:8
```

Check for leaks...

jesperpedersen:
```
    #1 0x7f6c47576ecd in pgmoneta_base64_decode src/libpgmoneta/utils.c:940:14
    #2 0x7f6c47503370 in pgmoneta_management_read_json src/libpgmoneta/management.c:891:11
    #3 0x558d8fde7033 in accept_mgt_cb src/main.c:827:8
```

jesperpedersen:
Rebase

GuChad369:
This update no longer has memory leaks related to the newly added functions. While there might still be some memory leaks, they are not connected to any of the new functions.

jesperpedersen:
Merged.

Issue #392, opened 2024-10-11
Correct base64 signature


GuChad369:
Adjust the base64 encode and decode function signatures to make them more compatible with various data types, rather than being limited to the `char` type.

PR #393, opened 2024-10-11
[#392] Correct the base64 functions signature

GuChad369:
Update the base64 encode and decode function signatures to make them more compatible with different data types.


GuChad369:
I'm not sure why the check failed. The error might not be related to the changes I made.

jesperpedersen:
Merged.

Issue #397, opened 2024-10-13
CLI: Translation of retention


jesperpedersen:
Add translation of retention fields

```
      RetentionDays: 7
      RetentionMonths: -1
      RetentionWeeks: -1
      RetentionYears: -1
```

->

```
      RetentionDays: 7
      RetentionMonths: Unspecified
      RetentionWeeks: Unspecified
      RetentionYears: Unspecified
```

Test with `pgmoneta-cli status details`

jesperpedersen:
@ashu3103 Maybe you want this one ?

ashu3103:
Sure I’ll take that

jesperpedersen:
@ashu3103 Thanks !

PR #398, opened 2024-10-13
[#397] Translation of retention fields

ashu3103:


src/cli.c
```diff
@@ -1733,6 +1736,16 @@ translate_response_argument(struct json* response)
    free(translated_used_space);
 }
 
+static void
+translate_server_retention_argument(struct json* response, char* tag)
+{
+   if ((int32_t)pgmoneta_json_get(response, tag) <= 0)
```
jesperpedersen, line 1742: 0 is a valid value

jesperpedersen (changes requested):

src/cli.c
ashu3103: Done

jesperpedersen:
Merged.

Issue #373, opened 2024-09-19
Support escape characters in JSON


Jubilee101:
Support escape character in [pgmoneta_json_parse_string](https://github.com/pgmoneta/pgmoneta/blob/c3c903b7996c59856817f63117c42e7f2b1dd061/src/include/json.h#L216) and [pgmoneta_json_to_string](https://github.com/pgmoneta/pgmoneta/blob/c3c903b7996c59856817f63117c42e7f2b1dd061/src/include/json.h#L236). I think what we need is mostly `"\""` and `"\'"` but it would be nice if we support the rest, if not too hard.

Note that for to_string this should only apply to `FORMAT_JSON`, for `FORMAT_TEXT` there's no need since strings are not wrapped in double quotation marks. Checkout [string_to_string_cb](https://github.com/pgmoneta/pgmoneta/blob/c3c903b7996c59856817f63117c42e7f2b1dd061/src/libpgmoneta/value.c#L411) for details.

Jubilee101:
Hey @ashu3103, let me know if you are interested in this.

ashu3103:
> Hey @ashu3103, let me know if you are interested in this.

Sure

ashu3103:
@Jubilee101, I have started working on this,

Just one thing to confirm, we only need escape characters support for the values of the json fields right?
Or does it has to be implemented for the keys

Thanks

jesperpedersen:
Did you check the spec - https://www.json.org/json-en.html ?

ashu3103:
> Did you check the spec - https://www.json.org/json-en.html ?

Very helpful, Thanks for sharing

jesperpedersen:
@Jubilee101 Can you take this one

jesperpedersen:
@ashu3103 Submit your PR, and Haoran will work with you on it

Jubilee101:
> Just one thing to confirm, we only need escape characters support for the values of the json fields right?
Or does it has to be implemented for the keys as well

Hmm, I would prefer support for both key and value, unless we have problems with supporting keys, in that case let us know and we can discuss what's best.

It's mostly for handling escape characters when parsing a json string, you need to be able to identify them and parse each of them into corresponding character. In terms of printing the most important thing is to treat `"` specially -- print it out as `\"` to avoid ambiguity, and as for the rest just use your judgement to decide what to do. Note that it's just for printing, when a value is stored in a string and not printed out, it should always be one char per escape character.

[...]

PR #404, opened 2024-10-16
[#373] support escape characters in json

ashu3103:
# WIP: Support escape characters

Escape characters support for both keys and values of a json object for escape characters :-
- `'`
- `"`

for now (once logic is on track will add more)

Since it is used only for printing purpose we are parsing the string with escape characters and store them as characters in the value of `json` struct. Then at the time of conversion from string to json, the escape character are added and the output is displayed

# Tutorial

I have tried using escape characters in Timestamp of request of `cli`

Command:
`./pgmoneta-cli -c pgmoneta.conf --format text status`

Output:
```
Header: 
  ClientVersion: 0.15.0
  Command: status
  Output: text
  Timestamp: "20241016234836"
Outcome: 
  Status: true
  Time: 00:00:00
Request: 

Response: 
  FreeSpace: 63.00GB
  NumberOfServers: 1
  Offline: false
  ServerVersion: 0.15.0
  Servers: 
    - HotStandbySize: 0
      NumberOfBackups: 0
      RetentionDays: 7
      RetentionMonths: Unspecified
      RetentionWeeks: Unspecified
      RetentionYears: Unspecified
      Server: primary
      ServerSize: 16.00MB
      Workers: 0
  TotalSpace: 164.74GB
  UsedSpace: 16.00MB
  Workers: 0
```

Command:
`./pgmoneta-cli -c pgmoneta.conf --format json status`

Output:
```
{
  "Header": {
    "ClientVersion": "0.15.0",
    "Command": "status",
    "Output": "json",
    "Timestamp": "\"20241016234843\""
  },
  "Outcome": {
    "Status": true,
    "Time": "00:00:00"
  },
  "Request": {},
  "Response": {
    "FreeSpace": "63.00GB",
    "NumberOfServers": 1,
    "Offline": false,
    "ServerVersion": "0.15.0",
    "Servers": [
      {
        "HotStandbySize": 0,
        "NumberOfBackups": 0,
        "RetentionDays": 7,
        "RetentionMonths": "Unspecified",
        "RetentionWeeks": "Unspecified",
        "RetentionYears": "Unspecified",
        "Server": "primary",
        "ServerSize": "16.00MB",
        "Workers": 0
      }
    ],
    "TotalSpace": "164.74GB",
    "UsedSpace": "16.00MB",
    "Workers": 0
  }
}
```

@Jubilee101 PTAL


src/libpgmoneta/value.c
```diff
@@ -456,14 +457,24 @@ string_to_string_cb(uintptr_t data, int32_t format, char* tag, int indent)
...
       if (format == FORMAT_JSON || format == FORMAT_JSON_COMPACT)
       {
-         snprintf(buf, MISC_LENGTH, "\"%s\"", str);
+         for (int i = 0; i < strlen(str); i++)
+         {
+            if (str[i] == '\'' || str[i] == '\"')
+            {
+               translated_ec_string = pgmoneta_append_char(translated_ec_string, '\\');
+            }
+            translated_ec_string = pgmoneta_append_char(translated_ec_string, str[i]);
```
Jubilee101, line 466: Append is kinda expensive, perhaps do a scan first to decide if you need translation, otherwise just do snprintf.

src/libpgmoneta/value.c
```diff
@@ -456,14 +457,24 @@ string_to_string_cb(uintptr_t data, int32_t format, char* tag, int indent)
    {
       if (format == FORMAT_JSON || format == FORMAT_JSON_COMPACT)
       {
-         snprintf(buf, MISC_LENGTH, "\"%s\"", str);
+         for (int i = 0; i < strlen(str); i++)
```
Jubilee101, line 460: `strlen()` is O(n) and you are calling it every loop, instead you should store the length as variable somewhere else.

src/libpgmoneta/json.c
```diff
@@ -620,6 +620,24 @@ parse_string(char* str, uint64_t* index, struct json** obj)
          // The key
          while (idx < len && str[idx] != '"')
```
Jubilee101, line 621: We should have more security check about this now in case there's a `"` in the middle that's missing preceding `\`

src/libpgmoneta/json.c
```diff
@@ -738,6 +756,25 @@ fill_value(char* str, char* key, uint64_t* index, struct json* o)
...
+            {
+               case '\'': // single quote
+               case '\"': // double quote
+                  val = pgmoneta_append_char(val, str[idx++]);
+                  break;
+               default:
+                  goto error;
+            }
+            continue;
+         }
```
Jubilee101, line 777: Looks good~ Perhaps wrap the code as a static handle_escape_char() to avoid duplicate code.

Jubilee101:
Looks like you are on the right track~

src/libpgmoneta/value.c
Jubilee101: I would prefer just calculate how much space I need, malloc the memory and copy the characters in place one by one. Then the whole thing would be O(n).

src/libpgmoneta/json.c
ashu3103: I think the rest of the code logic will take care of it unless someone deliberately introduce some weird cases like-

**(Extreme cases)**

`{"na":"me"}":"ashutosh"}` 
where expected key - `na":"me"}` and value is `ashutosh` 
while we will get key - `na` and value - `me`

or

`{"na":"me"," : ":"ashutosh"}`
where he expects key - `na":"me` and value - `:"ashutosh`
but will get key1 - `na`, value1 - `me` and key2 - `:`, value2 - `ashutosh`

While the second case is not invalid. The above case can be mitigated by parenthesis matching which is O(n).

@Jubilee101 Any thoughts on this

src/libpgmoneta/json.c
```diff
@@ -738,6 +743,11 @@ fill_value(char* str, char* key, uint64_t* index, struct json* o)
       idx++;
       while (idx < len && str[idx] != '"')
       {
+         // check for escape characters in value
+         if(handle_escape_char(str, &idx, len))
```
Jubilee101, line 747: Does the logic still work when you are dealing with `\t` or `\n`?

Perhaps you should still do append inside and only go into the function when you actually find the `\`.

src/libpgmoneta/json.c
Jubilee101: > {"na":"me"}":"ashutosh"}

Current logic is more of a "max effort" way, that it parses as much as it can, and stops when it has finshed parsing an entire json item (when it sees the enclosing `}`), regardless of if there are more gibberish left to parse. The fix should be easy, if [the parsing has finished](https://github.com/pgmoneta/pgmoneta/blob/b1f97d07a8ea71a9da0d5517d9716d104d0e1d1f/src/libpgmoneta/json.c#L539) but there are still `!isspace()` characters left, the function should error out. You can try and see if it solves the problem, otherwise we open a separate issue to address it.

> {"na":"me"," : ":"ashutosh"}

Nah this is a valid case, if you forget to use `\` and end up getting something unexpected, we are not responsible for that. 

I was mostly thinking about simpler cases like `"ab"c":"DE"`, if the later logic takes care of it then we can just leave it.

ashu3103: Hmm, will try that

src/libpgmoneta/json.c
ashu3103: > Does the logic still work when you are dealing with `\t` or `\n`?
> 

Actually I was not thinking of implementing for `\n` and `\t` at all.

> Perhaps you should still do append inside and only go into the function when you actually find the `\`.

In that case too when we encounter `\n` or `\t` what will we append to the key or value as a character? This is where I was stuck and thought to not provide support for `\n` and `\t`. Like with what character we will replace `\n` and `\t`.

Jubilee101: Append `'\n'` or `'\t'` of course~ It's still one character.

Jubilee101: Just note that in to_string you don't have to do any conversion, since printf() recognizes them and takes care of it.

src/libpgmoneta/json.c
```diff
@@ -844,6 +861,36 @@ fill_value(char* str, char* key, uint64_t* index, struct json* o)
...
+   }
+   switch (str[idx])
+   {
+      case '\'': // single quote
+         key = pgmoneta_append_char(key, '\'');
+         break;
+      case '\"': // double quote
+         key = pgmoneta_append_char(key, '\"');
+         break;
+      case '\n': // line feed
```
Jubilee101, line 881: This doesn't look right, have you tested it?

ashu3103: Yeah I know, still working on it. I’ll ping you once done.

ashu3103: Yeah, It will not capture any escape character at the starting of the string. (since key is NULL).

For that I am thinking of taking the corresponding escape character (after parsing) outside of `handle_escape_char` and then append it.

Jubilee101: No you missed the point. The string comes in as something like "\n", now you've seen the '\', so in the switch cases you should be checking for 'n' or 't', not '\n' or '\t'.

>  it will not capture any escape character at the starting of the string. (since key is NULL).

If you are talking about `pgmoneta_append_char()`, the function works even with NULL string.

ashu3103: > No you missed the point. The string comes in as something like "\n", now you've seen the `'\'`, so in the switch cases you should be checking for 'n' or 't', not '\n' or '\t'.
> 

Yeah, this is wrong too, I might have missed it because while testing it, it worked totally fine.

It should have given me -
```
ERROR main.c:830 Management: Bad payload (1)
```
but actually it is not even entering the `handle_escape_char` function. Maybe while parsing `\n` is treated as single character.

Anyways will make the change. Thanks for pointing it!

> > it will not capture any escape character at the starting of the string. (since key is NULL).
> 
> If you are talking about `pgmoneta_append_char()`, the function works even with NULL string.

True, but in a case where we are sending an escape character at the starting of the string and passing pointer of `key/value` (which don't point to anything --> NULL) to the function `handle_escape_char` will not affect the value of the pointer and so it will still remain `NULL`.

Because in the caller the pointer was NULL and in the callee (where we are using malloc (realloc will call malloc if NULL is provided)) its reference changed.

Jubilee101: > but actually it is not even entering the handle_escape_char function. Maybe while parsing \n is treated as single character.

Right, so if you want to build a test cases in C you should use something like "\\n" to escape the backlash, you need to check other characters as well.

> Because in the caller the pointer was NULL and in the callee (where we are using malloc (realloc will call malloc if NULL is provided)) its reference changed.

That's a good point, the current implemention looks correct.

I think we still need `\\` and `\r`. Be careful with cases like `"\\""`, which would be illegal since the backslash itself is escaped.

ashu3103: > 
> Right, so if you want to build a test cases in C you should use something like `"\\n"` to escape the backlash, you need to check other characters as well.
> 

What I am actually thinking is, while parsing if the character is treated as `\n` it won't simply go inside the handle_escape_character but will be stored as desired, while if we encounter `\n` as two separate characters `\` and `n` (for example in case where we are taking input from stdin or reading from a file) in that case it will get inside `handle_escape_character` and we will simply have to store `\n` character in place of `\n` (both characters treated individually).

> 
> That's a good point, the current implemention looks correct.
> 
> I think we still need `\\` and `\r`. Be careful with cases like `"\\""`, which would be illegal since the backslash itself is escaped.

In the my latest draft I have added checks for illegal use case.

Jubilee101: > What I am actually thinking is, while parsing if the character is treated as \n it won't simply go inside the handle_escape_character but will be stored as desired, while if we encounter \n as two separate characters \ and n (for example in case where we are taking input from stdin or reading from a file) in that case it will get inside handle_escape_character and we will simply have to store \n character in place of \n (both characters treated individually).

Yes, but you still need to test it and make sure it works.

> In the my latest draft I have added checks for illegal use case.

Sounds good, just push when it's ready.

ashu3103: @Jubilee101 PTAL

src/libpgmoneta/value.c
```diff
@@ -456,14 +460,60 @@ string_to_string_cb(uintptr_t data, int32_t format, char* tag, int indent)
...
+               case '\'':
+               case '\"':
+                  translated_ec_string[idx] = '\\';
+                  idx++;
+                  translated_ec_string[idx] = str[i];
+                  break;
+               case '\n':
+                  translated_ec_string[idx] = '\\';
+                  idx++;
+                  translated_ec_string[idx] = 'n';
```
Jubilee101, line 488: Should we translate this? Perhaps we can just leave it... I remember you said parsing works with char `\n`. Same comment goes to other white space characters.

src/libpgmoneta/json.c
```diff
@@ -738,7 +744,12 @@ fill_value(char* str, char* key, uint64_t* index, struct json* o)
       idx++;
       while (idx < len && str[idx] != '"')
       {
-         val = pgmoneta_append_char(val, str[idx++]);
+         char ec_ch;
+         if (handle_escape_char(str, &idx, len, &ec_ch))
```
Jubilee101, line 748: I'm not quite sure about handling non-escaped characters inside the function as well. It's a little ambiguous, like you are treating normal `ch` as `ec_ch` as well. 

I still think it's better to only go into this handler when you encounter the backslash.

Jubilee101: And you get rid of the `escaped` flag this way.

src/libpgmoneta/value.c
Jubilee101: I think we can leave it as long as it can be read or parsed with no ambiguity, we can worry about aesthetic later. Thoughts?

src/libpgmoneta/json.c
ashu3103: > I'm not quite sure about handling non-escaped characters inside the function as well. It's a little ambiguous, like you are treating normal `ch` as `ec_ch` nontheless.
> 
> I still think it's better to only go into this handler when you encounter the backslash.

Okay, I think maybe we can just change the name of the function if it is misleading. we can make it something like ‘get_next_char’

ashu3103: > And you get rid of the `escaped` flag this way.

the main idea of escape flag was to detect any illegal use case.

src/libpgmoneta/value.c
ashu3103: It is required while formatting in json as per the specs - https://www.json.org/json-en.html

because if not done so it will treat `\n` as a newline which we don't want in a string according to the specs

Jubilee101: Ok then, so we are good

src/libpgmoneta/json.c
Jubilee101: > Okay, I think maybe we can just change the name of the function if it is misleading. we can make it something like ‘get_next_char’

Yeah, but in terms of helper functions I think it's better to have their scopes smaller, instead of something that does a little bit of everything.

> the main idea of escape flag was to detect any illegal use case.

Well, you can still do that without it, If you only go into function when the condition is met, right?. The code would be cleaner.

ashu3103: > > Okay, I think maybe we can just change the name of the function if it is misleading. we can make it something like ‘get_next_char’
> 
> Yeah, but in terms of helper functions I think it's better to have their scopes smaller, instead of something that does a little bit of everything.
> 
> > the main idea of escape flag was to detect any illegal use case.
> 
> Well, you can still do that without it, If you only go into function when the condition is met, right?. The code would be cleaner.


Okay on it.

src/libpgmoneta/json.c
```diff
@@ -620,6 +621,23 @@ parse_string(char* str, uint64_t* index, struct json** obj)
...
+            {
+               if (handle_escape_char(str, &idx, len, &ec_ch))
+               {
+                  goto error;
+               }
+               key = pgmoneta_append_char(key, ec_ch);
+               continue;
+            }
+            // check for illegal use cases
+            if (str[idx] == '\'' || str[idx] == '\"' || str[idx] == '\\')
```
ashu3103, line 636: This step is necessary in case where we get something like `{"name":"as'hu"}` here `'` is not escaped so will bypass the `if (str[idx] == '\\')` condition and fallback to this check.

`str[idx] == "\""` is redundant however

Jubilee101: Seems that `'` is not required to be escaped in https://www.json.org/json-en.html. Also it won't cause any ambiguity when we only use double quotes for string enclosement. I think we can allow it. As for `\`, I can't really think of cases where it falls into the `if` condition, insteading of going into `handle_escape_char` and then erroring out. Though feel free to raise counter example. In terms of `"`, yeah I think we've talked about it being redundant.

If none of the case is necessary, we can safely get rid of the check.

src/libpgmoneta/json.c
```diff
@@ -844,6 +878,40 @@ fill_value(char* str, char* key, uint64_t* index, struct json* o)
...
+   idx++;
+   if (idx == len)   // security check
+   {
+      return 1;
+   }
+
+   // Check the next character after checking '\' character
+   switch (str[idx])
+   {
+      case '\'':
```
Jubilee101, line 894: This can go if the json standard allows it.

src/libpgmoneta/json.c
ashu3103: > Seems that `'` is not required to be escaped in https://www.json.org/json-en.html. Also it won't cause any ambiguity when we only use double quotes for string enclosement. I think we can allow it. As for `\`, I can't really think of cases where it falls into the `if` condition, insteading of going into `handle_escape_char` and then erroring out. Though feel free to raise counter example. In terms of `"`, yeah I think we've talked about it being redundant.
> 
> If none of the case is necessary, we can safely get rid of the check.

Yeah, absolutely correct, We actually don't need that extra check, my bad.

src/libpgmoneta/json.c
ashu3103: We can get rid of this as well

Jubilee101: Yeah we probably should not include single quote in the logic at all. Instead we should probably error out when it's escaped. Perhaps you need to check all the other logic related to it. It's my bad, the initial comment in the issue is misleading, apologies :)

ashu3103: No issue, let's get it done :)

Jubilee101:
You missed the keys. They need to be escaped as well when converted to string output. See [art_to_json_string_cb](https://github.com/pgmoneta/pgmoneta/blob/40046f9fda3c08f64c957cc49874fb6776c97ebf/src/libpgmoneta/art.c#L1467).

I can think of two ways to do this without duplicate code, either you wrap your translation logic to a utility function inside `utils.c`, or you can create a temporary `value` instance to wrap the key as ValueString, and call `pgmoneta_value_to_string`, which will call the `string_to_string_cb` you just modified, wichever you would like, though remember to destroy the temp value after you are done.

Jubilee101:
I think we're very close~

Jubilee101:
Also can you double check the validation? I changed `MANAGEMENT_CATEGORY_HEADER` from `"Header"` to `"Header\"\n\t\""`. When cli sent json payload to main side without correct translation of the key, `pgmoneta_management_read_json` on main side should have errored out because of parsing. Instead seems that it returned successfully.

ashu3103:
> You missed the keys. They need to be escaped as well when converted to string output. See [art_to_json_string_cb](https://github.com/pgmoneta/pgmoneta/blob/40046f9fda3c08f64c957cc49874fb6776c97ebf/src/libpgmoneta/art.c#L1467).
> 

Oh I see, actually I was assuming that key and value are translated in the same so never run any tests on key, anyways my bad.

> I can think of two ways to do this without duplicate code, either you wrap your translation logic to a utility function inside `utils.c`, or you can create a temporary `value` instance to wrap the key as ValueString, and call `pgmoneta_value_to_string`, which will call the `string_to_string_cb` you just modified, wichever you would like, though remember to destroy the temp value after you are done should you choose the latter.

The former approach seems great -- a function which will take a `char* originalStr` as an argument, translate it and return the translated string.

Smart Move!

ashu3103:
@Jubilee101 PTAL

src/libpgmoneta/utils.c
```diff
@@ -4037,6 +4037,60 @@ pgmoneta_indent(char* str, char* tag, int indent)
    return str;
 }
 
+char*
+pgmoneta_translate_json_format(char* str)
+{
```
Jubilee101, line 4042: Some NULL check for the str maybe? In case we use it somewhere else later.

Jubilee101:
> actually I was assuming that key and value are translated in the same so never run any tests on key, anyways my bad

Yeah, the key is not wrapped in the `value` instance like the values. Maybe it's some design we can improve upon.

Anyways the function seems to work nicely. Just some minor issue with the naming~

Jubilee101:
Merged, thanks for your contribution, great work!

Issue #401, opened 2024-10-14
CLI: conf get | set | ls


jesperpedersen:
Create a port of pgagroal's

```
  conf <action>            Manages the configuration (e.g., reloads the configuration
                           The subcommand <action> can be:
                           - 'reload' to issue a configuration reload;
                           - 'get' to obtain information about a runtime configuration value;
                                   conf get <parameter_name>
                           - 'set' to modify a configuration value;
                                   conf set <parameter_name> <parameter_value>;
                           - 'ls'  lists the configuration files used.
```

functionality.

We have `reload` already

jesperpedersen:
@MohanadKh03 Can you take this one ?

@Jubilee101 can help with core stuff

MohanadKh03:
Sure! 
Regarding the networking part though when writing the commands into pgmoneta's socket should it be the same as pgagroal's way ? I can see some pattern but correct me if im wrong
1.Send the header
2.Send the size of the payload
3.Send the payload itself
4.Read the response as a json
5.Extract whatever needed and then print it out in the console

Is that correct ? **assuming no failures/errors in any of the steps**

jesperpedersen:
Yes, the pgmoneta way is the new way of doing it, so you need to extract information from the "Response" part of the JSON payload based on the input. That is the easy part - think of `conf get` as a simple https://jqlang.github.io/jq/ command.

The harder part is the `conf set` command, but it shouldn't be that hard either - we need to send a request with the key and the new value

jesperpedersen:
Remember to add yourself to the authors info...

Jubilee101:
Just let me know if you need anything, welcome! @MohanadKh03

jesperpedersen:
Yes, lets get the full solution now because we need a chapter in the advanced manual that describes how to get the configuration, a field and how to set a field.

The manual work can be separate from this

jesperpedersen:
@MohanadKh03 Where are you on this ?

MohanadKh03:
Hey @jesperpedersen My bad I have been pretty busy and haven’t had a chance to continue working on this. It might be better if someone else takes over to keep things going and not block progress. Thanks and apologies for the late progress.

jesperpedersen:
@MohanadKh03 Thanks for the update !

ashu3103:
@jesperpedersen I can continue with this issue, kindly assign me this.

jesperpedersen:
@ashu3103 Ok, but this is an important task very soon

[...]

PR #405, opened 2024-10-16
[#401] CLI: conf get | set | ls

MohanadKh03:
Currently this contains the logic for ``conf get`` which works as expected and similar to the other commands. If all is good with this way it will basically be the same idea for ``conf set`` and ``conf ls``  

Got a problem with double free though not sure I tried debugging through the changed parts and even stopped freeing anything myself in ``configuration.c`` , But still this comes up ``free(): double free detected in tcache 2`` so any ideas how I can fix this ? Is there a better way to look through this other than gdb or manual printing/debugging ?

### DIFF
--- a/src/cli.c
+++ b/src/cli.c
@@ -115,6 +115,7 @@ static int details(SSL* ssl, int socket, uint8_t compression, int32_t output_for
 static int ping(SSL* ssl, int socket, uint8_t compression, int32_t output_format);
 static int reset(SSL* ssl, int socket, uint8_t compression, int32_t output_format);
 static int reload(SSL* ssl, int socket, uint8_t compression, int32_t output_format);
+static int config_get(SSL* ssl, int socket,char* config_key, uint8_t compression, int32_t output_format);
 static int retain(SSL* ssl, int socket, char* server, char* backup_id, uint8_t compression, int32_t output_format);
 static int expunge(SSL* ssl, int socket, char* server, char* backup_id, uint8_t compression, int32_t output_format);
 static int decrypt_data(SSL* ssl, int socket, char* path, uint8_t compression, int32_t output_format);
@@ -331,6 +332,14 @@ const struct pgmoneta_command command_table[] = {
       .action = MANAGEMENT_RELOAD,
       .deprecated = false,
       .log_message = "<conf reload>"
+   },
+   {
+      .command = "conf",
+      .subcommand = "get",
+      .accepted_argument_count = {1},
+      .action = MANAGEMENT_CONFIG_GET,
+      .deprecated = false,
+      .log_message = "<conf get>"
    },
    {
       .command = "clear",
@@ -736,6 +745,10 @@ password:
    else if (parsed.cmd->action == MANAGEMENT_RELOAD)
    {
       exit_code = reload(s_ssl, socket, compression, output_format);
+   }
+   else if(parsed.cmd->action == MANAGEMENT_CONFIG_GET)
+   {
+      exit_code = config_get(s_ssl,socket,parsed.args[0],compression,output_format);
    }
    else if (parsed.cmd->action == MANAGEMENT_RETAIN)
    {
@@ -1264,6 +1277,25 @@ reload(SSL* ssl, int socket, uint8_t compression, int32_t output_format)
       goto error;
    }
 
+   if (process_result(ssl, socket, output_format))
+   {
+      goto error;
+   }
+
+   return 0;
+
+error:
+
+   return 1;
+}
+
+static int
+config_get(SSL* ssl, int socket,char* config_key, uint8_t compression, int32_t output_format)
+{
+   if (pgmoneta_management_conf_get(ssl, socket, config_key, compression, output_format))
+   {
+      goto error;
+   }
    if (process_result(ssl, socket, output_format))
    {
       goto error;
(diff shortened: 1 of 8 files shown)


jesperpedersen:
I think we need to think about how we can send the entire configuration file over, and then filter the key we need out. So, it is like the translate feature, f.ex. `conf get host` will give `[pgmoneta]`'s `host` setting, and `conf get primary.host` will give `[primary]`'s `host` setting.

`conf ls` is a lot easier.

`conf set` can either get the configuration, set the value, and send it back - or - just the single value back. Doing the former is likely better long term

jesperpedersen:
For the double `free()` - make sure that you compile with `Debug` and `clang`, and maybe track the `free()` with a `bool`

MohanadKh03:
> I think we need to think about how we can send the entire configuration file over, and then filter the key we need out. 

Can we not just filter the arguments passed by section (if found then we find the corresponding key otherwise ``pgmoneta``'s section by default) and the config key instead of sending the whole file and then filtering what we need ?  I think sending the whole configuration just to get a single value is not worth it  we could just do a single key-value 
WDYT ? @jesperpedersen

jesperpedersen:
@MohanadKh03 Sure, we can start that way.

I'm just thinking further ahead, where a client want all values, and the server version might be different than the client - or even not our CLI

jesperpedersen:
Look at functions like https://github.com/pgmoneta/pgmoneta/blob/main/src/cli.c#L1475 and friends, and think about passing the JSON request / result from `cli.c` -> `management.c` -> `main.c` -> `configuration.c`.

The two middle ones should almost be a copy'n'paste...

jesperpedersen:
I would rather see if we could get it correct the first time.

For `set` you would transform the JSON into a `struct configuration` and call https://github.com/pgmoneta/pgmoneta/blob/main/src/libpgmoneta/configuration.c#L3037

MohanadKh03:
Not sure If i got this right but this is how I see it
We first create the request in ``cli.c`` sending only the key in the payload e.g: ``conf get host`` , ``conf get primary.host`` by using ``management.c`` then this is read in ``main.c``'s event loop and the payload sent is received by calling the function in ``configuration.c`` which should be modified rn to accept other servers e.g ``conf get primary.host`` and put the corresponding value (if found) )in its response right ?
So basically now I just need to fix the receiving part am I right ? ``conf ls`` it is pretty easy and ``conf set`` is just sending the configuration file over (tbh I think just sending the key is enough but since you think sending it over is better for the long term then alright) to the receiver part and then at ``configuration.c`` it calls https://github.com/pgmoneta/pgmoneta/blob/main/src/libpgmoneta/configuration.c#L3037

Do I get this correct now ? Cuz I got kind of confused when reading the ``translate`` parts in ``cli.c``

jesperpedersen:
Yeah, so `conf get` will get the whole configuration as a JSON object, so don't need the "3rd" parameter. Then `cli.c` will filter out the correct value based on the "3rd".

`conf set` will do `conf get`, modify, and send back, then `main.c` will delegate to `configuration.c` to change the actual value if possible

jesperpedersen:
The "translate" part is the extraction of the value from `conf get XYZ` - think https://jqlang.github.io/jq/ - as SMALL as you can make it

jesperpedersen:
it would be the same in `doc/CONFIGURATION.md` and `doc/manual/user-02-configuration.md`

MohanadKh03:
Yes ill handle the docs for it in the Advanced Guide 
Apologies for being late though as I am a bit busy but by the end of the week it will be done both implementation and docs

jesperpedersen:
Just take your time to find the best solution

MohanadKh03:
@jesperpedersen Sorry for being late

 I tried the whole thing again and implemented the ``conf ls`` easily but in the ``conf get`` and ``conf set`` you said we should 
 > I think we need to think about how we can send the entire configuration file over, and then filter the key we need out. So, it is like the translate feature, f.ex. conf get host will give [pgmoneta]'s host setting, and conf get primary.host will give [primary]'s host setting.

but if we go through this approach starting from ``cli.c`` -> ``management.c`` -> ``main.c`` -> ``configuration.c`` we would not send the configuration file over right ? we will just receive the key-value that is given to us , So could you please elaborate what exactly do you mean by send the entire configuration file over? Do you mean not sending the ``config_key`` in here ? https://github.com/pgmoneta/pgmoneta/blob/33b9fde6d0951c7e9a40631518465fa60d3ab567/src/libpgmoneta/management.c#L482
As far as I understand clients - even non-CLI clients - will be forced to write their own logic of handling the config file so I think it would be better handling the config file at ``configuration.c`` which is kind of the pgmoneta-server-side right ?

jesperpedersen:
You need to make a representation of `pgmoneta.conf` in JSON format.

So, in `management.h` you will do

```
#define MANAGEMENT_CONFIGURATION_HOST  "Host"
#define MANAGEMENT_CONFIGURATION_PORT  "Port"
...
```

and then in `configuration.c` in `pgmoneta_configuration_get` fill out the JSON document with `struct configuration` and include the `struct server`s too.

The `pgmoneta_configuration_set` you should get a key/value pair in the `Request` part that you parse to an entry in `struct configuration` or in one of the `struct server`s

jesperpedersen:
Superseeded by #409

PR #409, opened 2024-11-06
[#401] CLI: conf ls | get | set

ashu3103:
#### In Progress

- added support for new cli command - `conf ls`

#### TODO

- [x] add support for `conf get`
- [x] add support for `conf set`

@jesperpedersen PTAL

### DIFF
--- a/src/cli.c
+++ b/src/cli.c
@@ -123,8 +123,17 @@ static int decompress_data(SSL* ssl, int socket, char* path, uint8_t compression
 static int compress_data(SSL* ssl, int socket, char* path, uint8_t compression, uint8_t encryption, int32_t output_format);
 static int info(SSL* ssl, int socket, char* server, char* backup, uint8_t compression, uint8_t encryption, int32_t output_format);
 static int annotate(SSL* ssl, int socket, char* server, char* backup, char* command, char* key, char* comment, uint8_t compression, uint8_t encryption, int32_t output_format);
+static int conf_ls(SSL* ssl, int socket, uint8_t compression, uint8_t encryption, int32_t output_format);
+static int conf_get(SSL* ssl, int socket, char* config_key, uint8_t compression, uint8_t encryption, int32_t output_format);
+static int conf_set(SSL* ssl, int socket, char* config_key, char* config_value, uint8_t compression, uint8_t encryption, int32_t output_format);
 
-static int  process_result(SSL* ssl, int socket, int32_t output_format);
+static int process_result(SSL* ssl, int socket, int32_t output_format);
+static int process_get_result(SSL* ssl, int socket, char* param, int32_t output_format);
+static int process_ls_result(SSL* ssl, int socket, int32_t output_format);
+static int process_set_result(SSL* ssl, int socket, char* config_key, int32_t output_format);
+
+static int get_config_key_result(char* config_key, struct json* j, uintptr_t* r, int32_t output_format);
+static int get_conf_path_result(struct json* j, uintptr_t* r);
 
 static char* translate_command(int32_t cmd_code);
 static char* translate_output_format(int32_t out_code);
@@ -132,8 +141,15 @@ static char* translate_valid(int32_t valid);
 static char* translate_compression(int32_t compression_code);
 static char* translate_encryption(int32_t encryption_code);
 static char* translate_size(int64_t size);
+static char* translate_storage_engine(int32_t storage_engine);
+static char* translate_create_slot(int32_t create_slot);
+static char* translate_hugepage(int32_t hugepage);
+static char* translate_log_type(int32_t log_type);
+static char* translate_log_level(int32_t log_level);
+static char* translate_log_mode(int32_t log_mode);
 static char* int_to_hex(uint32_t num);
 static void translate_backup_argument(struct json* j);
+static void translate_configuration(struct json* j);
 static void translate_response_argument(struct json* j);
 static void translate_servers_argument(struct json* j);
 static void translate_server_retention_argument(struct json* j, char* tag);
@@ -190,6 +206,11 @@ usage(void)
    printf("  status [details]         Status of pgmoneta, with optional details\n");
    printf("  conf <action>            Manage the configuration, with one of subcommands:\n");
    printf("                           - 'reload' to reload the configuration\n");
+   printf("                           - 'ls' to print the configurations used\n");
+   printf("                           - 'get' to obtain information about a runtime configuration value\n");
+   printf("                             conf get <parameter_name>\n");
+   printf("                           - 'set' to modify a configuration value;\n");
+   printf("                             conf set <parameter_name> <parameter_value>;\n");
    printf("  clear <what>             Clear data, with:\n");
    printf("                           - 'prometheus' to reset the Prometheus statistics\n");
    printf("\n");
@@ -335,6 +356,30 @@ const struct pgmoneta_command command_table[] = {
       .log_message = "<conf reload>"
    },
    {
+      .command = "conf",
+      .subcommand = "ls",
+      .accepted_argument_count = {0},
+      .action = MANAGEMENT_CONF_LS,
+      .deprecated = false,
+      .log_message = "<conf ls>"
+   },
+   {
+      .command = "conf",
+      .subcommand = "get",
+      .accepted_argument_count = {0, 1},
+      .action = MANAGEMENT_CONF_GET,
+      .deprecated = false,
+      .log_message = "<conf get> [%s]"
+   },
+   {
+      .command = "conf",
+      .subcommand = "set",
+      .accepted_argument_count = {2},
+      .action = MANAGEMENT_CONF_SET,
+      .deprecated = false,
+      .log_message = "<conf set> [%s]"
+   },
+   {
       .command = "clear",
       .subcommand = "prometheus",
       .accepted_argument_count = {0},
@@ -378,7 +423,6 @@ main(int argc, char** argv)
    int c;
    int option_index = 0;
    /* Store the result from command parser*/
-   bool matched = false;
    size_t size;
    char un[MAX_USERNAME_LENGTH];
    struct configuration* config = NULL;
@@ -623,8 +667,6 @@ main(int argc, char** argv)
       goto done;
    }
 
-   matched = true;
-
    if (configuration_path != NULL)
    {
       /* Local connection */
@@ -800,6 +842,25 @@ password:
    {
       exit_code = annotate(s_ssl, socket, parsed.args[0], parsed.args[1], parsed.args[2], parsed.args[3], parsed.args[4], compression, encryption, output_format);
    }
+   else if (parsed.cmd->action == MANAGEMENT_CONF_LS)
+   {
+      exit_code = conf_ls(s_ssl, socket, compression, encryption, output_format);
+   }
+   else if (parsed.cmd->action == MANAGEMENT_CONF_GET)
+   {
+      if (parsed.args[0])
+      {
+         exit_code = conf_get(s_ssl, socket, parsed.args[0], compression, encryption, output_format);
+      }
+      else
+      {
+         exit_code = conf_get(s_ssl, socket, NULL, compression, encryption, output_format);
+      }
+   }
+   else if (parsed.cmd->action == MANAGEMENT_CONF_SET)
+   {
+      exit_code = conf_set(s_ssl, socket, parsed.args[0], parsed.args[1], compression, encryption, output_format);
+   }
 
 done:
 
@@ -817,15 +878,6 @@ done:
    }
 
    pgmoneta_disconnect(socket);
-
-   if (configuration_path != NULL)
-   {
-      if (matched && exit_code != 0)
-      {
-         warnx("No connection to pgmoneta on %s", config->unix_socket_dir);
-      }
-   }
-
    pgmoneta_stop_logging();
    pgmoneta_destroy_shared_memory(shmem, size);
 
@@ -959,6 +1011,9 @@ help_conf(void)
 {
    printf("Manage the configuration\n");
    printf("  pgmoneta-cli conf [reload]\n");
+   printf("  pgmoneta-cli conf [ls]\n");
+   printf("  pgmoneta-cli conf [get] <parameter_name>\n");
+   printf("  pgmoneta-cli conf [set] <parameter_name> <parameter_value>\n");
 }
 
 static void
@@ -1468,6 +1523,66 @@ error:
 }
 
 static int
+conf_ls(SSL* ssl, int socket, uint8_t compression, uint8_t encryption, int32_t output_format)
+{
+   if (pgmoneta_management_request_conf_ls(ssl, socket, compression, encryption, output_format))
+   {
+      goto error;
+   }
+
+   if (process_ls_result(ssl, socket, output_format))
+   {
+      goto error;
+   }
+
+   return 0;
+
+error:
+
+   return 1;
+}
+
+static int
+conf_get(SSL* ssl, int socket, char* config_key, uint8_t compression, uint8_t encryption, int32_t output_format)
+{
+   if (pgmoneta_management_request_conf_get(ssl, socket, compression, encryption, output_format))
+   {
+      goto error;
+   }
+
+   if (process_get_result(ssl, socket, config_key, output_format))
+   {
+      goto error;
+   }
+
+   return 0;
+
+error:
+
+   return 1;
+}
+
+static int
+conf_set(SSL* ssl, int socket, char* config_key, char* config_value, uint8_t compression, uint8_t encryption, int32_t output_format)
+{
+   if (pgmoneta_management_request_conf_set(ssl, socket, config_key, config_value, compression, encryption, output_format))
+   {
+      goto error;
+   }
+
+   if (process_set_result(ssl, socket, config_key, output_format))
+   {
+      goto error;
+   }
+
+   return 0;
+
+error:
+
+   return 1;
+}
+
+static int
 process_result(SSL* ssl, int socket, int32_t output_format)
 {
    struct json* read = NULL;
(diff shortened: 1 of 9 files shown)


jesperpedersen:
Yes, this looks correct.

Continue alongs these lines, add the `get` to `configuration.h` and transform the configuration into JSON (main and servers). Then do the filtering in `cli.c`

`set` will belong there too in the end

jesperpedersen:
So, `fork()` in `main.c` with a call to `pgmoneta_configuration_get_json()` and so on

jesperpedersen:
See `_STATUS` how to do it

jesperpedersen:
@ashu3103 How is the `get` work going ?

ashu3103:
> @ashu3103 How is the `get` work going ?

I am able to get the main and servers configuration to cli in json only the filtering part is remaining.

jesperpedersen:
Great, sounds like you are making good progress !

jesperpedersen:
No, `CONF_GET` only has the command - all filtering is done in `cli.c`

ashu3103:
@jesperpedersen , I actually wanted a suggestion from you, After getting JSON configuration in `cli` we are stepping into `process_result` function (where read and print happens), so now the filtering should come somewhere in between reading and printing the response, in this case shall I create a new function `filter_process_result` (which read, filters and prints the response) or it has to be done in some other way?

jesperpedersen:
@ashu3103 Yes, just create a new `filter_process_result` function that `CONF_GET` uses. `CONF_GET` is special anyway, because it will output the single result of the query, or an error.

So, read, transform and then filter is a good way forward

jesperpedersen:
Yes, you do

```
  "Response": {
    "Configuration": {
      "pgmoneta": {
```

->

```
  "Response": {
    "Configuration": {
```

and

```
"primary": {
```

->

```
"Servers": {
      "primary": {
```

So, you would write

```
... conf get Servers.primary.host
```

to get that field.

Try and start with `pgmoneta_json_to_string(struct json* object, int32_t format, char* tag, int indent)` with `FORMAT_TEXT` and indent of `0`

@Jubilee101 Other ideas ?

ashu3103:
@jesperpedersen PTAL

src/cli.c
```diff
@@ -958,7 +987,7 @@ static void
 help_conf(void)
 {
    printf("Manage the configuration\n");
-   printf("  pgmoneta-cli conf [reload]\n");
+   printf("  pgmoneta-cli conf [reload | ls]\n");
```
jesperpedersen, line 990: Add 'get'

src/libpgmoneta/conf.c
```diff
@@ -0,0 +1,103 @@
+/*
```
jesperpedersen, line 1: You can put this in `configuration.c`

src/libpgmoneta/configuration.c
```diff
@@ -2012,6 +2021,146 @@ pgmoneta_reload_configuration(bool* restart)
...
+      pgmoneta_json_put(server_conf, CONFIGURATION_ARGUMENT_WORKERS, (uintptr_t)config->servers[i].workers, ValueInt64);
+      pgmoneta_json_put(server_conf, CONFIGURATION_ARGUMENT_BACKUP_MAX_RATE, (uintptr_t)config->servers[i].backup_max_rate, ValueInt64);
+      pgmoneta_json_put(server_conf, CONFIGURATION_ARGUMENT_NETWORK_MAX_RATE, (uintptr_t)config->servers[i].network_max_rate, ValueInt64);
+      pgmoneta_json_put(server_conf, CONFIGURATION_ARGUMENT_MANIFEST, (uintptr_t)config->servers[i].manifest, ValueInt64);
+      pgmoneta_json_put(server_conf, CONFIGURATION_ARGUMENT_TLS_CERT_FILE, (uintptr_t)config->servers[i].tls_cert_file, ValueString);
+      pgmoneta_json_put(server_conf, CONFIGURATION_ARGUMENT_TLS_CA_FILE, (uintptr_t)config->servers[i].tls_ca_file, ValueString);
+      pgmoneta_json_put(server_conf, CONFIGURATION_ARGUMENT_TLS_KEY_FILE, (uintptr_t)config->servers[i].tls_key_file, ValueString);
+      pgmoneta_json_put(server_conf, CONFIGURATION_ARGUMENT_EXTRA, (uintptr_t)config->servers[i].extra, ValueString);
+
+      pgmoneta_json_put(server_conf_j, config->servers[i].name, (uintptr_t)server_conf, ValueJSON);
```
jesperpedersen, line 2145: You need to use `json_append`

src/include/conf.h
```diff
@@ -0,0 +1,66 @@
+/*
```
jesperpedersen, line 1: This can go into `configuration.h`

jesperpedersen (changes requested):

jesperpedersen:
You are missing some documentation - search for `git grep pgmoneta-cli`.

You are going in the right direction !

fluca1978:
In case it helps, see this initial work for `conf get` <https://github.com/fluca1978/pgagroal/commit/ea07bd4eae1aa81c80da66ce05302501f57f4148>

ashu3103:
> @ashu3103 Actually, thinking some more about it we can just move everything to top-level, like
> 
> ```
> "Response": {
>    "metrics": true,
>    ...
>    "primary": {
>       "host": "localhost",
>       ...
>    }
> }
> ```
> 
> Should be a simple change

I am a bit confused at this point, if we bring all the keys to the top level and if we query `conf get host`, the response will already have `host: localhost` (example) in its top-level. So, what exactly we expect in `Response`.

ashu3103:
Also, can you tell me what is the need of bringing all the configuration to cli and do filtering at cli rather than at management handler of main.

jesperpedersen:
This is the JSON reply from the server, so yes - you can do `pgmoneta-cli conf get host` and that will give you the value in `struct configuration`::`host` - e.g.

pgmoneta.conf
```
[pgmoneta]
host = mypgmoneta
...

[primary]
host = mypsql
...
```

```
pgmoneta-cli conf get host
mypgmoneta

pgmoneta-cli conf get primary.host
mypgsql
```

So, you just have to remove the `Configuration` and `Servers` nodes, and move the data "up"

jesperpedersen:
Because then a client - other than pgmoneta-cli - can display the entire configuration with a single request

ashu3103:
> This is the JSON reply from the server, so yes - you can do `pgmoneta-cli conf get host` and that will give you the value in `struct configuration`::`host` - e.g.
> 
> pgmoneta.conf
> 
> ```
> [pgmoneta]
> host = mypgmoneta
> ...
> 
> [primary]
> host = mypsql
> ...
> ```
> 
> ```
> pgmoneta-cli conf get host
> mypgmoneta
> 
> pgmoneta-cli conf get primary.host
> mypgsql
> ```
> 

Okay, but isn't that different from other commands which prints `Header`, `Request`, `Response` etc. (just confirming....)

> So, you just have to remove the `Configuration` and `Servers` nodes, and move the data "up"

Yeah, that's done :+1:

jesperpedersen:
I think we will keep the output formats for `conf`, so

```
pgmoneta-cli -F text conf get encryption
none

pgmoneta-cli -F raw conf get encryption
0

pgmoneta-cli -F json conf get encryption
{"encryption": "none"}
```

jesperpedersen:
The CLI will only use the `Response` part for the `conf` command - `Header` and `Request` won't be displayed. So, it will be like my example above

ashu3103:
> I think we will keep the output formats for `conf`, so
> 
> ```
> pgmoneta-cli -F text conf get encryption
> none
> 
> pgmoneta-cli -F raw conf get encryption
> 0
> 
> pgmoneta-cli -F json conf get encryption
> {"encryption": "none"}
> ```

Yeah, this is better. Thanks!

src/cli.c
```diff
@@ -1502,6 +1571,439 @@ process_result(SSL* ssl, int socket, int32_t output_format)
...
+   {
+      configuration_js = (struct json*)pgmoneta_json_get(response, context);
+   }
+
+   if (!configuration_js)
+   {
+      goto error;
+   }
+
+   if (!strcmp(key, "host"))
```
Jubilee101, line 1700: This looks like a PITA.

An idea is that you can use iterator. Our json iterator works not only for json arrays but for json items as well. So just create an iterator and traverse the `configuration_js`, and put the string of the value into the response directly.

So something like:
```
if (!strcmp(key, iter->key) {
   pgmoneta_json_put(response, key, (uintptr_t)pgmoneta_value_to_string(iter->value), ValueString);
}
```
It'll save a lot of trouble for you. 

Sorry for proposing this 2 days later, after you've been through this... I've been busy with work recently.

Feel free to take the idea and see if you can make other parts simpler.

jesperpedersen:
Note, #416

src/cli.c
ashu3103: @Jubilee101, This was really helpful. Thanks a lot!

ashu3103:
@jesperpedersen PTAL

jesperpedersen:
Currently we have

```
pgmoneta-cli -F text conf get encryption
encryption: none

pgmoneta-cli -F raw conf get encryption
{
  "encryption": 0
}

pgmoneta-cli -F json conf get encryption
{
  "encryption": 0
}

```

instead of

```
pgmoneta-cli -F text conf get encryption
none

pgmoneta-cli -F raw conf get encryption
0

pgmoneta-cli -F json conf get encryption
{"encryption": "none"}
```

For the first 2 we only need the value, not the key - and the last has to be a JSON_COMPACT format

src/include/configuration.h
```diff
@@ -102,6 +164,31 @@ pgmoneta_validate_admins_configuration(void* shmem);
 int
 pgmoneta_reload_configuration(bool* restart);
 
+/**
+ * Get the json format of main configuration
+ * @param res The response of management
+ */
+void
+pgmoneta_add_configuration_response(struct json* res);
```
jesperpedersen, line 172: Can be `static int add_configuration_response()`

src/include/configuration.h
```diff
@@ -102,6 +164,31 @@ pgmoneta_validate_admins_configuration(void* shmem);
...
+
+/**
+ * Get the json format of servers configuration
+ * @param res The response of management
+ */
+void
+pgmoneta_add_servers_configuration_response(struct json* res);
```
jesperpedersen, line 179: Can be `static int add_servers_configuration_response()`

jesperpedersen (changes requested):

jesperpedersen:
I think for `conf ls` will use the same infrastructure as `conf get`.

So, `conf ls` will output

```
pgmoneta-cli conf ls
pgmoneta_admins.conf
pgmoneta.conf
pgmoneta_users.conf
```

But, you can also use

```
pgmoneta-cli conf get admins_configuration_path
pgmoneta-cli conf get main_configuration_path
pgmoneta-cli conf get users_configuration_path
```

to get the same output. For `conf ls` `-F text` and `-F raw` will be the same, and `-F json` will use the JSON_COMPACT format

ashu3103:
@jesperpedersen, PTAL

jesperpedersen:
Do a run with Debug with clang

```
    #1 0x7f12738e28f1 in pgmoneta_append src/libpgmoneta/utils.c:1165:15
    #2 0x7f12738e2c5c in pgmoneta_append_char src/libpgmoneta/utils.c:1186:11
    #3 0x7f127383f54d in get_retention_string src/libpgmoneta/configuration.c:3238:19
    #4 0x7f127383e618 in add_servers_configuration_response src/libpgmoneta/configuration.c:2107:83
    #5 0x7f127383c687 in pgmoneta_conf_get src/libpgmoneta/configuration.c:2144:4
    #6 0x56073b150c04 in accept_mgt_cb src/main.c:1186:10
```

```
    #1 0x7f9cda65bb6d in pgmoneta_json_create src/libpgmoneta/json.c:373:21
    #2 0x55e9e3f6783b in get_config_key_result src/cli.c:1688:8
    #3 0x55e9e3f673cb in filter_process_result src/cli.c:1604:8
    #4 0x55e9e3f6557d in conf_get src/cli.c:1543:8
    #5 0x55e9e3f63c9f in main src/cli.c:841:22
```

jesperpedersen:
`conf get` looks ok apart from the leaks, but `conf ls` is stitll the same

ashu3103:
> `conf get` looks ok apart from the leaks, but `conf ls` is stitll the same

Ohh I forgot to make changes for ‘ls’, my bad
On it!

jesperpedersen:
Once you are done with `conf get` then it is about `conf set` which is

```
pgmoneta-cli conf set key value
```

f.ex.

```
pgmoneta-cli conf set encryption aes
```

so that need to be translated before sent to the server to

```
Request: {
   "Key": "encryption",
  "Value": 0
}
``` 

and so on

Please, coordinate with @Jubilee101 in case there are something that can be improved...

ashu3103:
> Do a run with Debug with clang
> 
> ```
>     #1 0x7f12738e28f1 in pgmoneta_append src/libpgmoneta/utils.c:1165:15
>     #2 0x7f12738e2c5c in pgmoneta_append_char src/libpgmoneta/utils.c:1186:11
>     #3 0x7f127383f54d in get_retention_string src/libpgmoneta/configuration.c:3238:19
>     #4 0x7f127383e618 in add_servers_configuration_response src/libpgmoneta/configuration.c:2107:83
>     #5 0x7f127383c687 in pgmoneta_conf_get src/libpgmoneta/configuration.c:2144:4
>     #6 0x56073b150c04 in accept_mgt_cb src/main.c:1186:10
> ```
> 
> ```
>     #1 0x7f9cda65bb6d in pgmoneta_json_create src/libpgmoneta/json.c:373:21
>     #2 0x55e9e3f6783b in get_config_key_result src/cli.c:1688:8
>     #3 0x55e9e3f673cb in filter_process_result src/cli.c:1604:8
>     #4 0x55e9e3f6557d in conf_get src/cli.c:1543:8
>     #5 0x55e9e3f63c9f in main src/cli.c:841:22
> ```

Fixed, PTAL

jesperpedersen:
```
pgmoneta-cli -c pgmoneta.conf -F text conf ls
admin_configuration_path: ''
main_configuration_path: pgmoneta.conf
users_configuration_path: pgmoneta_users.conf

pgmoneta-cli -c pgmoneta.conf -F raw conf ls
{"admin_configuration_path":"","main_configuration_path":"pgmoneta.conf","users_configuration_path":"pgmoneta_users.conf"}

pgmoneta-cli -c pgmoneta.conf -F raw conf ls
{"admin_configuration_path":"","main_configuration_path":"pgmoneta.conf","users_configuration_path":"pgmoneta_users.conf"}

```

`-F text` and `-F raw` should only be the value -- they should both be the same

ashu3103:
> ```
> pgmoneta-cli -c pgmoneta.conf -F text conf ls
> admin_configuration_path: ''
> main_configuration_path: pgmoneta.conf
> users_configuration_path: pgmoneta_users.conf
> 
> pgmoneta-cli -c pgmoneta.conf -F raw conf ls
> {"admin_configuration_path":"","main_configuration_path":"pgmoneta.conf","users_configuration_path":"pgmoneta_users.conf"}
> 
> pgmoneta-cli -c pgmoneta.conf -F raw conf ls
> {"admin_configuration_path":"","main_configuration_path":"pgmoneta.conf","users_configuration_path":"pgmoneta_users.conf"}
> ```
> 
> `-F text` and `-F raw` should only be the value -- they should both be the same, so
> 
> ```
> 
> pgmoneta.conf
> pgmoneta_users.conf
> ```

But that way how one's gonna figure out which file is for which configuration path?

ashu3103:
> For `conf get` both `-F text` and `-F json` doesn't translate.
> 

Strange, the current logic is translating fields for me.

> `-F raw conf get` is ok

Can you send me the output at your end

ashu3103:
Screenshot from 2024-11-18 23-10-17

ashu3103:
> > Strange, the current logic is translating fields for me.
> 
> Check the `Header` - the `Compression` and `Encryption` fields

This must be at the `pgmoneta` side and not the `cli` side because we are not using translations at server's side only client side responses are translated

jesperpedersen:
The `conf get XYZ` is ok with translation -- its the `conf get` command w/o the XYZ

ashu3103:
> The `conf get XYZ` is ok with translation -- its the `conf get` command w/o the XYZ

Ohh, okay I'll fix that

ashu3103:
@jesperpedersen, PTAL

jesperpedersen:
It is weird,

```
-F text conf get
  Compression: none
  Encryption: none
```

```
-F raw
    "Compression": 2,
    "Encryption": 0,
```

```
-F json
    "Compression": "gzip",
    "Encryption": "none",
```

Where is `2` and `gzip` coming from ? I didn't specify a `-C` option...

ashu3103:
@jesperpedersen Check now, actually I swapped compression and out_format while calling `pgmoneta_management_request_conf_get` and `pgmoneta_management_request_conf_ls`, my bad

jesperpedersen:
@ashu3103 Yes, this looks good !

So, next is `set` which always `key` and `value` as arguments. Those get put into `ConfigKey` and `ConfigValue` in the `Request` part.

We could just output the input as the result - extracted from `Response` so

```
-F raw conf set entryption 0
0

-F text conf set entryption none
none

-F json conf set entryption none
{"encryption": "none"}
```

ashu3103:
@jesperpedersen PTAL

src/main.c
```diff
@@ -1149,6 +1149,66 @@ accept_mgt_cb(struct ev_loop* loop, struct ev_io* watcher, int revents)
...
+
+      pgmoneta_management_response_ok(NULL, client_fd, start_time, end_time, compression, encryption, payload);
+   }
+   else if (id == MANAGEMENT_CONF_GET)
+   {
+      pid = fork();
+      if (pid == -1)
+      {
+         pgmoneta_management_response_error(NULL, client_fd, server, MANAGEMENT_ERROR_STATUS_NOFORK, compression, encryption, payload);
+         pgmoneta_log_error("Status: No fork %s (%d)", server, MANAGEMENT_ERROR_STATUS_NOFORK);
```
jesperpedersen, line 1174: These need to be `_CONF_GET_` ones

src/main.c
```diff
@@ -1149,6 +1149,66 @@ accept_mgt_cb(struct ev_loop* loop, struct ev_io* watcher, int revents)
...
+         pgmoneta_conf_get(NULL, client_fd, compression, encryption, pyl);
+      }
+   }
+   else if (id == MANAGEMENT_CONF_SET)
+   {
+      pid = fork();
+      if (pid == -1)
+      {
+         pgmoneta_management_response_error(NULL, client_fd, server, MANAGEMENT_ERROR_STATUS_NOFORK, compression, encryption, payload);
+         pgmoneta_log_error("Status: No fork %s (%d)", server, MANAGEMENT_ERROR_STATUS_NOFORK);
```
jesperpedersen, line 1195: These need to be `_CONF_SET_` ones

src/main.c
```diff
@@ -1149,6 +1149,66 @@ accept_mgt_cb(struct ev_loop* loop, struct ev_io* watcher, int revents)
...
+         goto error;
+      }
+      else if (pid == 0)
+      {
+         struct json* pyl = NULL;
+         char* config_key = (char*)pgmoneta_json_get(request, MANAGEMENT_ARGUMENT_CONFIG_KEY);
+         char* config_value = (char*)pgmoneta_json_get(request, MANAGEMENT_ARGUMENT_CONFIG_VALUE);
```
jesperpedersen, line 1202: Extract those inside `pgmoneta_conf_set`

jesperpedersen (changes requested):

jesperpedersen:
```
pgmoneta-cli -c pgmoneta.conf -F json conf set unknown aes-256
```

need to return an error response

src/include/management.h
```diff
@@ -158,6 +161,8 @@ extern "C" {
 #define MANAGEMENT_ARGUMENT_VALID                 "Valid"
 #define MANAGEMENT_ARGUMENT_WAL                   "WAL"
 #define MANAGEMENT_ARGUMENT_WORKERS               "Workers"
+#define MANAGEMENT_ARGUMENT_CONFIG_KEY            "ConfigKey"
+#define MANAGEMENT_ARGUMENT_CONFIG_VALUE          "ConfigValue"
```
jesperpedersen, line 165: Sorting

jesperpedersen (changes requested):

src/libpgmoneta/configuration.c
```diff
@@ -2032,6 +2038,1038 @@ pgmoneta_reload_configuration(bool* restart)
...
+
+   pgmoneta_json_destroy(payload);
+
+   pgmoneta_disconnect(client_fd);
+
+   pgmoneta_stop_logging();
+
+   exit(0);
+error:
+   pgmoneta_log_info("Inside main error");
```
jesperpedersen, line 3062: This need to be an `pgmoneta_log_error` and need to have context of why it failed

jesperpedersen (changes requested):

ashu3103:
@jesperpedersen PTAL

Also are there any docs which provide a descriptive idea about the error code to the client, if not can we add that too in this or another PR or we can add the description of the error code in the outcome field of the response itself.

Thoughts?

jesperpedersen:
```
pgmoneta-cli conf get unknown
```

doesn't work

jesperpedersen:
>Also are there any docs which provide a descriptive idea about the error code to the client, if not can we add that too in this or another PR or we can add the description of the error code in the Outcome field of the response itself.

We need a chapter in the manuals about it, but it is separate from this pull request I think

ashu3103:
> ```
> pgmoneta-cli conf get unknown
> ```
> 
> doesn't work.
> 
> `-F text` and `-F raw` should give `Error`, `-F json` should give `{"Outcome": false}`

This can be only possible if we generate the error messages in `cli.c` because from main we will always get a response with whole json configuration and since we are filtering in `cli` we have to create error response explicitly.

jesperpedersen:
I passed in `unknown` as the key, so that shouldn't resolve to anything

ashu3103:
> I passed in `unknown` as the key, so that shouldn't resolve to anything

Yeah, so it will not be displaying anything except
`pgmoneta-cli: No connection to pgmoneta on /tmp/`

which is displayed because exit_code is 1.

jesperpedersen:
Yes, but if you can't resolve `unknown` then the user still need a valid response

ashu3103:
> Yes, but if you can't resolve `unknown` then the user still need a valid response

Understood, will be sticking with this format of output.
-F text and -F raw should give `Error`, -F json should give `{"Outcome": false}`

jesperpedersen:
You can keep the exit code of 1 as well

ashu3103:
> It says
> 
> ```
> Error
> pgmoneta-cli: No connection to pgmoneta on /tmp/
> ```
> 
> we don't need the last line -- only if there is no connection to pgmoneta

It will appear as long as we have `exit_code = 1` because of this section of code

```
   pgmoneta_disconnect(socket);

   if (configuration_path != NULL)
   {
      if (matched && exit_code != 0)
      {
         warnx("No connection to pgmoneta on %s", config->unix_socket_dir);
      }
   }
```

jesperpedersen:
Better move that warnx to another part of the code

jesperpedersen:
Log the warn when you know what it is... in this case we have a connection to pgmoneta - but the error was client side

jesperpedersen:
If it isn't clear - just remove the code for now

ashu3103:
> If it isn't clear - just remove the code for now

Okay so I'll comment it for now.

ashu3103:
> If it isn't clear - just remove the code for now

Done

jesperpedersen:
```
./pgmoneta-cli -c pgmoneta.conf -F XYZ conf set unknown aes
```

is still wrong

ashu3103:
> ```
> ./pgmoneta-cli -c pgmoneta.conf -F XYZ conf set unknown aes
> ```
> 
> is still wrong

I am getting:
```
Header: 
  ClientVersion: 0.15.0
  Command: conf set
  Compression: none
  Encryption: none
  Output: text
  Timestamp: 20241124234310
Outcome: 
  Error: 2204
  Status: false
Request: 
  ConfigKey: unknown
  ConfigValue: aes
Response: 
  ServerVersion: 0.15.0
```

For conf set where the errors are coming from main, I am displaying the error in `Outcome`

jesperpedersen:
```
./pgmoneta-cli -c pgmoneta.conf -F text conf set unknown aes
```

should give

```
Error
```

```
./pgmoneta-cli -c pgmoneta.conf -F raw conf set unknown aes
```

should give

```
Error
```

```
./pgmoneta-cli -c pgmoneta.conf -F json conf set unknown aes
```

should give

```
{"Outcome": false}
```

ashu3103:
> ```
> ./pgmoneta-cli -c pgmoneta.conf -F text conf set unknown aes
> ```
> 
> should give
> 
> ```
> Error
> ```
> 
> ```
> ./pgmoneta-cli -c pgmoneta.conf -F raw conf set unknown aes
> ```
> 
> should give
> 
> ```
> Error
> ```
> 
> ```
> ./pgmoneta-cli -c pgmoneta.conf -F json conf set unknown aes
> ```
> 
> should give
> 
> ```
> {"Outcome": false}
> ```

okay

jesperpedersen:
Rebased, and merged.

Issue #423, opened 2024-11-21
Metrics: Time and speed


jesperpedersen:
We should add Prometheus metrics about how much time is spent in each step of a backup - including how that translates into MB/s processing.

Metrics:
* Total
* Base backup
* Compression
* Encryption
* Linking
* Manifest
* Remote

so each "area" will translate into 2 Prometheus metrics.

Look at `wf_backup` @ https://github.com/pgmoneta/pgmoneta/blob/main/src/libpgmoneta/workflow.c#L208 for the different stages.

The information could be saved in `backup.info`

jesperpedersen:
@shikharish Can you take this one ?

shikharish:
Yes, please assign me.

jesperpedersen:
You could look into

```
#TYPE gauge
XYZ_total_lastime{} 8

#TYPE gauge
XYZ_total_mbs{} 87.6

#HELP Provide a histogram for the total backup time -- run-time only, data isn't persisted 
#TYPE histogram
XYZ_total{}
```

shikharish:
@jesperpedersen Currently we calculate time in seconds for any time-related metric. For better accuracy, I think we should use milliseconds, etc. How about I change all time metrics to double from int, like 1.25sec instead of 1sec?

shikharish:
That way the MB/s speed will be accurate too.

jesperpedersen:
@shikharish Yeah, feel free to change the data types - as long as it is reflected in the Prometheus interface... I don't think we need like 8 digits ... X.YYY is more than enough.

Maybe there are some best practices in Prometheus interfaces for this area...

PR #442, opened 2024-12-25
[#423] Add metrics for time and speed of each step of backup workflow

shikharish:
cc: @jesperpedersen

### DIFF
--- a/src/include/info.h
+++ b/src/include/info.h
@@ -40,27 +40,38 @@ extern "C" {
 /* system */
 #include <stdlib.h>
 
-#define INFO_PGMONETA_VERSION "PGMONETA_VERSION"
-#define INFO_BACKUP           "BACKUP"
-#define INFO_CHKPT_WALPOS     "CHKPT_WALPOS"
-#define INFO_COMMENTS         "COMMENTS"
-#define INFO_COMPRESSION      "COMPRESSION"
-#define INFO_ELAPSED          "ELAPSED"
-#define INFO_ENCRYPTION       "ENCRYPTION"
-#define INFO_END_TIMELINE     "END_TIMELINE"
-#define INFO_END_WALPOS       "END_WALPOS"
-#define INFO_EXTRA            "EXTRA"
-#define INFO_HASH_ALGORITHM   "HASH_ALGORITM"
-#define INFO_KEEP             "KEEP"
-#define INFO_LABEL            "LABEL"
-#define INFO_MAJOR_VERSION    "MAJOR_VERSION"
-#define INFO_MINOR_VERSION    "MINOR_VERSION"
-#define INFO_RESTORE          "RESTORE"
-#define INFO_START_TIMELINE   "START_TIMELINE"
-#define INFO_START_WALPOS     "START_WALPOS"
-#define INFO_STATUS           "STATUS"
-#define INFO_TABLESPACES      "TABLESPACES"
-#define INFO_WAL              "WAL"
+#define INFO_PGMONETA_VERSION          "PGMONETA_VERSION"
+#define INFO_BACKUP                    "BACKUP"
+#define INFO_CHKPT_WALPOS              "CHKPT_WALPOS"
+#define INFO_COMMENTS                  "COMMENTS"
+#define INFO_COMPRESSION               "COMPRESSION"
+#define INFO_ELAPSED                   "ELAPSED"
+#define INFO_BASEBACKUP_ELAPSED        "BASEBACKUP_ELAPSED"
+#define INFO_MANIFEST_ELAPSED          "MANIFEST_ELAPSED"
+#define INFO_COMPRESSION_ZSTD_ELAPSED  "COMPRESSION_ZSTD_ELAPSED"
+#define INFO_COMPRESSION_GZIP_ELAPSED  "COMPRESSION_GZIP_ELAPSED"
+#define INFO_COMPRESSION_BZIP2_ELAPSED "COMPRESSION_BZIP2_ELAPSED"
+#define INFO_COMPRESSION_LZ4_ELAPSED   "COMPRESSION_LZ4_ELAPSED"
+#define INFO_ENCRYPTION_ELAPSED        "ENCRYPTION_ELAPSED"
+#define INFO_LINKING_ELAPSED           "LINKING_ELAPSED"
+#define INFO_REMOTE_SSH_ELAPSED        "REMOTE_SSH_ELAPSED"
+#define INFO_REMOTE_S3_ELAPSED         "REMOTE_S3_ELAPSED"
+#define INFO_REMOTE_AZURE_ELAPSED      "REMOTE_AZURE_ELAPSED"
+#define INFO_ENCRYPTION                "ENCRYPTION"
+#define INFO_END_TIMELINE              "END_TIMELINE"
+#define INFO_END_WALPOS                "END_WALPOS"
+#define INFO_EXTRA                     "EXTRA"
+#define INFO_HASH_ALGORITHM            "HASH_ALGORITM"
+#define INFO_KEEP                      "KEEP"
+#define INFO_LABEL                     "LABEL"
+#define INFO_MAJOR_VERSION             "MAJOR_VERSION"
+#define INFO_MINOR_VERSION             "MINOR_VERSION"
+#define INFO_RESTORE                   "RESTORE"
+#define INFO_START_TIMELINE            "START_TIMELINE"
+#define INFO_START_WALPOS              "START_WALPOS"
+#define INFO_STATUS                    "STATUS"
+#define INFO_TABLESPACES               "TABLESPACES"
+#define INFO_WAL                       "WAL"
 
 #define VALID_UNKNOWN -1
 #define VALID_FALSE    0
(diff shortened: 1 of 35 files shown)


src/include/info.h
```diff
@@ -45,7 +45,7 @@ extern "C" {
 #define INFO_CHKPT_WALPOS     "CHKPT_WALPOS"
 #define INFO_COMMENTS         "COMMENTS"
 #define INFO_COMPRESSION      "COMPRESSION"
-#define INFO_ELAPSED          "ELAPSED"
+#define INFO_TOTAL_ELAPSED          "TOTAL_ELAPSED"
```
jesperpedersen, line 48: Just expose all `XYZ_ELAPSED` - and keep `ELAPSED` as total elapsed

src/libpgmoneta/info.c
```diff
@@ -840,9 +926,36 @@ pgmoneta_get_backup_file(char* fn, struct backup** backup)
...
          }
-         else if (!strcmp(INFO_ELAPSED, &key[0]))
+         else if (strstr(&key[0], "ELAPSED"))
          {
-            bck->elapsed_time = atoi(&value[0]);
+            if (strstr(&key[0], "TOTAL"))
+            {
+               bck->total_elapsed_time = atof(&value[0]);
+            }
+            else if (strstr(&key[0], WORKFLOW_NAME_BASEBACKUP))
```
jesperpedersen, line 935: Make them like `MANIFEST_ELAPSED` and so on

jesperpedersen (changes requested):

jesperpedersen:
Other than that it looks good

shikharish:
@jesperpedersen Should I add different names for different compression types?
And different names for different remote types?
The difference between the two is that there can be multiple remote types. And I noticed that the code will not work right in case of multiple remote types. I think I would have to add `remote_azure_elapsed_time`, `remote_s3_elapsed_time`, etc. fields in `struct backup` for all remote types. How should I handle this?

shikharish:
@jesperpedersen updated the patch. I think this is better.  Exposing all `INFO_XYZ` is kind of redundant when I have already defined `WORKFLOW_NAME_XYZ`.

src/include/workflow.h
```diff
@@ -47,6 +47,18 @@ extern "C" {
 #define WORKFLOW_TYPE_WAL_SHIPPING  5
 #define WORKFLOW_TYPE_VERIFY        6
 
+#define WORKFLOW_NAME_BASEBACKUP "BASEBACKUP"
```
jesperpedersen, line 50: Remove

src/include/workflow.h
```diff
@@ -80,6 +92,7 @@ typedef int (* teardown)(int, char*, struct deque*);
  */
 struct workflow
 {
+   const char* name;      /**< The name of the workflow */
```
jesperpedersen, line 95: Remove

src/libpgmoneta/backup.c
```diff
@@ -130,12 +137,25 @@ pgmoneta_backup(int client_fd, int server, uint8_t compression, uint8_t encrypti
    current = workflow;
    while (current != NULL)
    {
+      if (current->name)
```
jesperpedersen, line 140: Just calculate and set in the `wf_` - you can use a process local variable with init in `setup()` and set in `teardown()`

jesperpedersen (changes requested):

jesperpedersen:
I have to check the Prometheus changes as there seems to be a lot of unrelated changes -- otherwise it looks good

src/libpgmoneta/backup.c
shikharish: @jesperpedersen Wouldn't it be better if we add the framework to measure the time of _any_ workflow? That is why I decided to add the `name` field...
Also, if I measure the time between `setup()` and `teardown()`, that will be inaccurate right? Because all workflows first run `setup()` then `execute()` and then `teardown()`.
May I also ask the thought behind this design choice. Why is not that one workflow completes first(setup->execute->teardown) and then only the next workflow starts?

jesperpedersen:
@shikharish We have a lot on our wish-list for a new work flow architecture, planned for 0.17, unless you want to try and tackle after this task.

Timing is def part of that wish-list

src/libpgmoneta/backup.c
jesperpedersen: For now, we just need the timing part - having a general solution integrated is planned

shikharish: Alright.

And I think I should only measure time of `execute()`, for reason I stated above ?

jesperpedersen: Yes, that is ok

jesperpedersen:
You need to squash, and rebase

src/include/info.h
```diff
@@ -46,6 +46,17 @@ extern "C" {
 #define INFO_COMMENTS         "COMMENTS"
 #define INFO_COMPRESSION      "COMPRESSION"
 #define INFO_ELAPSED          "ELAPSED"
+#define INFO_BASEBACKUP_ELAPSED         "BASEBACKUP_ELAPSED"
```
jesperpedersen, line 49: Make sure they are aligned

src/libpgmoneta/se_azure.c
```diff
@@ -27,6 +27,7 @@
  */
 
 /* pgmoneta */
+#include "workflow.h"
```
jesperpedersen, line 30: Remove

src/libpgmoneta/se_local.c
```diff
@@ -27,6 +27,7 @@
  */
 
 /* pgmoneta */
+#include "utils.h"
```
jesperpedersen, line 30: its `#include <utils.h>` and sorted

src/libpgmoneta/se_s3.c
```diff
@@ -27,6 +27,7 @@
  */
 
 /* pgmoneta */
+#include "workflow.h"
```
jesperpedersen, line 30: Remove

src/libpgmoneta/se_ssh.c
```diff
@@ -29,6 +29,7 @@
  */
 
 /* pgmoneta */
+#include "workflow.h"
```
jesperpedersen, line 32: Remove

src/libpgmoneta/wf_backup.c
```diff
@@ -356,6 +360,8 @@ basebackup_execute(int server, char* identifier, struct deque* nodes)
    pgmoneta_update_info_unsigned_long(backup_base, INFO_START_TIMELINE, start_timeline);
    pgmoneta_update_info_unsigned_long(backup_base, INFO_END_TIMELINE, end_timeline);
    pgmoneta_update_info_unsigned_long(backup_base, INFO_HASH_ALGORITHM, hash);
+   pgmoneta_update_info_double(backup_base, INFO_BASEBACKUP_ELAPSED, total_seconds);
```
jesperpedersen, line 363: `BASEBACKUP` isn't `total_seconds`

src/libpgmoneta/wf_encryption.c
```diff
@@ -176,16 +177,20 @@ encryption_execute(int server, char* identifier, struct deque* nodes)
...
+   minutes = ((int)total_seconds % 3600) / 60;
+   seconds = (int)total_seconds % 60 + (total_seconds - ((long)total_seconds));
 
    memset(&elapsed[0], 0, sizeof(elapsed));
-   sprintf(&elapsed[0], "%02i:%02i:%02i", hours, minutes, seconds);
+   sprintf(&elapsed[0], "%02i:%02i:%.4f", hours, minutes, seconds);
 
    pgmoneta_log_debug("Encryption: %s/%s (Elapsed: %s)", config->servers[server].name, identifier, &elapsed[0]);
 
+   pgmoneta_update_info_double(backup_base, INFO_ENCRYPTION_ELAPSED, total_seconds);
```
jesperpedersen, line 192: Same - use the explicit vaiable names

src/libpgmoneta/wf_link.c
```diff
@@ -160,15 +163,20 @@ link_execute(int server, char* identifier, struct deque* nodes)
...
+         seconds = (int)total_seconds % 60 + (total_seconds - ((long)total_seconds));
 
          memset(&elapsed[0], 0, sizeof(elapsed));
-         sprintf(&elapsed[0], "%02i:%02i:%02i", hours, minutes, seconds);
+         sprintf(&elapsed[0], "%02i:%02i:%.4f", hours, minutes, seconds);
 
          pgmoneta_log_debug("Link: %s/%s (Elapsed: %s)", config->servers[server].name, identifier, &elapsed[0]);
+
+         backup_base = (char*)pgmoneta_deque_get(nodes, NODE_BACKUP_BASE);
+         pgmoneta_update_info_double(backup_base, INFO_LINKING_ELAPSED, total_seconds);
```
jesperpedersen, line 179: So, call them what you do in `struct backup` ...

src/libpgmoneta/wf_lz4.c
```diff
@@ -27,6 +27,7 @@
  */
 
 /* pgmoneta */
+#include "info.h"
```
jesperpedersen, line 30: `#include <info.h>` - sorted

src/libpgmoneta/wf_manifest.c
```diff
@@ -25,10 +25,12 @@
  * TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
+#include "info.h"
```
jesperpedersen, line 28: `#include <info.h>` - sorted

src/libpgmoneta/wf_manifest.c
```diff
@@ -25,10 +25,12 @@
...
 #include <csv.h>
 #include <json.h>
 #include <logging.h>
 #include <manifest.h>
+#include <time.h>
```
jesperpedersen, line 33: Move to system includes

jesperpedersen (changes requested):

jesperpedersen:
Why so many changes in `prometheus.c` ?

shikharish:
> Why so many changes in `prometheus.c` ?

We need 2 metrics each for each part of the backup workflow. And I added different metrics for each compression type and remote type. Only the used compression and remote types will show any valid value, the unused ones will be zero.

jesperpedersen:
Merged.

Issue #444, opened 2024-12-30
`compress` command doesn't acknowledge non-default compression type(gz, lz4, bzip2)


shikharish:
The `compress` command seems to ignore the `-C` flag.

I ran pgmoneta in offline mode:
```
❯ ./pgmoneta -c pgmoneta.conf -u pgmoneta_users.conf --offline
```
 And compress command:
```
❯ ./pgmoneta-cli -c pgmoneta.conf compress text -C gz
```

Info says the compression is gzip, but the compressed file is .zstd
```
Header: 
  ClientVersion: 0.16.0
  Command: compress
  Compression: gzip
  Encryption: none
  Output: text
  Timestamp: 20241231024353
Outcome: 
  Status: true
  Time: 00:00:0.0044
Request: 
  SourceFile: text
Response: 
  DestinationFile: text.zstd
  ServerVersion: 0.16.0
```

Logs:
```
2024-12-31 02:43:53 DEBUG main.c:839 Management 17: {
  "Header": {
    "ClientVersion": "0.16.0",
    "Command": 17,
    "Compression": 1,
    "Encryption": 0,
    "Output": 0,
    "Timestamp": "20241231024353"
  },
  "Request": {
    "SourceFile": "text"
  }
}
2024-12-31 02:43:53 INFO  zstandard_compression.c:658 ZSTD: text (Elapsed: 00:00:0.0044)
```

shikharish:
I am looking into it.

jesperpedersen:
What is `text` ? Did you mean `-F text` ?

shikharish:
It's the filename

shikharish:
@jesperpedersen I believe there are two types of compression types involved when we run the `compress`(or similar) command: one used internally to send and receive json messages, the other passed by the `-C` flag used to compress the file.

I believe the two should be unrelated? The former should be set through `pgmoneta.conf` right?

jesperpedersen:
Yes, internal usage is based on the `pgmoneta.conf` setting.

For the CLI commands you will use the `-C` option, like

```
pgmoneta-cli -C zstd decompress backups/primary/wal/000000010000000000000014.zstd
pgmoneta-cli -C zstd compress backups/primary/wal/000000010000000000000014
```

Same for `encrypt`/`decrypt`

jesperpedersen:
You can "guess" the algorithm based on the file extension such that `-C` is optional

shikharish:
[...]

PR #445, opened 2025-01-01
[#444] Correct use of compression and encryption types

shikharish:
cc: @jesperpedersen

### DIFF
--- a/src/cli.c
+++ b/src/cli.c
@@ -117,10 +117,10 @@ static int reset(SSL* ssl, int socket, uint8_t compression, uint8_t encryption, 
 static int reload(SSL* ssl, int socket, uint8_t compression, uint8_t encryption, int32_t output_format);
 static int retain(SSL* ssl, int socket, char* server, char* backup_id, uint8_t compression, uint8_t encryption, int32_t output_format);
 static int expunge(SSL* ssl, int socket, char* server, char* backup_id, uint8_t compression, uint8_t encryption, int32_t output_format);
-static int decrypt_data(SSL* ssl, int socket, char* path, uint8_t compression, uint8_t encryption, int32_t output_format);
-static int encrypt_data(SSL* ssl, int socket, char* path, uint8_t compression, uint8_t encryption, int32_t output_format);
-static int decompress_data(SSL* ssl, int socket, char* path, uint8_t compression, uint8_t encryption, int32_t output_format);
-static int compress_data(SSL* ssl, int socket, char* path, uint8_t compression, uint8_t encryption, int32_t output_format);
+static int decrypt_data(SSL* ssl, int socket, char* path, uint8_t compression, uint8_t encryption, int32_t encryption_type, int32_t output_format);
+static int encrypt_data(SSL* ssl, int socket, char* path, uint8_t compression, uint8_t encryption, int32_t encryption_type, int32_t output_format);
+static int decompress_data(SSL* ssl, int socket, char* path, uint8_t compression, uint8_t encryption, int32_t compression_type, int32_t output_format);
+static int compress_data(SSL* ssl, int socket, char* path, uint8_t compression, uint8_t encryption, int32_t compression_type, int32_t output_format);
 static int info(SSL* ssl, int socket, char* server, char* backup, uint8_t compression, uint8_t encryption, int32_t output_format);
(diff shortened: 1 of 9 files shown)


shikharish:
We can probably have better variable names than `compression` and `compression_type`...

jesperpedersen:
Well, you can change compression type between backups.

So,

```
[pgmoneta]
compression = gzip
```

to

```
[pgmoneta]
compression = zstd
```

So, we can't assume that `config->compression` is the same as on-disk.

Try with commands to explain what you are seeing

shikharish:
@jesperpedersen  Don't understand what you mean to say...Whenever we run `pgmoneta-cli -c pgmoneta.conf -c <command>`, the compression type will get updated in the config...

jesperpedersen:
Yes, but it shouldn't.

`compress`, `decompress`, `encrypt` and `decrypt` just operate on the files on the disk based on the `-C`/`-E` options. They don't change the configuration of `pgmoneta.conf`

jesperpedersen:
.. that is what `conf set` is for ..

shikharish:
Yes, I understand that. I meant that if the `compression` is changed in pgmoneta.conf then the compression type for internal wire protocol will be updated.

shikharish:
> Yes, I understand that. I meant that if the `compression` is changed in `pgmoneta.conf` then the compression type for internal wire protocol will be updated.

Is that not the intended behavior?

jesperpedersen:
The management protocol is only controlled by `-C` and `-E` right now.

We could change it to if `-C` and `-E` are not set then it will use the values from `pgmoneta.conf` - but `none` need to work for testing

jesperpedersen:
Just to be clear, you can do

```
cli -C bz2 -E aes decompress wal/000000000001.zstd
```

The `.zstd` is server side in pgmoneta

shikharish:
With my patch, this is the behavior:
- `compress`: For wire protocol, it uses the `compression` from `pgmoneta.conf` and for compression of the file, it uses the type specified by `-C`
- `decompress`: For wire protocol, it uses the `compression` from `pgmoneta.conf` and for decompression of the file, it ignore the `-C` flag, and guesses the type from the file extension
- `encrypt`: For wire protocol, it uses the `encryption` from `pgmoneta.conf` and for compression of the file, it uses the type specified by `-E`
- `decrypt`: For wire protocol, it uses the `encryption` from `pgmoneta.conf` and for decompression of the file, it uses the type specified by `-E`

Is this wrong, am I missing something?

jesperpedersen:
`-C` and `-E` are only for the wire protocol - nothing on the server side.

Only server side is reached - if there is a server running - then it need to resolve the file type independant of `-C` and `-E`.

I see a `#define MANAGEMENT_COMPRESSION_UNDEFINED -1` and in the CLI load from `pgmoneta.conf` if `-C` isn't set

shikharish:
Ah, I get it now. I misunderstood the whole workflow. Then I guess there was no bug in the first place...

The behavior I mentioned in the original issue is alright?
> I ran pgmoneta in offline mode:
> 
> ```
> ❯ ./pgmoneta -c pgmoneta.conf -u pgmoneta_users.conf --offline
> ```
> 
> And compress command:
> 
> ```
> ❯ ./pgmoneta-cli -c pgmoneta.conf compress text -C gz
> ```
> 
> Info says the compression is gzip, but the compressed file is .zstd
> 
> ```
> Header: 
>   ClientVersion: 0.16.0
>   Command: compress
>   Compression: gzip
>   Encryption: none
>   Output: text
>   Timestamp: 20241231024353
> Outcome: 
>   Status: true
>   Time: 00:00:0.0044
> Request: 
>   SourceFile: text
> Response: 
>   DestinationFile: text.zstd
>   ServerVersion: 0.16.0
> ```

jesperpedersen:
Yes, `compress` should take the setting from `pgmoneta.conf`, so the above looks correct.

I think this issue is more about if we want the `_XYZ_UNDEFINED -1` flags, and override with the value from `pgmoneta.conf`. I have no strong opinion on that.

So, feel free to close the issues, if you are happy - or do we need documentation on this ?

shikharish:
I will close this. Will also update `CLI.md` more clarity.

Issue #456, opened 2025-01-12
Support `-F` flag for `pgmoneta-admin`


shikharish:
We would like to have full JSON/text output for all `pgmoneta-admin` commands.

shikharish:
I will take this first.

PR #458, opened 2025-01-14
[#456] Add -F/--format flag for pgmoneta-admin

shikharish:
cc: @jesperpedersen

### DIFF
--- a/src/admin.c
+++ b/src/admin.c
@@ -28,10 +28,13 @@
 
 /* pgmoneta */
 #include <pgmoneta.h>
-#include <logging.h>
-#include <security.h>
 #include <aes.h>
+#include <json.h>
+#include <logging.h>
+#include <management.h>
+#include <security.h>
 #include <utils.h>
+#include <value.h>
 
 /* system */
 #include <ctype.h>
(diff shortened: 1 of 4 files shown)


src/admin.c
```diff
@@ -131,21 +129,22 @@ usage(void)
...
    printf("\n");
    printf("Options:\n");
-   printf("  -f, --file FILE         Set the path to a user file\n");
-   printf("  -U, --user USER         Set the user name\n");
-   printf("  -P, --password PASSWORD Set the password for the user\n");
-   printf("  -g, --generate          Generate a password\n");
-   printf("  -l, --length            Password length\n");
-   printf("  -V, --version           Display version information\n");
-   printf("  -?, --help              Display help\n");
+   printf("  -f, --file FILE                 Set the path to a user file\n");
```
jesperpedersen, line 132: We only need 2 spaces after the longest argument

jesperpedersen (changes requested):

jesperpedersen:
We need a `Response` section so

```
Outcome: 
  Status: true
  Time: 00:00:0.0013
Response:
  Users:
     User: repl1
     User: repl2
```

Each command should list the users in the file

shikharish:
> We need a `Response` section so
> 
> ```
> Outcome: 
>   Status: true
>   Time: 00:00:0.0013
> Response:
>   Users:
>     User: repl1
>     User: repl2
> ```
> 
> Each command should list the users in the file

Each command should list _all_ users?

shikharish:
Outcome will only contain `Status` and `Time`. And response will contain the actual response of the command?
For ex, what should the output of `user del` look like?

jesperpedersen:
I think it should display the current users in the file, so `pgmoneta_users.conf`

```
repl1:...
repl2:...
```

`-U repl2 user del` will give

```
Response:
  Users:
    User: repl1
```

Issue #434, opened 2024-12-14
walinfo: Support all WAL sizes


jesperpedersen:
Get the file size - the pre-allocated size, not the real size - and use that in calculations.

Fall-back to `DEFAULT_WAL_SEGZ_BYTES`

jesperpedersen:
F.ex.

```
initdb -k --wal-segsize=1024 /tmp/16
```

PR #459, opened 2025-01-14
[#434] Get the segment size as an optional argument

sh-soltanpour:


jesperpedersen:
Can we detect the size ? Even it they are `.zstd.aes` ?

Or should we go with the user supplied value for now ?

jesperpedersen:
`fseek(f, 0, SEEK_END);` or `struct stat st; stat(filename, &st); size = st.st_size;` ?

sh-soltanpour:
Yeah I think this is also possible, for encrypted/compressed files also we can get it after decryption/decompression.
I thought you wanted to get it explicitly from the user based on [this example](https://github.com/pgmoneta/pgmoneta/issues/434#issuecomment-2543645402)

jesperpedersen:
We will get to the C API later

jesperpedersen:
We can assume that the WAL segment sizes are the same within a directory

jesperpedersen:
Merged.

PR #488, opened 2025-02-22
admin: support password through environment

tacerus:
Passing the password through the "--password" command line argument is potentially insecure, as the value would be exposed in a process listing or shell history. Read an environment variable as an alternative.


jesperpedersen:
Merged.

Issue #499, opened 2025-03-04
HTTPS/metrics


shikharish:
Add support for serving prometheus metrics from a HTTPS endpoint.

PR #500, opened 2025-03-04
[#499] Add HTTPS support for prometheus metrics

shikharish:


jesperpedersen:
@shikharish You need some documentation for this - update the tutorial too

jesperpedersen:
Add a basic self-signed certificate section that you use

jesperpedersen:
You should use OpenSSL instead of mkcert.

You need to update the manuals as well.

jesperpedersen:
And, the man pages... Find an existing key and then `git grep` is your friend

shikharish:
@jesperpedersen Pls have a look.

src/include/pgmoneta.h
```diff
@@ -248,6 +248,8 @@ struct server
    char tls_cert_file[MISC_LENGTH];         /**< TLS certificate path */
    char tls_key_file[MISC_LENGTH];          /**< TLS key path */
    char tls_ca_file[MISC_LENGTH];           /**< TLS CA certificate path */
+   char prometheus_cert_file[MISC_LENGTH];         /**< TLS certificate path */
+   char prometheus_key_file[MISC_LENGTH];          /**< TLS key path */
```
jesperpedersen, line 252: Align the comments with the rest

src/include/security.h
```diff
@@ -199,6 +199,26 @@ pgmoneta_close_ssl(SSL* ssl);
 int
 pgmoneta_get_hash_algorithm(char* algorithm);
 
+/**
+ * Create a SSL context
+ * @param client Is this a client context
+ * @param ctx The SSL context
+ * @return 0 upon success, otherwise 1
+ */
+int  pgmoneta_create_ssl_ctx(bool client, SSL_CTX** ctx);
```
jesperpedersen, line 208: uncrustify - or align code.

Return value is on its own line

src/include/security.h
```diff
@@ -199,6 +199,26 @@ pgmoneta_close_ssl(SSL* ssl);
...
+ * Create a SSL server
+ * @param ctx The SSL context
+ * @param socket The socket
+ * @param key The key file path
+ * @param cert The certificate file path
+ * @param ca The ca file path
+ * @param ssl The SSL structure
+ * @return 0 upon success, otherwise 1
+ */
+int  pgmoneta_create_ssl_server(SSL_CTX* ctx, int socket, char* key, char* cert, char* ca, SSL** ssl);
```
jesperpedersen, line 220: Same

src/include/pgmoneta.h
jesperpedersen: We should use `MAX_PATH`

src/include/pgmoneta.h
```diff
@@ -366,6 +368,9 @@ struct configuration
    char tls_key_file[MISC_LENGTH];  /**< TLS key path */
    char tls_ca_file[MISC_LENGTH];   /**< TLS CA certificate path */
 
+   char prometheus_cert_file[MISC_LENGTH]; /**< TLS certificate path */
+   char prometheus_key_file[MISC_LENGTH];  /**< TLS key path */
```
jesperpedersen, line 372: `MAX_PATH`

src/libpgmoneta/configuration.c
```diff
@@ -643,6 +643,37 @@ pgmoneta_read_configuration(void* shm, char* filename)
                      unknown = true;
                   }
                }
+               else if (!strcmp(key, "prometheus_cert_file"))
+               {
+                  if (!strcmp(section, "pgmoneta"))
+                  {
+                     max = strlen(value);
+                     if (max > MISC_LENGTH - 1)
```
jesperpedersen, line 651: Change for `MAX_PATH`

jesperpedersen (changes requested):

src/include/pgmoneta.h
jesperpedersen: You can change all `tls_X` to `MAX_PATH` too

src/include/security.h
shikharish: weirdly uncrustify doesnt do it :/

jesperpedersen:
You don't need 

```
| prometheus_cert_file | | String | No | Certificate file for TLS for Prometheus metrics. This file must be owned by either the user running pgmoneta or root. |
| prometheus_key_file | | String | No | Private key file for TLS for Prometheus metrics. This file must be owned by either the user running pgmoneta or root. Additionally permissions must be at least `0640` when owned by root or `0600` otherwise. |
```

in the server section - we are protecting `metrics` - so you can call them

```
| metrics_cert_file | | String | No | Certificate file for TLS for the metrics. This file must be owned by either the user running pgmoneta or root. |
| metrics_key_file | | String | No | Private key file for TLS for the metrics. This file must be owned by either the user running pgmoneta or root. Additionally permissions must be at least `0640` when owned by root or `0600` otherwise. |
```

in top-level.

Also, there are memory leaks - so you need to run with clang w/ Debug

jesperpedersen:
Try and do an incorrect configuration - maybe it is on my end but nothing is working.

Check if each file exists / can be loaded - and fix all the leaks in the error paths

jesperpedersen:
`http://` need to redirect to `https://` is configured

jesperpedersen:
You should do some basic checks in the configuration validation - and `FATAL` on them

shikharish:
We also need to install the certificate on the system. I forgot to add instructions for that in the tutorial, will do that.

This is the current behaviour: 
- if no `metrics_cert_file` is provided, only http works 
- if `metrics_cert_file` is provided, only https work

http redirect to https is not configured. I think I would have to send a 301 request from the http endpoint to configure that. I will add that today.

jesperpedersen:
I mean you should also add some `pgmoneta_exists()` checks to `pgmoneta_validate_configuration` at least with the rules. So, do the combinations, cert file with no key file, ...

jesperpedersen:
Yes

shikharish:
@jesperpedersen I have removed the logic of creating a new SSL context for every request. We only need one context for all requests. 

For redirection, I am checking the first few bytes to see if its an SSL handshake or not. If not it is redirecting to the https page.

Have a look.

jesperpedersen:
See CI

shikharish:
Fixed

jesperpedersen:
Otherwise it need to be an `ERROR` and only enable `http://`

shikharish:
Have made changes for better error checking.

jesperpedersen:
Yes, followed the tutorial, and

```
metrics_cert_file = ./localhost.crt
metrics_key_file = ./localhost.crt
```

jesperpedersen:
```
2025-03-12 13:32:33 ERROR prometheus.c:110 Failed to read message
```

shikharish:
@jesperpedersen Can you please test this patch. I have added instructions for installing the certificate in the docs.
And if the certificate file is not found then we fall back to HTTP. 
I have tested this patch on my Fedora and Ubuntu VM.

shikharish:
At the end I decided to create new context for each request as it was very difficult managing memory with the current structure the other way.

I think I will make another patch after this improving the way metrics are handled. I dont understand why we are forking for every new request. A better way would be thread pooling. What do you think?

shikharish:
Oh. Why not?

jesperpedersen:
What happens if there is a `SIGx` in a thread ?

jesperpedersen:
I'm getting

```
2025-03-18 17:53:36 ERROR prometheus.c:99 Failed to accept SSL connection
```

when I doing access, and redirects...

shikharish:
how are you testing? can you test via wget once?
```
wget "http://localhost:5001"
```

shikharish:
this is what you should be getting:
```
--2025-03-19 03:31:33--  http://localhost:5001/
Resolving localhost (localhost)... 127.0.0.1
Connecting to localhost (localhost)|127.0.0.1|:5001... connected.
HTTP request sent, awaiting response... 302 Found
Location: https://localhost:5001/ [following]
--2025-03-19 03:31:33--  https://localhost:5001/
Connecting to localhost (localhost)|127.0.0.1|:5001... connected.
HTTP request sent, awaiting response... 200 OK
Length: unspecified [text/html]
Saving to: ‘index.html’

index.html                          [ <=>                                                   ]  19.69K  --.-KB/s    in 0.001s

2025-03-19 03:31:33 (25.1 MB/s) - ‘index.html’ saved [20165]
```

jesperpedersen:
Just using firefox, and force it to redirect on certain URLs

shikharish:
Does wget work? 
Is the a firefox specific issue? (https://bugzilla.mozilla.org/show_bug.cgi?id=1034124)

jesperpedersen:
No, wget fails too

jesperpedersen:
```
curl -v http://localhost:5001
* Host localhost:5001 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:5001...
* Connected to localhost (::1) port 5001
> GET / HTTP/1.1
> Host: localhost:5001
> User-Agent: curl/8.9.1
> Accept: */*
> 
* Request completely sent off
< HTTP/1.1 302 Found
< Location: https://localhost:5001/
< Date: Tue Mar 18 19:17:05 2025
< Content-Length: 0
< Connection: close
< 
* shutting down connection #0
*
```

seems like `301` is better

shikharish:
updated.

shikharish:
does `curl -v -L "http://localhost:5001"` redirect?

jesperpedersen:
```
2025-03-18 20:01:56 ERROR prometheus.c:99 Failed to accept SSL connection
```

```
curl -v -L http://localhost:5001
* Host localhost:5001 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:5001...
* Connected to localhost (::1) port 5001
> GET / HTTP/1.1
> Host: localhost:5001
> User-Agent: curl/8.9.1
> Accept: */*
> 
* Request completely sent off
< HTTP/1.1 301 Moved Permanently
< Location: https://localhost:5001/
< Date: Tue Mar 18 20:01:56 2025
< Content-Length: 0
< Connection: close
< 
* shutting down connection #0
* Clear auth, redirects scheme from http to https
* Issue another request to this URL: 'https://localhost:5001/'
* Hostname localhost was found in DNS cache
*   Trying [::1]:5001...
* Connected to localhost (::1) port 5001
* ALPN: curl offers h2,http/1.1
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
*  CAfile: /etc/pki/tls/certs/ca-bundle.crt
*  CApath: none
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
* TLSv1.3 (IN), TLS handshake, Certificate (11):
* TLSv1.3 (OUT), TLS alert, unknown CA (560):
* SSL certificate problem: self-signed certificate
* closing connection #1
curl: (60) SSL certificate problem: self-signed certificate
More details here: https://curl.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the webpage mentioned above.
```

shikharish:
And you can find the certificate in /etc/ssl/certs ?
```
[shikhar@lima-fedora src]$ ls /etc/ssl/certs | grep local
localhost.pem
```

jesperpedersen:
That isn't part of the tutorial -- maybe you want a `metrics_root_file`...

shikharish:
I simply followed the steps of the tutorial...

Generate the private key and self-signed certificate (valid for 3650 days)
```
openssl req -x509 -nodes -days 3650 -newkey rsa:2048 \
  -keyout localhost.key -out localhost.crt \
  -subj "/CN=localhost" \
  -addext "subjectAltName=DNS:localhost,IP:127.0.0.1"
```

Now install the certificate on your system:
- For Fedora:
```
# Create directory for certificates if it doesn't exist
sudo mkdir -p /etc/pki/ca-trust/source/anchors/

# Copy your certificate
sudo cp localhost.crt /etc/pki/ca-trust/source/anchors/

# Update CA certificates store
sudo update-ca-trust
```
 
 Created a new vm and completed the above steps and it works

jesperpedersen:
We can't touch `/etc/pki/` in a test setup - so this need to be fixed.

If we are talking about real certificates in a production environment, then we can "lower" the requirements for creating the context

shikharish:
@jesperpedersen But we have to install the certificate in the system right? What other way than this?

jesperpedersen:
Take a look at how pgagroal does it -- search `git grep ca_file`

jesperpedersen:
@shikharish Can you rebase, and fix the above ?

jesperpedersen:
@shikharish Still ok ?

jesperpedersen:
@shikharish Where are you on this ?

shikharish:
@jesperpedersen  For a browser to access the page at `https://localhost...` I think we would need to install the certificates....
I am doing something very similar to pgagroal's implementation. But I do not understand how a browser would trust the certificate if it is not installed in the system.

jesperpedersen:
@shikharish You can test with https://curl.se/docs/sslcerts.html

jesperpedersen:
@shikharish Ok, send a PTAL when you have that and `metrics_ca_file`

shikharish:
@jesperpedersen PTAL

src/main.c
```diff
@@ -1693,10 +1695,25 @@ accept_metrics_cb(struct ev_loop* loop, struct ev_io* watcher, int revents)
    {
       ev_loop_fork(loop);
       shutdown_ports();
+      if (strlen(config->metrics_cert_file) > 0 && strlen(config->metrics_key_file) > 0 && strlen(config->metrics_ca_file))
```
jesperpedersen, line 1698: But CA is optional if it is installed on the system, right ?

jesperpedersen (changes requested):

src/main.c
shikharish: Right, this check is not necessary.

shikharish:
are you running this command?
```
curl -v -L "https://localhost:5001" --cacert ca.crt --cert server.crt --key server.key
```

jesperpedersen:
This was Firefox... curl works

shikharish:
@jesperpedersen Yes, getting the same error on firefox. I believe the complications are because we have enabled client certificate authentication. I am not really sure if we _need_ it.
I solved the firefox issue by first export the certificate in p12 format and then manually import the certificate in firefox. I can add instructions for that. I would probably have to add different instructions for different browsers.

jesperpedersen:
Sure, you can add those instructions.

I'm more worried about the 

> 2025-04-10 13:40:50 ERROR prometheus.c:111 Failed to read message

Maybe add TRACE logging when loading the various parts, and we have to figure out why these ERRORs happen

jesperpedersen:
@shikharish Yes ?

shikharish:
Testing rigoursly, then i will push. Works well on mac, will test on a linux machine today and push.

shikharish:
@jesperpedersen PTAL

shikharish:
I want to point out that we have to provide the CA file to pgmoneta.conf if we are signing our certificates with it. I think that is obvious.

jesperpedersen:
Merged.

jesperpedersen:
@shikharish Please, port to pgexporter and pgagroal

PR #506, opened 2025-03-08
Use GET_ENV macro for environment variable

### LINKED ISSUE #502: MacOS Build Fails: secure_getenv() Is Not Supported
I encountered a build failure while compiling pgmoneta on macOS. The issue is caused by the use of secure_getenv(), which is not available on macOS as it is a Linux-specific function.
This function was introduced in commit #488.

Image

Annupamaa:
-Implemented a macro that automatically chooses getenv on Apple platforms and secure_getenv on other systems to ensure macOS compatibility
-fixes #502

### DIFF
--- a/src/admin.c
+++ b/src/admin.c
@@ -394,6 +394,13 @@ master_key(char* password, bool generate_pwd, int pwd_length, int32_t output_for
       goto error;
    }
 
+   #if defined(HAVE_DARWIN) || defined(HAVE_OSX)
+      #define GET_ENV(name) getenv(name)
+   #else
+      #define GET_ENV(name) secure_getenv(name)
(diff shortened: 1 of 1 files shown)


jesperpedersen:
Can you check `src/CMakeLists.txt` and see if we should use `HAVE_DARWIN` or `HAVE_OSX` ?

jesperpedersen:
Merged.

jesperpedersen:
Can you port the latest `admin.c` to pgagroal and pgexporter ?

Annupamaa:
Yes, sure.

jesperpedersen:
@Annupamaa Check Mac for CI too, of course

Issue #474, opened 2025-02-14
tsclient impl


ashu3103:
## Task

Implement a new client `tsclient`, to handle postgres and pgmoneta operations while running tests in test environment.

jesperpedersen:
No, there is no CLI tool in this.

It is a test suite API that executes the commands, and look at the entire response - especially `Outcome` of course.\

So, a test suite / case will create a tsclient, execute, get the response and then close it.

The test suite should use the same configuration that pgmoneta was started with

jesperpedersen:
Maybe look at https://github.com/PGBuildFarm/client-code for ideas

jesperpedersen:
Yes, it is just an API that make the calls to the server

jesperpedersen:
Yes.

@Arnabdaz is working on the generator

ashu3103:
Hi @jesperpedersen,

I was thinking of initializing the tsclient in runner (not in each testcase) and connect to the main server in the initialisation itself (so that we don’t need a new connection for every testcase), so testcase will only execute APIs on the socket (from initialisation) rather than making a new connection every single time.

Is there anything that could go wrong in this?

jesperpedersen:
No, should be ok. And, you can do `pgmoneta_socket_isvalid` for each test case

PR #529, opened 2025-03-21
[#474] tscli impl

ashu3103:
## WIP

@jesperpedersen, I was facing some `Broken Pipe 13` issues when I was using the same connection for all the testcase APIs. So currently, the implementation will get a separate connection for each API.

Any ideas why it should behave like that?


jesperpedersen:
Likely because the client side is closed... you didn't change `SO_KEEPALIVE`, right ?

I have no idea on the CI atm...

ashu3103:
> Likely because the client side is closed... you didn't change `SO_KEEPALIVE`, right ?
> 

No, I didn't change them, and so by default it should be `on`.

ashu3103:
Hi @jesperpedersen,

Currently the final build of our CI pipeline is in `clang` and `debug` mode, but I could see some MemoryLeaks in `pgmoneta-admin` while executing testsuites which either exits the testsuite with exit code 1 and therefore providing error or result in partial execution of script.

I was thinking, can we keep our final build in `clang` and `Release` mode?

ashu3103:
Also, some changes that we can do in this PR in testsuite script

- We are just executing the testcases without caring about the output (so the CI pipeline passed even when all tests were not successful) [my bad!].
- Currently we are waiting for 5 seconds to ensure server has started completely, instead we can just do ping in a loop.
- Also there were some areas where we can improve error handling.

jesperpedersen:
We should run the testsuite for both GCC/Debug and clang/Debug.

A memory leak counts as a failure.

We can loop on `pg_isready`

ashu3103:
> We should run the testsuite for both GCC/Debug and clang/Debug.
> 
> A memory leak counts as a failure.
> 

Ohh okay.

> We can loop on `pg_isready`

Yeah, that's more appropriate, thanks!

jesperpedersen:
Can you change the CI too

ashu3103:
Hi,

> Can you change the CI too

Can you please tell what changes are you referring to in CI exactly?

jesperpedersen:
Making sure that the testsuite runs with both GCC/Debug and clang/Debug

ashu3103:
> Making sure that the testsuite runs with both GCC/Debug and clang/Debug

Yeah I'll do that

ashu3103:
@jesperpedersen

Incorporated https://github.com/pgmoneta/pgmoneta/pull/529#issuecomment-2746147479 and https://github.com/pgmoneta/pgmoneta/pull/529#issuecomment-2746299115

PTAL

.github/workflows/ci.yml
```diff
@@ -68,6 +80,13 @@ jobs:
     - name: GCC/make
       run: make
       working-directory: /home/runner/work/pgmoneta/pgmoneta/build/
+    - name: GCC/Run Test Suite
+      id: gcc-test
+      run: $(which bash) ./testsuite.sh
+      working-directory: /home/runner/work/pgmoneta/pgmoneta/build
+    - name: GCC/Cleanup
```
jesperpedersen, line 87: GCC/Testsuite Clean

.github/workflows/ci.yml
```diff
@@ -68,6 +80,13 @@ jobs:
     - name: GCC/make
       run: make
       working-directory: /home/runner/work/pgmoneta/pgmoneta/build/
+    - name: GCC/Run Test Suite
```
jesperpedersen, line 83: GCC/Testsuite

.github/workflows/ci.yml
```diff
@@ -80,29 +99,19 @@ jobs:
...
-        sudo curl -o /usr/share/postgresql-common/pgdg/apt.postgresql.org.asc --fail https://www.postgresql.org/media/keys/ACCC4CF8.asc
-        sudo sh -c 'echo "deb [signed-by=/usr/share/postgresql-common/pgdg/apt.postgresql.org.asc] https://apt.postgresql.org/pub/repos/apt $(lsb_release -cs)-pgdg main" > /etc/apt/sources.list.d/pgdg.list'
-        sudo apt update
-        sudo apt install -y postgresql-17
-    - name: Set Env Path Variable
-      run: |
-        echo "PATH=$PATH:/usr/lib/postgresql/17/bin" >> $GITHUB_ENV
-        echo $PATH
-    - name: Run Test Suite
+    - name: CLANG/Run Test Suite
```
jesperpedersen, line 102: clang/Testsuite

.github/workflows/ci.yml
```diff
@@ -80,29 +99,19 @@ jobs:
...
+      id: clang-test
       run: $(which bash) ./testsuite.sh
       working-directory: /home/runner/work/pgmoneta/pgmoneta/build
-    - name: Cleanup
+    - name: CLANG/Cleanup
```
jesperpedersen, line 106: clang/Testsuite clean

test/CMakeLists.txt
```diff
@@ -28,20 +28,42 @@
...
   set(SOURCES
-    testcases/common.c
+    testcases/tsclient.c
     testcases/pgmoneta_test_1.c
     testcases/pgmoneta_test_2.c
     testcases/runner.c
   )
 
+  if (CMAKE_BUILD_TYPE MATCHES Debug)
+    add_compile_options(-O0)
```
jesperpedersen, line 38: I think for the testsuite we can add these 2 options always

test/testcases/pgmoneta_test_1.c
```diff
@@ -28,93 +28,31 @@
  */
 
 #include "pgmoneta_test_1.h"
-#include "common.h"
+#include "tsclient.h"
```
jesperpedersen, line 31: Lets see if we can get these to `#include <tsclient.h>` instead

test/testcases/runner.c
```diff
@@ -56,5 +62,8 @@ main(int argc, char* argv[])
    number_failed = srunner_ntests_failed(sr);
    srunner_free(sr);
 
+error:
```
jesperpedersen, line 65: You should do a `return` above this to make it clear it is a success path

test/testcases/tsclient.c
```diff
@@ -0,0 +1,298 @@
...
+    {
+        goto error;
+    }
+
+    // Fallbacks
+    if (backup_id == NULL)
+    {
+        backup_id = "newest";
+    }
+    if (position == NULL)
```
jesperpedersen, line 161: `position` is optional, so we need to account for that

jesperpedersen (changes requested):

test/testcases/runner.c
ashu3103: Ahh, It was a more like `done` label not an `error`, since we have to do `pgmoneta_tsclient_destroy` everytime the process exits.

I'll change that

test/testcases/pgmoneta_test_1.c
ashu3103: We probably have to create a shared library for `tsclient` if I am not wrong?

jesperpedersen: You can do that...

Isn't `include_directories` in `CMakeLists.txt` enough ?

ashu3103: Not a pro of `cmake`, I can see if that's an option.

ashu3103: I have changed the test directory structure a bit, check if its ok

jesperpedersen:
@ashu3103 Yes, it is ok

jesperpedersen:
Rebased, and merged.

PR #508, opened 2025-03-09
[#415] Implemented the translation and filtering

Mohab96:
Resolves #415

Now the user can type his OID mappings in json file and run this command to get the names of relations, databases, or tablespaces in the description of the XLOG records by running this command:
```
pgmoneta-walinfo --translate --mapping mapping.json walfile
```
and if `--mapping` was missing it will use `/etc/pgmoneta/pgmoneta_walinfo_conf.json` by default.

PR #531, opened 2025-03-22
[#530] feat/cli-parser

### LINKED ISSUE #530: Command line parsing
We are running into situations where `getopt_long` doesn't work since we can't do `-XY` with a `--xyz` key as well.

Create a `cmd.h` and `cmd.c` that can take a struct that contains a short name, a long name, and specifies if an argument is required.

The API should be as simple as possible, so `argc`, `argv` and a `deque` of these structs f.ex. Output could be a `deque` with another struct that says the short name, and the argument if required

Mohab96:
Resolves #530 

This PR introduces a new library for parsing CLI arguments to replace `getopt_long`.

The key change in this lib is that now we can make short options also strings, not chars 
so we can do something like this `-RD, --databases` which we couldn't do in `getopt_long`.

Example Usage:
```c
int
main(int argc, char** argv)
{
   cli_option options[] = {
      {"RD", "databases", true},     // -RD or --databases requires an argument
      {"v", "verbose", false},       // -v or --verbose doesn't require an argument
      {"h", "help", false},          // -h or --help doesn't require an argument
      {"o", "output", true}          // -o or --output requires an argument
   };
   int num_options = sizeof(options) / sizeof(options[0]);

   cli_result results[10]; // Assuming we won't have more than 10 options

   int num_results = cmd_parse(argc, argv, options, num_options, results, 10);

   if (num_results < 0)
   {
      printf("Error parsing command line\n");
      return 1;
   }

   printf("Found %d options:\n", num_results);
   for (int i = 0; i < num_results; i++)
   {
      if (results[i].argument)
      {
         printf("Option: %s, Argument: %s\n", results[i].option_name, results[i].argument);
      }
      else
      {
         printf("Option: %s (no argument)\n", results[i].option_name);
      }
   }

   return 0;
}
```


src/include/cmd.h
```diff
@@ -0,0 +1,48 @@
+#ifndef CMD_H
```
jesperpedersen, line 1: License header

src/include/cmd.h
```diff
@@ -0,0 +1,48 @@
...
+
+#include <stdbool.h>
+
+/**
+ * @struct cli_option
+ * Struct to hold option definition
+ */
+typedef struct
+{
+   char* short_name; /**< Short option name */
```
jesperpedersen, line 12: Comment alignment

src/include/cmd.h
```diff
@@ -0,0 +1,48 @@
...
+   bool requires_arg; /**< Whether this option requires an argument */
+} cli_option;
+
+/**
+ * @struct cli_result
+ * Struct to hold parsed option result
+ */
+typedef struct
+{
+   char* option_name; /**< The matched option name (short or long) */
```
jesperpedersen, line 23: Comment alignment

src/include/cmd.h
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @return Number of results found, or -1 on error
+ */
+int cmd_parse(
+   int argc,
+   char** argv,
+   cli_option* options,
+   int num_options,
+   cli_result* results,
+   int max_results
```
jesperpedersen, line 45: num_results

src/libpgmoneta/cmd.c
```diff
@@ -0,0 +1,88 @@
+#include "cmd.h"
```
jesperpedersen, line 1: License header

jesperpedersen: <cmd.h>

src/libpgmoneta/cmd.c
```diff
@@ -0,0 +1,88 @@
...
+#include <err.h>
+
+int
+cmd_parse(
+   int argc,
+   char** argv,
+   cli_option* options,
+   int num_options,
+   cli_result* results,
+   int max_results
```
jesperpedersen, line 13: num_results

jesperpedersen (changes requested):

jesperpedersen:
Change all files using `getopt_long` right now

Mohab96:
I will go for these 4 files first because the updates that need to be done are straight forward, and move `utils.h` to the end of the queue.

image

Mohab96:
I will continue the rest tomorrow

Mohab96:
I updated the branch, please review it whenever you have time.

Mohab96:
I am trying to solve the problem with the CI.

jesperpedersen:
You can submit the admin.c change separate as a bug

jesperpedersen:
Cc: @ashu3103

Mohab96:
> You can submit the admin.c change separate as a bug

If so, then you can merge this patch now.

jesperpedersen:
You know how it works - fix roadblocks first

Mohab96:
> You know how it works - fix roadblocks first

Then you want me revert changes made in `admin.c` and make it use `getopt_long` to make the CI happy then in other issue describe what's needed to be done to change the `admin.c` to use the new parser instead of `getopt_long`, right?

Mohab96:
The problem is that when it finds a non-option arg (something not preceded with - or --) it terminates the parsing, so I think this problem will appear in `cli.c` also, so I will report it in the same issue I will open.

jesperpedersen:
You start on a new branch based on `main`, fix the issue reported by ASAN that you are running into, then we merge, then you rebase this PR, and then we can continue from there

Mohab96:
But this issue appeared because of the new `cmd_parse` function introduced in this PR, so if I do this, I think I need this to be done in this PR. Or I revert changes done in `admin.c` and `cli.c` and you merge then I open another PR to migrate them from `getopt_long` to `cmd_parse` and fix the issues that appear in that new PR. Which one do you prefer?

jesperpedersen:
Ok, do everything in this PR then

src/admin.c
```diff
@@ -155,83 +156,95 @@ usage(void)
...
-   int option_index = 0;
    size_t command_count = sizeof(command_table) / sizeof(struct pgmoneta_command);
    struct pgmoneta_parsed_command parsed = {.cmd = NULL, .args = {0}};
    int32_t output_format = MANAGEMENT_OUTPUT_FORMAT_TEXT;
 
    // Disable stdout buffering (i.e. write to stdout immediatelly).
    setbuf(stdout, NULL);
 
-   while (1)
+   cli_option options[] = {
```
jesperpedersen, line 171: Move variable definitions to the top

src/libpgmoneta/cmd.c
```diff
@@ -0,0 +1,143 @@
...
+   cli_result* results,
+   int num_results,
+   bool use_last_arg_as_filename,
+   char** filename,
+   int* optind
+   )
+{
+   int result_count = 0;
+   *filename = NULL;
+   int i = 1;
```
jesperpedersen, line 49: Do we need this ?

src/libpgmoneta/cmd.c
```diff
@@ -0,0 +1,143 @@
...
+   int num_options,
+   cli_result* results,
+   int num_results,
+   bool use_last_arg_as_filename,
+   char** filename,
+   int* optind
+   )
+{
+   int result_count = 0;
+   *filename = NULL;
```
jesperpedersen, line 48: Separate with a newline

src/libpgmoneta/cmd.c
```diff
@@ -0,0 +1,143 @@
...
+   int i = 1;
+
+   for (; i < argc && result_count < num_results; i++)
```
jesperpedersen, line 51: Put it here

src/libpgmoneta/cmd.c
```diff
@@ -0,0 +1,143 @@
...
+   {
+      const char* arg = argv[i];
```
jesperpedersen, line 53: I don't like `const`

src/main.c
```diff
@@ -251,64 +251,77 @@ main(int argc, char** argv)
...
    int ret;
-   int c;
    char* os = NULL;
 
    int kernel_major, kernel_minor, kernel_patch;
 
    argv_ptr = argv;
 
-   while (1)
+   cli_option options[] = {
```
jesperpedersen, line 260: Same here - move to variable definitions

src/main.c
```diff
@@ -251,64 +251,77 @@ main(int argc, char** argv)
...
+         version();
+      }
+      else if (strcmp(optname, "?") == 0 || strcmp(optname, "help") == 0)
+      {
+         usage();
+         exit(0);
       }
    }
 
+   return 0;
```
jesperpedersen, line 323: ?

src/walinfo.c
```diff
@@ -106,155 +102,173 @@ main(int argc, char** argv)
...
+      {"e", "end", true},
+      {"x", "xid", true},
+      {"l", "limit", true},
+      {"v", "verbose", false},
+      {"V", "version", false},
+      {"?", "help", false},
+   };
+
+   int num_options = sizeof(options) / sizeof(options[0]);
+   char* filepath;
```
jesperpedersen, line 125: Always init variables

jesperpedersen (changes requested):

jesperpedersen:
Left a few comments for now - apply them across the patch

ashu3103:
@Mohab96 

The part in `testsuite.sh` script where we are creating a master-key using the command
`./pgmoneta-admin master-key -P $PGPASSWORD || true` (The exact command we are executing)

is giving error :-
`pgmoneta-admin: Too many arguments provided for command 'master-key'`

If you can fix that, the CI will execute smoothly, but unfortunately there are still some memory leaks in there.

jesperpedersen:
@ashu3103 Yes, you can do that - it need to be fixed

Mohab96:
I think the parser is working fine now. This memory leak is neither related to this patch nor introduced by it, so unless there are comments in the code review, I think this patch is ok.

src/cli.c
```diff
@@ -442,140 +441,158 @@ main(int argc, char** argv)
    size_t command_count = sizeof(command_table) / sizeof(struct pgmoneta_command);
    struct pgmoneta_parsed_command parsed = {.cmd = NULL, .args = {0}};
 
+   cli_option options[] = {
```
jesperpedersen, line 444: I would make this one the last definition

src/cli.c
```diff
@@ -442,140 +441,158 @@ main(int argc, char** argv)
...
-         {"password", required_argument, 0, 'P'},
-         {"logfile", required_argument, 0, 'L'},
-         {"verbose", no_argument, 0, 'v'},
-         {"version", no_argument, 0, 'V'},
-         {"format", required_argument, 0, 'F'},
-         {"help", no_argument, 0, '?'},
-         {"compress", required_argument, 0, 'C'},
-         {"encrypt", required_argument, 0, 'E'}
-      };
+      warnx("Error parsing command line\n");
```
jesperpedersen, line 471: errx

jesperpedersen:
Some comments - apply them across the entire patch

Mohab96:
Done

jesperpedersen:
Rebased, and merged.

jesperpedersen:
@Mohab96 Please, port to pgagroal and pgexporter

Mohab96:
Should I go to finish #508 first? @jesperpedersen

jesperpedersen:
@Mohab96 Whatever order you want

Issue #533, opened 2025-03-26
Add docker support to pgmoneta


Arshdeep54:
Add docker or podman support to pgmoneta 

Similar to [Dockerize pgexporter
](https://github.com/pgexporter/pgexporter/issues/167)

Arshdeep54:
@jesperpedersen , can you please assign me this issue?

jesperpedersen:
@Arshdeep54 Go for it !

PR #540, opened 2025-03-27
[#533] Feat : Add docker support to pgmoneta

Arshdeep54:


contrib/docker/pgmoneta.conf
```diff
@@ -0,0 +1,20 @@
+host = *
```
jesperpedersen, line 1: Missing `[pgmoneta]`

jesperpedersen: Does that work ? if so it is a bug...

jesperpedersen (changes requested):

contrib/docker/pgmoneta.conf
Arshdeep54: No, it was not working, so I kept the PR as a Draft. Now its working, Thanks

Arshdeep54:
@jesperpedersen In the Alpine-based Dockerfile, the build requires `systemd` dependencies, unlike `pgexporter`, where they were optional. However, since Alpine does not have systemd, is there an alternative solution, or would it be better to use a different base image?

jesperpedersen:
@Arshdeep54 #545

jesperpedersen:
Merged.

Issue #525, opened 2025-03-18
Split configuration


jesperpedersen:
We need a common configuration such that we can do main configuration, and configuration for the other tools like walinfo and cli

jesperpedersen:
@Mohab96 Can you take this ?

Mohab96:
Yeah, of course. I was going to do this task eventually after finishing #508

Mohab96:
What this issue aims to accomplish is detailed in [this comment](https://github.com/pgmoneta/pgmoneta/pull/508#issuecomment-2720906804) along with the four subsequent comments.

_Just for documenting the process_

Mohab96:
We can say that the current `configuration` structure is the `main_configuration`, and the `walinfo_configuration` can be easily created (its attributes are stated previously in #508) and its purpose is obvious. What I can't understand is what is the purpose of `common_configuration` and subsequently, can neither know which attributes will need to be used inside it nor where it is expected to be used. So can you please explain to me the purpose of `common_configuration`?

jesperpedersen:
@Mohab96 `common_configuration` is the base `struct` where all configuration properties that are shared between configurations are located.

So, that would be all the logging configuration for now.

Of course `struct common_configuration` has to be first field in the other `struct`s (main, walinfo, cli, ...)

Mohab96:
I might need #508 to get merged first to start updating the parts of accessing configuration attributes based on this update that will be done because I implemented `walinfo` configurations there. 

For example,

change this `config->log_type = PGMONETA_LOGGING_TYPE_CONSOLE;`
to this `config->base->log_type = PGMONETA_LOGGING_TYPE_CONSOLE;`

what do you think?

jesperpedersen:
@Mohab96 Yes, you do all the common, main and walinfo changes in this patch, then you rebase and work on finishing #508

PR #548, opened 2025-03-28
[#525] Fixed a bug in `zstd_decompress`

Mohab96:
This solves the infinite loop that occurs on my machine. However, the file's compression appears to be somewhat broken since the `input.pos` is not changing in the inner while loop.

```
pgmoneta@apollo:~$ pgmoneta-walinfo 000000010000000000000004.zstd.aes
Error while reading/describing WAL file
pgmoneta@apollo:~$ cat /tmp/pgmoneta.log
2025-03-28 14:18:06 TRACE utils.c:2548 FILETRACKER | Copy | 000000010000000000000004.zstd.aes | /tmp |
2025-03-28 14:18:06 ERROR aes.c:932 EVP_CipherFinal_ex: failed to process final cipher block
2025-03-28 14:18:06 ERROR zstandard_compression.c:471 ZSTD: Could not decompress /tmp/000000010000000000000004.zstd
2025-03-28 14:18:06 FATAL walfile.c:380 Failed to decompress WAL file at 000000010000000000000004.zstd.aes
pgmoneta@apollo:~$
```

Please test on your machine and tell me whether this strange output is still there. @jesperpedersen

### DIFF
--- a/src/libpgmoneta/restore.c
+++ b/src/libpgmoneta/restore.c
@@ -1544,7 +1544,7 @@ restore_backup_incremental(struct art* nodes)
    char target_root_combine[MAX_PATH];
    char target_base_combine[MAX_PATH_CONCAT];
    char excluded_file_path[MAX_PATH_CONCAT];
-   char tmp_excluded_file_path[MAX_PATH_CONCAT+sizeof(TMP_SUFFIX)];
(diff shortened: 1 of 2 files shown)


jesperpedersen:
Looks like compress has the same issue -- can you go over the entire file ?

Mohab96:
PTAL @jesperpedersen 
If the patch is good, I will rebase now.

jesperpedersen:
Rebased, and merged.

PR #558, opened 2025-04-01
[#537] Optimized CRC32 to use PCLMULQWQ when available

### LINKED ISSUE #537: Optimize CRC32C
We can optimize the CRC32C implementation based on

 https://www.postgresql.org/message-id/flat/PH8PR11MB82869FF741DFA4E9A029FF13FBF72%40PH8PR11MB8286.namprd11.prod.outlook.com

We don't need all the checks - we will use it if the platform has the instruction

Mohab96:
Resolves #537


Mohab96:
I wasn't able to thoroughly test it, and still working on how to do so, but you can review now if you wish. @jesperpedersen

Mohab96:
I wasn't able to thoroughly test it, and still working on how to do so, but you can review now if you wish. @jesperpedersen

jesperpedersen:
I wonder if we should make it a run-time check... it might be compiled on a platform with `PCLMUL` but executed on a platform with only `SSE 4.2`.

WDYT ?

Mohab96:
I can't imagine a case where the system is compiled on a machine and executed on another machine.

jesperpedersen:
yum.postgreql.org ? We don't know the machines in the various build-farms that build our binaries for their distribution...

I think we are fairly safe on `x86_64` - its more `aarch64` I'm worried about

Mohab96:
No problem, I can do it. Would you like me to determine this every time inside `pgmoneta_create_crc32c_buffer`, or should I opt for a more modular approach that handles CPU features in general for any purpose other than CRC32? 
(e.g. `struct cpu_features`)

jesperpedersen:
`struct cpu_features` sounds good, otherwise put it `struct common_configuration`

jesperpedersen:
You can do an `uint32_t` and do flags to pack it

Mohab96:
But I recommend it to be a struct because not all attributes will be boolean values. Perhaps in the future, there will be values that are something else. Is this okay, or do you recommend the int solution and using right/left shifting to obtain values?

Mohab96:
Currently, I need to add some `config->common.x = y;` statements. However, in the existing implementation, I would have to duplicate them in every `pgmoneta_init_*_configuration`, resulting in unnecessary code duplication. To avoid this, I plan to create a new `pgmoneta_init_common_configuration`. If I proceed with this now on this branch, things will be messy when we attempt to merge #508. Therefore, is it possible to merge #508 first and then continue with this one? 

I can mimic this update that needs to be done here, and continue working to split `pgmoneta_create_crc32c_buffer` based on the criteria we agreed on so that this PR doesn't get blocked until #508 is merged.

@jesperpedersen

jesperpedersen:
Maybe you can extract the configuration split into another PR - will that make it easier ?

jesperpedersen:
I don't see how the configuration split can be a problem... but ok.

Yes, or you can do `cpu_feature_pclmul` -- the key part is the run-time detection

Mohab96:
> Yes, or you can do `cpu_feature_pclmul` -- the key part is the run-time detection

Do you recommend this instead of this?
```c
struct cpu_features {
   bool pclmul;
   bool sse42;
} __attribute__ ((aligned (64)));
```

jesperpedersen:
Do what you think is best

Mohab96:
PTAL @jesperpedersen

jesperpedersen:
@Mohab96 Where are they initialized and detected as `true` ?

Mohab96:
They're not initialized, so whatever their values they're garbage values.

> in `pgmoneta_create_crc32c_buffer` and pretend that these variables are initialized, and then merge this PR (it will not be working), and then in #508 we can do the `init_config` split.

This feature is not working; just try to initialize them elsewhere to be able to test.

jesperpedersen:
Otherwise, we will just use the compile time ones...

jesperpedersen:
run-time would be better, so spend a bit of time on that

Mohab96:
Okay, but we agreed that I will do the split in #508 after this PR gets merged, and the mechanism of extraction of CPU features will be decided there, so this PR is ok if the modifications made until now are ok.

jesperpedersen:
You can decide the CPU features here

jesperpedersen:
Otherwise, just revert `struct cpu_features` part - and we can do it later

Mohab96:
Then you want me to add some more features other than sse42 and pclmul, right?

jesperpedersen:
?

No, `pclmul` and `sse4.2` are the important ones for this patch.

We are looking patch-by-patch at what we need

Mohab96:
> You can decide the CPU features here

Then what do you mean here? Since I've already decided on the CPU features I need.

jesperpedersen:
Just make it a compile time check for now, and we can take it later.

So, remove `struct cpu_features`

Mohab96:
PTAL @jesperpedersen

src/libpgmoneta/security.c
```diff
@@ -3053,14 +3163,28 @@ pgmoneta_create_crc32c_buffer(void* buffer, size_t size, uint32_t* crc)
...
+      return 1;
+   }
+
+   #ifdef HAVE_PCLMUL
+   return pgmoneta_crc32c_pclmul(buffer, size, crc);
+   #elif HAVE_CRC32
+   return pgmoneta_crc32c_sse42(buffer, size, crc);
    #endif
+
+   return pgmoneta_crc32c_software(buffer, size, crc);
```
jesperpedersen, line 3187: Put this inside an `#else`

jesperpedersen (changes requested):

src/libpgmoneta/security.c
Mohab96: I did this the first time, but clang told me that this function is not used so I put it this way.

Mohab96:
Done @jesperpedersen

jesperpedersen:
Rebased, and merged.

Issue #551, opened 2025-03-28
Incremental backup restore: worker support


Jubilee101:
Support workers when restoring incremental backup

PR #583, opened 2025-04-07
[#551] Enable workers for incremental backup combine

Jubilee101:


src/include/deque.h
```diff
@@ -280,6 +280,13 @@ pgmoneta_deque_to_string(struct deque* deque, int32_t format, char* tag, int ind
 void
 pgmoneta_deque_destroy(struct deque* deque);
 
+/**
+ * Set the deque to be thread safe
+ * @param deque The deque
+ */
+void
+pgmoneta_deque_set_thread_safe(struct deque* deque);
```
jesperpedersen, line 288: Why do we need this ? its _create(true, ...

src/libpgmoneta/restore.c
```diff
@@ -153,11 +172,28 @@ reconstruct_backup_file(int server,
...
+                        struct art* backups,
                         bool incremental,
                         int algorithm,
                         struct json* files);
 
+static void
+do_reconstruct_backup_file(struct worker_common* wc);
+
+static void
+create_reconstruct_backup_file_input(int server,
```
jesperpedersen, line 184: Should we wait for _load / _save ?

src/libpgmoneta/restore.c
```diff
@@ -698,13 +777,27 @@ pgmoneta_combine_backups(int server, char* label, char* base, char* input_dir, c
...
+   }
+
+   pgmoneta_art_destroy(backups);
    pgmoneta_deque_iterator_destroy(iter);
    free(server_dir);
    return 0;
 
 error:
-   pgmoneta_art_destroy(prior_backups);
+   if (number_of_workers > 0)
```
jesperpedersen, line 796: destroy should work on `NULL`, no ?

src/libpgmoneta/restore.c
```diff
@@ -2410,4 +2540,141 @@ construct_backup_label_chain(int server, char* newest_label, char* oldest_label,
...
+                               input->prior_labels,
+                               input->backups,
+                               input->incremental,
+                               input->algorithm,
+                               input->files))
+   {
+      goto error;
+   }
+   input->common.workers->outcome = true;
+   free(input);
```
jesperpedersen, line 2594: ?

src/libpgmoneta/restore.c
```diff
@@ -2410,4 +2540,141 @@ construct_backup_label_chain(int server, char* newest_label, char* oldest_label,
...
+   if (copy_backup_file(input->server,
+                        input->label,
+                        input->output_dir,
+                        input->relative_dir,
+                        input->file_name,
+                        input->exclude))
+   {
+      goto error;
+   }
+   free(input);
```
jesperpedersen, line 2617: ?

src/libpgmoneta/restore.c
Jubilee101: This is just for creating input for workers, not backup.info related... We can do another round with the naming and stuff though.

src/include/deque.h
Jubilee101: Yeah it's a little ad-hoc. The deque is inside json so it's not created with thread safe set and can only be set later. We can revisit this if we have more use cases as such.

src/libpgmoneta/restore.c
Jubilee101: Maybe, I'm just copying over. It's written this way in other places.

PR #585, opened 2025-04-08
[#533] Fix Docker builds

Arshdeep54:


src/libpgmoneta/utils.c
```diff
@@ -68,6 +67,10 @@
 #define EVBACKEND_IOURING  0x00000080U
 #endif
 
+#ifdef HAVE_SYSTEMD
```
jesperpedersen, line 70: We should have a HAVE_EXECINFO_H for this

src/libpgmoneta/utils.c
```diff
@@ -4923,7 +4926,7 @@ calculate_offset(uint64_t addr, uint64_t* offset, char** filepath)
 int
 pgmoneta_backtrace(void)
 {
-#ifdef HAVE_LINUX
+#ifdef HAVE_SYSTEMD
```
jesperpedersen, line 4929: Same

jesperpedersen (changes requested):

jesperpedersen:
@Arshdeep54 But you didn't define it...
@silver-ymz Cc:

silver-ymz:
@Arshdeep54  You need a `include(CheckIncludeFile)` in `CMakeLists.txt`

Arshdeep54:
Thanks @silver-ymz

jesperpedersen:
@Arshdeep54 You should be able to verify on FreeBSD 14.2

silver-ymz:
`calculate_offset` currently only supports linux. `macos` and `freebsd` have different format. We may need to use `HAVE_LINUX` now.

jesperpedersen:
@silver-ymz Ok, you could test FreeBSD 14.2, or use the CI - it has Mac and FreeBSD now

jesperpedersen:
@Arshdeep54 Where are we ?

Arshdeep54:
Was doing the load/save thing, will update this soon.

Arshdeep54:
>calculate_offset should use HAVE_EXECINFO_H too

Yes, although it doesn't use anything from execinfo.h but can be guarded for consistency and to prevent unnecessary compilation on unsupported systems. 

as @silver-ymz pointed out, `calculate_offset` is only for Linux, because there is no file as `/proc/self/maps` in macOS and FreeBSD 
And just noticed, this `pgmoneta_backtrace` function is used nowhere. 
My gnome-Boxes VM just crashed, will get the Azure student VM by today .Let me know if I should write the further implementation for FreeBSD

jesperpedersen:
@Arshdeep54 Expanding the usage of `pgmoneta_backtrace` on other platforms is a separate issue -- this is a function developers can use to better understand their bugs

jesperpedersen:
@Arshdeep54 Please, port if needed

Issue #524, opened 2025-03-14
Incremental backup restore size calculation


Jubilee101:
We record backup's upon restore inside backup.info. However the value is inaccurate for the incremental backups. Since it only reflects the size before backup combine. We need to calculate the backup size after combine, an idea is to go through incremental files and calculate their size after restore using `truncation_block_length`, note that this is for v17 incremental backup only.

ashu3103:
@Jubilee101, can I take this one?

Jubilee101:
Hey @ashu3103 , thanks for the interest! I created the issue mostly to remind myself, but sure you are welcome to take it. It's a good opportunity for you get a glimpse at how incremental backup/restore works for postgres 17. Note that the manifest contains the size of each file, so you can read from there directly if the file is not incremental. You should find a lot of helper APIs to accomplish the task, try not to reinvent a wheel.

ashu3103:
> Hey [@ashu3103](https://github.com/ashu3103) , thanks for the interest! I created the issue mostly to remind myself, but sure you are welcome to take it.  It's a good opportunity for you get a glimpse at how incremental backup/restore works for postgres 17. Note that the manifest contains the size of each file, so you can read from there directly if the file is not incremental. You should find a lot of helper APIs to accomplish the task, try not to reinvent a wheel.

Sure, I'll keep you informed before proceeding with the implementation.

Thanks :)

jesperpedersen:
@ashu3103 Where are we on this ?

ashu3103:
@jesperpedersen, Sorry preoccupied with some other work.

Will try to raise a PR by this Wednesday.

ashu3103:
Hi @Jubilee101,

If a table is truncated, dropped, or **unchanged**, its header block becomes a **12-byte block** containing three fields: a magic number, a number of incremental blocks set to 0, and a truncation block length of 1. The INCREMENTAL file itself also becomes a 12-byte file containing only the header. - Quoting this from [Internals of Postgres](https://www.interdb.jp/pg/pgsql10/05.html#:~:text=If%20a%20table%20is%20truncated,file%20containing%20only%20the%20header.)

My idea was to go through all the manifest entry of the incremental backup, now if a file is not incremental we will directly get the size from the entry, if the file is incremental and its size > 12 bytes, we will read the `truncation block length` from the header of the file but for incremental file with size = 12bytes there can be 2 cases:
[...]

PR #590, opened 2025-04-10
[#524] Incremental backup restore size calculation

ashu3103:
## Work in Progress

The objective is to calculate the approximate restoration size of incremental backup as per backup combine. The idea is to go through all the file entries in the manifest of the backup.

- If the file is non incremental get the size from the manifest file itself (in bytes).
- Else get the `block_length` using the function `find_reconstructed_block_length` -> which provides the total blocks in the actual restored file.

@Jubilee101, @jesperpedersen PTAL.


src/libpgmoneta/utils.c
```diff
@@ -4854,6 +4854,31 @@ pgmoneta_resolve_path(char* orig_path, char** new_path)
    return 1;
 }
 
+int
+pgmoneta_split_file_path(char* path, char** relative_path, char** bare_file_name)
```
Jubilee101, line 4858: Can you use `dirname()` here?

Jubilee101, line 4858: Can you use `dirname()` here?

src/libpgmoneta/restore.c
```diff
@@ -818,6 +818,108 @@ pgmoneta_combine_backups(int server, char* label, char* base, char* input_dir, c
...
+   struct json* manifest_read = NULL;
+   struct json* files = NULL;
+   struct main_configuration* config = NULL;
+   struct json_iterator *iter = NULL;
+   char* manifest_path = NULL;
+   unsigned long sz = 0;
+   uint64_t biggest_file_sz = 0;
+
+   config = (struct main_configuration*)shmem;
+   pgmoneta_art_create(&nodes);
```
Jubilee101, line 835: Why do you need nodes?

src/libpgmoneta/restore.c
```diff
@@ -818,6 +818,108 @@ pgmoneta_combine_backups(int server, char* label, char* base, char* input_dir, c
...
+   pgmoneta_json_iterator_create(files, &iter);
+   while (pgmoneta_json_iterator_next(iter))
+   {
+      struct json* file = NULL;
+      char* file_path = NULL;
+      uint64_t file_size = 0;
+
+      file = (struct json*)pgmoneta_value_data(iter->value);
+      file_path = (char*)pgmoneta_json_get(file, "Path");
+      if (pgmoneta_contains(file_path, INCREMENTAL_PREFIX)) // for incremental files get the `truncated_block_length`
```
Jubilee101, line 866: Use `pgmoneta_is_incremental_path()`, in case (though unlikely) some path has incremental prefix somewhere in the middle

src/include/utils.h
```diff
@@ -779,6 +779,16 @@ pgmoneta_strip_extension(char* s, char** name);
...
 
+/**
+ * Split the file path into relative and bare file path
+ * @param path The file path
+ * @param relative_path [out] The relative path
+ * @param bare_file_name [out] The bare file name
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_split_file_path(char* path, char** relative_path, char** bare_file_name);
```
Jubilee101, line 790: For now maybe keep it as a static function inside restore.c

src/libpgmoneta/restore.c
```diff
@@ -818,6 +818,108 @@ pgmoneta_combine_backups(int server, char* label, char* base, char* input_dir, c
...
+         char* relative_path = NULL;
+         char* bare_file_name = NULL;
+
+         if (pgmoneta_split_file_path(file_path, &relative_path, &bare_file_name))
+         {
+            pgmoneta_log_error("Unable to split file path %s", file_path);
+            goto error;
+         }
+
+         if (incremental_rfile_initialize(server, label, relative_path, bare_file_name, backup, &rf))
```
Jubilee101, line 879: backup.info hasn't been created when you calculate the size. And you don't actually need it here. I need it because I need the backup's compression and encryption method for potentially decompression and decryption. But when you calculate the size it hasn't been compressed nor encrypted.

Maybe simply change `incremental_rfile_initialize` to accept `compression` and `encryption` directly instead of the whole backup, and pass in `COMPRESSION_NONE` and `ENCRYPTION_NONE` here

src/include/restore.h
```diff
@@ -117,6 +117,17 @@ pgmoneta_rollup_backups(int server, char* newest_label, char* oldest_label);
...
+/**
+ * Get the size of the incremental backup
+ * @param server The server
+ * @param label The label
+ * @param size [out] The size of the incremental backup
+ * @param biggest_file_size [out] The size of the biggest file in the incremental backup
+ * @return 0 on success, 1 if otherwise
+ */
+int
+pgmoneta_backup_size_incremental(int server, char* label, unsigned long* size, uint64_t* biggest_file_size);
```
Jubilee101, line 129: Maybe just pgmoneta_backup_size()... The way you wrote it works on full backup as well.

Jubilee101 (changes requested):

src/libpgmoneta/restore.c
ashu3103: Yeah makes sense, will make the changes.

src/libpgmoneta/restore.c
jesperpedersen: This is a bit weird... can't you use info.h directly ?

ashu3103: Using `pgmoneta_get_backup_server()`?

But that will also eventually end up accessing the `backup.info` file while our plan is to shift the size calculation before this file creation.

ashu3103:
@Jubilee101,  @jesperpedersen I have made the desired changes, PTAL

Thanks

src/libpgmoneta/wf_backup.c
```diff
@@ -417,10 +418,21 @@ basebackup_execute(char* name, struct art* nodes)
 
    backup_data = pgmoneta_get_server_backup_identifier_data(server, label);
 
-   size = pgmoneta_directory_size(backup_data);
+   if (!incremental)
+   {
+      size = pgmoneta_directory_size(backup_data);
```
ashu3103, line 423: Shall I use `pgmoneta_backup_size` here too?

ashu3103: @Jubilee101

Jubilee101: Hmm, @jesperpedersen thoughts? Maybe we can use the one that's faster.

Jubilee101 (changes requested):

ashu3103:
@Jubilee101, Done

Jubilee101:
Looking good, can you restore an incremental backup and verify its size?

ashu3103:
Encountering this error again and again
```
2025-04-11 01:42:23 ERROR restore.c:2620 restore_backup_incremental: could not rename file /tmp/restore/primary-20250411011348/postgresql.conf.tmp to /tmp/restore/primary-20250411011348/postgresql.conf
```

Don't think it has something to do with our backup size calculation infra

Jubilee101:
Can you log the errno message?

ashu3103:
> Can you log the errno message?

No such file or directory

ashu3103:
Restore is working with base backups though

Jubilee101:
> No such file or directory

Hmm, that is odd. And postgresql.conf is present in your backup 20250411011348?

ashu3103:
> > No such file or directory
> 
> Hmm, that is odd. And postgresql.conf is present in your backup 20250411011348?

`postgresql.auto.conf` is present.

Jubilee101:
> postgresql.auto.conf is present.

Does your postgres server have postgres.conf in the first place? This feels off, I'm not 100% sure, but postgres.conf seems to be a must-have. I'm curious why you don't have it.

ashu3103:
> > postgresql.auto.conf is present.
> 
> Does your postgres server have postgres.conf in the first place? This feels off, I'm not 100% sure, but postgres.conf seems to be a must-have. I'm curious why you don't have it.

The `postgres.conf` is not directly under the data directory, its under `/etc/postgresql` directory.

ashu3103:
> @ashu3103 How did you do the `initdb` ?

I really can't recall how I did it initially, I am initializing new cluster `initdb -D </path/to/data/directory>`.

jesperpedersen:
@ashu3103 You forgot `-k`

jesperpedersen:
@ashu3103 No, we work towards production level installations - the documentation should reflect that as a minimum

ashu3103:
> ntation

Yeah, got your point!

src/libpgmoneta/restore.c
```diff
@@ -818,6 +831,99 @@ pgmoneta_combine_backups(int server, char* label, char* base, char* input_dir, c
...
+         {
+            pgmoneta_log_error("Unable to find block length for %s", bare_file_name);
+            goto error;
+         }
+         file_size = block_length * config->common.servers[server].block_size;
+         rfile_destroy(rf);
+         free(relative_path);
+         free(bare_file_name);
+      }
+      else // for non-incremental files get the file size from manifest itself
```
jesperpedersen, line 901: Include comment as `/* ... */` inside the code block

src/libpgmoneta/restore.c
```diff
@@ -1793,6 +1899,82 @@ incremental_rfile_initialize(int server, char* label, char* relative_dir, char*
    return 1;
 }
 
+static int
+split_file_path(char* path, char** relative_path, char** bare_file_name)
```
jesperpedersen, line 1903: All variables in the beginning of the function, and an `error` label to goto.

Clean up and return should be last the every function

src/libpgmoneta/restore.c
```diff
@@ -1793,6 +1899,82 @@ incremental_rfile_initialize(int server, char* label, char* relative_dir, char*
...
+   char* rel_path = NULL;
+   char* file_name = NULL;
+
+   if (path == NULL)
+   {
+      return 1;
+   }
+
+   char* path_copy = NULL;
+   path_copy = (char*)calloc(strlen(path) + 1, sizeof(char));
```
jesperpedersen, line 1917: pgmoneta_append

src/libpgmoneta/restore.c
```diff
@@ -1793,6 +1899,82 @@ incremental_rfile_initialize(int server, char* label, char* relative_dir, char*
...
+   if (path_copy == NULL)
+   {
+      return 1;
+   }
+   strcpy(path_copy, path);
```
jesperpedersen, line 1922: No

jesperpedersen (changes requested):

jesperpedersen:
Apart from `split_file_path` it looks ok

src/include/restore.h
```diff
@@ -117,6 +117,21 @@ pgmoneta_rollup_backups(int server, char* newest_label, char* oldest_label);
 int
 pgmoneta_extract_incremental_backup(int server, char* label, char** root, char** base);
 
+/**
```
jesperpedersen, line 120: Put this in info.h|.c

jesperpedersen: No, pgmoneta_backup_size

jesperpedersen (changes requested):

src/include/restore.h
ashu3103: The issue is I am using `incremental_rfile_initialize` inside this function which is static to `restore.c` translation unit, can you suggest a way to do it?

ashu3103: Either we can expose the rfile APIs in a seperate file `rfile.h` maybe?

jesperpedersen:
The part I don't like is this - https://github.com/pgmoneta/pgmoneta/pull/590/files#diff-260d37d8032cef06628f94f9c00c8136dc5c5c19450e08b1e4d5fe179da3adeeR39

Including `restore.h` in `wf_backup.c` sounds wrong - so you have to untangle it somehow

ashu3103:
> The part I don't like is this - https://github.com/pgmoneta/pgmoneta/pull/590/files#diff-260d37d8032cef06628f94f9c00c8136dc5c5c19450e08b1e4d5fe179da3adeeR39
> 
> Including `restore.h` in `wf_backup.c` sounds wrong - so you have to untangle it somehow

Yeah that doesn't look good to me too :). Okay I'll try to find some workaround.

@Jubilee101, thoughts?

src/include/restore.h
Jubilee101: I would keep rfile in `restore.c`. you only need it because you want the reconstructed file size. So expose a pgmoneta_restored_file_size() maybe.

jesperpedersen: I don't know what that means - I just don't want to see "restore" in a "backup" file.

Layer the functionality...

Jubilee101: > I don't know what that means - I just don't want to see "restore" in a "backup" file.

Fair point. Let's expose `restored_file_size()` then. If we are going to need more we can move `rfile` to `info.h`. Hopefully we don't have to.

ashu3103: Okay, so will it be okay to import `<restore.h>` in `info.c`? because we want our `pgmoneta_restored_file_size()` there

jesperpedersen: No

ashu3103: @Jubilee101, Then I was thinking of making `rfile` a separate module (in `rfile.h`) exposing all the related APIs like `rfile_create`, `rfile_destroy`, `rfile_destroy_cb`, `incremental_rfile_initialize` etc.

This module (`rfile.h`) will be imported in `info.c` (is this import ok? @jesperpedersen) and `restore.c`.

Jubilee101: rfile isn't needed anywhere else so we are not supposed to expose it. All you need is trucation block length, so let's do a native one without rfile inside info.c. Open the file, read the value and the value only, close the file, done.

Jubilee101: Ah, no, it needs block numbers as well.. @jesperpedersen I don't see why it's not ok having restore.h included inside info.h. We'll have to move and expose the entire rfile infra to info - only to get a size of a file. I don't like it, it's worth doing only for a file size calculation.

Jubilee101: > Then I was thinking of making rfile a separate module (in rfile.h) exposing all the related APIs like rfile_create, rfile_destroy, rfile_destroy_cb, incremental_rfile_initialize etc.

@ashu3103  Let's expose them then. Jesper's got a point, that meta data belongs to info.h... You don't need a new module. Putting them inside info.h is enough. Keep `rfile_destroy_cb` inside restore.c.

jesperpedersen:
It is about layers. We don't need rfile exposed.

Backup shouldn't include restore - it is called `get_backup_size` to stick it in `info.h` and move all the functionality that is necessary to there. `info.h` is the main file for information about a backup

Jubilee101:
> It is about layers. We don't need rfile exposed.

> I mean a dedicated rfile.h - you can have input/output parameters of course

Yeah, no need to do rfile.h. Maybe we'll have a separate incremental.h in the future when things get even bigger. For the moment Let's keep things inside info.

jesperpedersen:
Agreed.

Even though our `.h` are really internal APIs they should be as small as possible, and as simple and clear as possible. It is for developers to use to help expand our functionality, so each function need to be in the correct place

src/include/restore.h
ashu3103: I don't feel the need, but ideally this task deserves a separate issue, I'll do as you suggest.....

ashu3103:
@jesperpedersen, @Jubilee101

The `rfile_create()` internally uses `pgmoneta_extract_backup_file()` and `pgmoneta_file_finalname()` which are defined in `backup.h` so that will create a dependency of `info.c` on `backup.h`.

Jubilee101:
pgmoneta_file_finalname() can just be in utils.c, as for pgmoneta_extract_backup_file(), this is more of a restore thing... @jesperpedersen thoughts?

ashu3103:
> pgmoneta_file_finalname() can just be in utils.c, 

Okay!

> as for pgmoneta_extract_backup_file(), this is more of a restore thing... @jesperpedersen thoughts?

If we place `pgmoneta_extract_backup_file` in `restore.h` info.h will now have dependency on `restore.h`. Can't we also move this in utils, or in `info.h` itself?

jesperpedersen:
`pgmoneta_file_finalname` - what is that ? It sound static to me. I don't like the name.

`pgmoneta_extract_backup_file` sounds like `restore.h`.

We have to break it down, and maybe the higher level API is just a wrapper for a lower level one.

What are the core functions ? And, move them down, so we don't have dependencies in both directions

Jubilee101:
> pgmoneta_file_finalname - what is that ? It sound static to me. I don't like the name.

And it shall be. @ashu3103 Can you rebase? Sorry for the mess.

> Can't we also move this in utils, or in info.h itself?

Yes, let's keep it as a static function inside info.c at the moment. It's not used elsewhere anyway. Jesper and I will sort out this circular dependency nightmare in another issue.

ashu3103:
> > pgmoneta_file_finalname - what is that ? It sound static to me. I don't like the name.
> 
> And it shall be. @ashu3103 Can you rebase? Sorry for the mess.
> 

Sure

> > Can't we also move this in utils, or in info.h itself?
> 
> Yes, let's keep it as a static function inside info.c at the moment. It's not used elsewhere anyway. Jesper and I will sort out this circular dependency nightmare in another issue.

I think it will still pose some issues (if you define `pgmoneta_extract_backup_file` static) because this function is alos used in `restore.c` in function `copy_backup_file`. So defining it static may not be a good idea we have to expose it in `info.c`

Jubilee101:
> I think it will still pose some issues (if you define pgmoneta_extract_backup_file static) because this function is alos used in restore.c in function copy_backup_file. So defining it static may not be a good idea we have to expose it in info.c

Do what you must. We can do bigger refactoring work later.

Jubilee101:
> Also I had one doubt regarding your latest commit, you have made the functions pgmoneta_file_finalname (now file_final_name) static to restore.c but if we were anyways shifting rfile_create() in info.c, so shall I change that too.

Ok, what about moving it to a static function in info?

ashu3103:
> > Also I had one doubt regarding your latest commit, you have made the functions pgmoneta_file_finalname (now file_final_name) static to restore.c but if we were anyways shifting rfile_create() in info.c, so shall I change that too.
> 
> Ok, what about moving it to a static function in info?

Yeah that would be fine, since its only used in `rfile_create` and we are moving that to `info.c`

ashu3103:

src/include/info.h
```diff
@@ -298,6 +326,52 @@ pgmoneta_info_request(SSL* ssl, int client_fd, int server, uint8_t compression,
 void
 pgmoneta_annotate_request(SSL* ssl, int client_fd, int server, uint8_t compression, uint8_t encryption, struct json* payload);
 
+/**
+ * 
+ */
+int
+rfile_create(int server, char* label, char* relative_dir, char* base_file_name, int encryption, int compression, struct rfile** rfile);
```
Jubilee101, line 333: Add `pgmoneta_` prefix since you exposed them

src/libpgmoneta/info.c
```diff
@@ -27,6 +27,8 @@
  */
 
 /* pgmoneta */
+#include "aes.h"
```
Jubilee101, line 30: <> instead of ""

src/libpgmoneta/info.c
```diff
@@ -27,6 +27,8 @@
...
+#include "compression.h"
 #include <pgmoneta.h>
```
Jubilee101, line 32: Always put pgmoneta.h on top

src/include/info.h
```diff
@@ -298,6 +326,52 @@ pgmoneta_info_request(SSL* ssl, int client_fd, int server, uint8_t compression,
...
+
+
+/**
```
Jubilee101, line 336: comments?

src/include/info.h
```diff
@@ -298,6 +326,52 @@ pgmoneta_info_request(SSL* ssl, int client_fd, int server, uint8_t compression,
...
+ * 
+ */
```
Jubilee101, line 338: comments?

src/include/info.h
```diff
@@ -298,6 +326,52 @@ pgmoneta_info_request(SSL* ssl, int client_fd, int server, uint8_t compression,
...
+void
+rfile_destroy(struct rfile* rf);
+
+/**
+ *
```
Jubilee101, line 343: comments?

Jubilee101 (changes requested):

src/include/info.h
ashu3103: Oops totally forgot!

ashu3103:
@Jubilee101, @jesperpedersen PTAL

src/include/info.h
```diff
@@ -298,6 +327,68 @@ pgmoneta_info_request(SSL* ssl, int client_fd, int server, uint8_t compression,
...
+ * number of blocks in the fully restored file). For non-incremental files, simply use the size value 
+ * directly from the file entry in the manifest.
+ * @param server The server
+ * @param label The label
+ * @param size [out] The size of the incremental backup
+ * @param biggest_file_size [out] The size of the biggest file in the incremental backup
+ * @return 0 on success, 1 if otherwise
+ */
+int
+pgmoneta_backup_size(int server, char* label, unsigned long* size, uint64_t* biggest_file_size);
```
jesperpedersen, line 390: This should be the only change in info.h - so you have to use static for the rest, or a lower level .h

jesperpedersen (changes requested):

src/include/info.h
ashu3103: We discussed about exposing rfile and corresponding APIs in `info.h` right, please tell me if I am missing something.

jesperpedersen: info.h is about the `struct backup` - so if you can't use `rfile` as static only, then you need to create a `rfile.h` file and we can take it from there

Jubilee101: This doesn't change the fact that extract_backup belongs to restore.h. And if you don't want info to include restore, it's a dilemma. The patch looks good enough for me. And I don't see why rfile can't belong to info. It's still header metadata about backup file. Even if you put rfile in rfile.h, you'd still have extract_backup_file that belongs to nowhere. I'd like to merge this patch, and think about how we can rewire things in #592.

jesperpedersen: Ok, go ahead then

Jubilee101:
@ashu3103 It doesn't work when hotstandby is enabled. Free space check is checking the temp directory in workspace that has not been created yet. We use it to store the restored backup temporarily.

src/include/info.h
```diff
@@ -298,6 +327,68 @@ pgmoneta_info_request(SSL* ssl, int client_fd, int server, uint8_t compression,
 void
 pgmoneta_annotate_request(SSL* ssl, int client_fd, int server, uint8_t compression, uint8_t encryption, struct json* payload);
 
+/**
+ * Create an rfile structure of an incremental file
```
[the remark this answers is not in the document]
Jubilee101: LGTM other than this comment @jesperpedersen, though something is wrong with CI.

Jubilee101:

PR #565, opened 2025-04-02
[#557] Workers: Use Deque

dancingactor:
implement #557 

This PR replace the custom queue in `workers.c` with the existing deque in `deque.c` implementation.

## Details
1. The task is now included into `worker_common`, instead of being wrapped in a separate `struct task`. Since the deque is designed to hold generic values (with cleanup via value_config), storing the complete task in a single structure simplifies both insertion and memory management.
2. Move `struct semaphore* has_tasks;` from original queue to workers since deque doesn't have semaphore logic.

Issue #557, opened 2025-04-01
Workers: Use deque


jesperpedersen:
Instead of having a custom list in `workers.h` we can reimplement it with `deque.h`

dancingactor:
@jesperpedersen I would like to take this

dancingactor:
@jesperpedersen Do I need to write a test code to test workers functionality?

jesperpedersen:
@dancingactor This isn't working. Try `workers = 8` and do 2 backups with compression and encryption using clang/Debug

dancingactor:
Got it. Let me look into this

dancingactor:
I'm stuck on this issue. Could you provide some hints or ideas on why it's happening, I see it's a heap-buffer-overflow problem
This is the log
```
dancing@vbox:~/pgmoneta/build/src$ ./pgmoneta -c /etc/pgmoneta/pgmoneta.conf -u /var/lib/pgmoneta_users.conf
=================================================================
==4020==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x50200060c8e0 at pc 0x557ff274bc5f bp 0x7fba485ff790 sp 0x7fba485fef50
READ of size 3 at 0x50200060c8e0 thread T36
    #0 0x557ff274bc5e  (/home/dancing/pgmoneta/build/src/pgmoneta+0x63c5e) (BuildId: 10ea1d4ddd2d6364bbc7a264fac2a9d356ca8990)
    #1 0x7fba4ea06cda  (/home/dancing/pgmoneta/build/src/libpgmoneta.so.0+0x6cda) (BuildId: 5d4bd21cc33df02a4cf24424db49f1643b24e858)
    #2 0x7fba4ea05eab  (/home/dancing/pgmoneta/build/src/libpgmoneta.so.0+0x5eab) (BuildId: 5d4bd21cc33df02a4cf24424db49f1643b24e858)
    #3 0x7fba4eba9ee6  (/home/dancing/pgmoneta/build/src/libpgmoneta.so.0+0x1a9ee6) (BuildId: 5d4bd21cc33df02a4cf24424db49f1643b24e858)
    #4 0x557ff278725a  (/home/dancing/pgmoneta/build/src/pgmoneta+0x9f25a) (BuildId: 10ea1d4ddd2d6364bbc7a264fac2a9d356ca8990)
[...]

Issue #523, opened 2025-03-14
Automatic offline mode transition


Jubilee101:
Pgmoneta should switch to offline mode automatically, if all servers go down. 

An idea is to monitor WAL streaming, if the streaming stops for all servers, we switch to offline mode.

Ideally we should also have a command to bring it back to online mode when server is up again.

dancingactor:
Hi @Jubilee101 @jesperpedersen , I would like to take this issue

jesperpedersen:
@dancingactor Go for it !

dancingactor:
Hi @jesperpedersen @Jubilee101,

I would like to clarify a few things about this project before proceeding:

### Objectives
- Save resources  
- Reduce unexpected failures when PostgreSQL crashes  
- Reduce error log noise

### Question
- Should recovery to online mode be **manual**, **automatic**, or support **both**?

### Next Steps
1. Identify which operations are active/inactive in offline mode  
2. Define the best way to determine server status (likely via WAL streaming as mentioned)  
3. Decide how long to wait before declaring all servers down

Appreciate your feedback and happy to discuss further!

jesperpedersen:
@dancingactor It should be automatic to detect a server is offline, but I think that the CLI need 2 commands `online` and `offline` such that the DBA can put a server online or offline

dancingactor:
Got it. So I think there are two tasks:
1. add `online` and `offline` commands
2. Automatic offline mode transition

Can I open a new issue about adding `online` and `offline` commands

jesperpedersen:
Just do it as part of this issue - they belong together

dancingactor:
Understood

jesperpedersen:
Yes, because it doesn't connect to any servers during startup then we can bring servers back online with the CLI command

dancingactor:
I'm back to this issue!

dancingactor:
> Yes, because it doesn't connect to any servers during startup then we can bring servers back online with the CLI command

Hi @jesperpedersen , I still don't get it why we need `--offline` option if we have `online` and `offline` command
For example, if we used to run command like 
```
pgmoneta-cli --offline encrypt /tmp/myfile
```
now we can run
```
pgmoneta-cli offline
pgmoneta-cli encrypt /tmp/myfile
```

jesperpedersen:
* Boot PostgreSQL
* Boot pgmoneta
* Take a backup
* Shutdown pgmoneta
* Shutdown PostgreSQL
[...]

PR #596, opened 2025-04-14
[#557] Workers: Use Deque

dancingactor:
implement  https://github.com/pgmoneta/pgmoneta/issues/557
fix PR https://github.com/pgmoneta/pgmoneta/pull/565

This PR replace the custom queue in workers.c with the existing deque in deque.c implementation. Backup successfully with workers = 8

## Updates
Comparing to the original code, I’ve made several updates
1. **change semaphore value from binary to counting** -> This prevents a situation where worker threads go to sleep even though tasks are still available, which can happen if tasks are added faster than they’re polled from the deque.
2. **seperate the `worker_task` and `worker_common`** to avoid freeing them both when we want to free the task
3. **add `pgmoneta_workers_wait` to `pgmoneta_workers_destroy`**

### DIFF
--- a/src/include/workers.h
+++ b/src/include/workers.h
@@ -34,6 +34,7 @@ extern "C" {
 #endif
 
 #include <pgmoneta.h>
+#include <deque.h>
 
 #include <pthread.h>
 #include <stdio.h>
@@ -48,29 +49,17 @@ struct semaphore
 {
    pthread_mutex_t mutex; /**< The mutex of the semaphore */
    pthread_cond_t cond;   /**< The condition of the semaphore */
-   int value;             /**< The current value */
+   int count;             /**< The current count */
 };
 
-/** @struct task
- * Defines a task
+/** @struct worker_task
+ * Defines a worker task
  */
-struct task
+struct worker_task
 {
-   struct task* previous;                  /**< The previous task */
-   void (*function)(struct worker_common*); /**< The task */
-   struct worker_common* wc;                /**< The common base */
-};
-
-/** @struct queue
- * Defines a queue
- */
-struct queue
-{
-   pthread_mutex_t rwmutex;     /**< The read/write mutex */
-   struct task* front;          /**< The first task */
-   struct task* rear;           /**< The last task */
-   struct semaphore* has_tasks; /**< Are there any tasks ? */
-   int number_of_tasks;         /**< The number of tasks */
+   void (*function)(struct worker_common*); /**< The task function */
+   struct worker_common* wc;                /**< Pointer to the common data */
+   char* name;                              /**< The name of the function */
 };
 
 /** @struct worker
@@ -93,7 +82,8 @@ struct workers
    pthread_mutex_t worker_lock;    /**< The worker lock */
    pthread_cond_t worker_all_idle; /**< Are workers idle */
    bool outcome;                   /**< Outcome of the workers */
-   struct queue queue;             /**< The queue */
+   struct deque* queue;            /**< The task queue using deque */
+   struct semaphore* has_tasks;    /**< Semaphore for tasks dispatching*/
 };
 
 /** @struct worker_common
@@ -132,11 +122,11 @@ pgmoneta_workers_initialize(int num, struct workers** workers);
  * Add work to the queue
  * @param workers The workers
  * @param function The function pointer
- * @param wi The argument
+ * @param wc The argument
  * @return 0 upon success, otherwise 1.
  */
 int
-pgmoneta_workers_add(struct workers* workers, void (*function)(struct worker_common*), struct worker_common* wi);
+pgmoneta_workers_add(struct workers* workers, void (*function)(struct worker_common*), struct worker_common* wc);
 
 /**
  * Wait for all queued work units to finish
--- a/src/libpgmoneta/wf_hot_standby.c
+++ b/src/libpgmoneta/wf_hot_standby.c
@@ -387,6 +387,8 @@ error:
    free(old_manifest);
    free(new_manifest);
 
+   pgmoneta_workers_destroy(workers);
+
    if (source_root != NULL)
    {
       pgmoneta_delete_directory(source_root);
--- a/src/libpgmoneta/workers.c
+++ b/src/libpgmoneta/workers.c
@@ -27,8 +27,10 @@
  */
 
 #include <pgmoneta.h>
+#include <deque.h>
 #include <logging.h>
 #include <workers.h>
+#include <value.h>
 
 #include <errno.h>
 #include <signal.h>
@@ -44,17 +46,11 @@ static int worker_init(struct workers* workers, struct worker** worker);
 static void* worker_do(struct worker* worker);
 static void worker_destroy(struct worker* worker);
 
-static int queue_init(struct queue* queue);
-static void queue_clear(struct queue* queue);
-static void queue_push(struct queue* queue, struct task* task);
-static struct task* queue_pull(struct queue* queue);
-static void queue_destroy(struct queue* queue);
-
-static int semaphore_init(struct semaphore* semaphore, int value);
-static void semaphore_reset(struct semaphore* semaphore);
+static int semaphore_init(struct semaphore* semaphore);
 static void semaphore_post(struct semaphore* semaphore);
 static void semaphore_post_all(struct semaphore* semaphore);
 static void semaphore_wait(struct semaphore* semaphore);
+static void destroy_task_wrapper(uintptr_t data);
 
 int
 pgmoneta_workers_initialize(int num, struct workers** workers)
@@ -81,9 +77,22 @@ pgmoneta_workers_initialize(int num, struct workers** workers)
    w->number_of_working = 0;
    w->outcome = true;
 
-   if (queue_init(&w->queue))
+   if (pgmoneta_deque_create(true, &w->queue))
    {
-      pgmoneta_log_error("Could not allocate memory for queue");
+      pgmoneta_log_error("Could not allocate memory for deque");
+      goto error;
+   }
+
+   w->has_tasks = (struct semaphore*)malloc(sizeof(struct semaphore));
+   if (w->has_tasks == NULL)
+   {
+      pgmoneta_log_error("Could not allocate memory for workers semaphore");
+      goto error;
+   }
+
+   if (semaphore_init(w->has_tasks))
+   {
+      pgmoneta_log_error("Could not initialize workers semaphore");
       goto error;
    }
 
@@ -115,7 +124,8 @@ error:
 
    if (w != NULL)
    {
-      queue_destroy(&w->queue);
+      pgmoneta_deque_destroy(w->queue);
+      free(w->has_tasks);
       free(w);
    }
 
@@ -125,21 +135,27 @@ error:
 int
 pgmoneta_workers_add(struct workers* workers, void (*function)(struct worker_common*), struct worker_common* wc)
 {
-   struct task* t = NULL;
+   struct worker_task* task = NULL;
+   struct value_config config = {0};
 
    if (workers != NULL)
    {
-      t = (struct task*)malloc(sizeof(struct task));
-      if (t == NULL)
+      task = (struct worker_task*)malloc(sizeof(struct worker_task));
+      if (task == NULL)
       {
          pgmoneta_log_error("Could not allocate memory for task");
          goto error;
       }
 
-      t->function = function;
-      t->wc = wc;
+      task->function = function;
+      task->wc = wc;
+      task->name = NULL;
 
-      queue_push(&workers->queue, t);
+      config.destroy_data = destroy_task_wrapper;
+
+      pgmoneta_deque_add_with_config(workers->queue, NULL, (uintptr_t)task, &config);
+
+      semaphore_post(workers->has_tasks);
 
       return 0;
    }
@@ -156,7 +172,7 @@ pgmoneta_workers_wait(struct workers* workers)
    {
       pthread_mutex_lock(&workers->worker_lock);
 
-      while (workers->queue.number_of_tasks || workers->number_of_working)
+      while (pgmoneta_deque_size(workers->queue) > 0 || workers->number_of_working > 0)
       {
          pthread_cond_wait(&workers->worker_all_idle, &workers->worker_lock);
       }
@@ -182,18 +198,19 @@ pgmoneta_workers_destroy(struct workers* workers)
       time (&start);
       while (tpassed < timeout && workers->number_of_alive)
       {
-         semaphore_post_all(workers->queue.has_tasks);
+         semaphore_post_all(workers->has_tasks);
          time(&end);
          tpassed = difftime(end, start);
       }
 
       while (workers->number_of_alive)
       {
-         semaphore_post_all(workers->queue.has_tasks);
+         semaphore_post_all(workers->has_tasks);
          SLEEP(1000000000L);
       }
 
-      queue_destroy(&workers->queue);
+      pgmoneta_deque_destroy(workers->queue);
+      free(workers->has_tasks);
 
       for (int n = 0; n < worker_total; n++)
       {
@@ -309,8 +326,7 @@ error:
 static void*
 worker_do(struct worker* worker)
 {
-   void (*func_ref)(struct worker_common*);
-   struct task* t;
+   struct worker_task* task;
    struct workers* workers = worker->workers;
 
    pthread_mutex_lock(&workers->worker_lock);
@@ -319,7 +335,7 @@ worker_do(struct worker* worker)
 
    while (worker_keepalive)
    {
-      semaphore_wait(workers->queue.has_tasks);
+      semaphore_wait(workers->has_tasks);
 
       if (worker_keepalive)
       {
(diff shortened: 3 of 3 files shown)


jesperpedersen (changes requested):

jesperpedersen:
Maybe add some TRACE logging when adding and finishing a task - and showing how many tasks are running

src/include/workers.h
```diff
@@ -93,7 +81,8 @@ struct workers
    pthread_mutex_t worker_lock;    /**< The worker lock */
    pthread_cond_t worker_all_idle; /**< Are workers idle */
    bool outcome;                   /**< Outcome of the workers */
-   struct queue queue;             /**< The queue */
+   struct deque* queue;       /**< The task queue using deque */
+   struct semaphore* has_tasks;    /**< Are there any tasks ? */
```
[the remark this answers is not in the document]
jesperpedersen: that comments are at the same position like
```
   /**< X */
   /**< Y */
```

Not
```
 /**< X */
   /**< Y */
```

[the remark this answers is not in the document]
dancingactor: > Can we use deque_size ?
Sorry, I meant to ask: what does deque_size mean?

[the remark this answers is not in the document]
jesperpedersen: https://github.com/pgmoneta/pgmoneta/blob/main/src/include/deque.h#L241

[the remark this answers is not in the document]
dancingactor: Do you mean replace `semaphore` with `pgmoneta_deque_size`?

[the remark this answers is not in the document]
jesperpedersen: Yes, if it is possible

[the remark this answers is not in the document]
dancingactor: I don't think it can be changed, the `semaphore` in the `workers` struct is essential for dispatching tasks to workers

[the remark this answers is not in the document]
jesperpedersen: Maybe expand the documentation then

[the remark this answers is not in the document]
dancingactor: You're right, let me do that

src/libpgmoneta/workers.c
```diff
@@ -176,24 +190,27 @@ pgmoneta_workers_destroy(struct workers* workers)
 
    if (workers != NULL)
    {
+      pgmoneta_workers_wait(workers);
```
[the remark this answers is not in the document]
dancingactor: But other code like [restore.c](https://github.com/pgmoneta/pgmoneta/blob/main/src/libpgmoneta/restore.c#L1125) and [wf_hot_standby.c](https://github.com/pgmoneta/pgmoneta/blob/main/src/libpgmoneta/wf_hot_standby.c#L308) will still use `_wait` independently(not with `_destroy`)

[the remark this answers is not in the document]
dancingactor: But `_destroy` and `_wait` serve different purposes, so I think we should keep both public API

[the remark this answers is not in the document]
jesperpedersen: Ok, lets keep them separate then.

However, the code in `wf_hot_standby.c` is buggy so that need to be fixed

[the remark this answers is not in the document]
dancingactor: got it. Thanks for your time! I will investigate it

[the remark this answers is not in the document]
dancingactor: > However, the code in wf_hot_standby.c is buggy so that need to be fixed

Is the mistake that `_destroy` is not called in the `error:` section?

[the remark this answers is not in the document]
jesperpedersen: It should always be safe to call `_destroy` even if `workers` is `NULL`, so yes - but also the success path...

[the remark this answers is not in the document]
dancingactor: I think success path doesn't have a bug

The snippet at [wf_hot_standby.c](https://github.com/pgmoneta/pgmoneta/blob/main/src/libpgmoneta/wf_hot_standby.c#L306-L313) doesn't call `_destroy` because `workers` is still needed for the upcoming `pgmoneta_copy_directory()` at [line 322](https://github.com/pgmoneta/pgmoneta/blob/main/src/libpgmoneta/wf_hot_standby.c#L322). The call  to `_destroy` is correctly placed later at [line 355](https://github.com/pgmoneta/pgmoneta/blob/main/src/libpgmoneta/wf_hot_standby.c#L335).

src/libpgmoneta/wf_gzip.c
```diff
@@ -264,7 +264,6 @@ gzip_execute_uncompress(char* name __attribute__((unused)), struct art* nodes)
 
    if (number_of_workers > 0)
    {
-      pgmoneta_workers_wait(workers);
```
jesperpedersen, line 267: We should keep this then

src/libpgmoneta/wf_hot_standby.c
```diff
@@ -387,6 +387,10 @@ hot_standby_execute(char* name __attribute__((unused)), struct art* nodes)
    free(old_manifest);
    free(new_manifest);
 
+   if (number_of_workers > 0) {
```
jesperpedersen, line 390: We shouldn't need the `if`

src/libpgmoneta/wf_lz4.c
```diff
@@ -262,7 +262,6 @@ lz4_execute_uncompress(char* name __attribute__((unused)), struct art* nodes)
 
    if (number_of_workers > 0)
    {
-      pgmoneta_workers_wait(workers);
```
jesperpedersen, line 265: Keep

src/libpgmoneta/workers.c
```diff
@@ -81,9 +77,22 @@ pgmoneta_workers_initialize(int num, struct workers** workers)
...
+   if (pgmoneta_deque_create(true, &w->queue))
+   {
+      pgmoneta_log_error("Could not allocate memory for deque");
+      goto error;
+   }
+
+   w->has_tasks = (struct semaphore*)malloc(sizeof(struct semaphore));
+   if (w->has_tasks == NULL)
+   {
+      pgmoneta_log_error("Could not allocate memory for task semaphore");
```
jesperpedersen, line 89: task -> workers

src/libpgmoneta/workers.c
```diff
@@ -81,9 +77,22 @@ pgmoneta_workers_initialize(int num, struct workers** workers)
...
+      goto error;
+   }
+
+   if (semaphore_init(w->has_tasks, 0))
    {
-      pgmoneta_log_error("Could not allocate memory for queue");
+      pgmoneta_log_error("Could not initialize task semaphore");
```
jesperpedersen, line 95: task -> workers

src/libpgmoneta/workers.c
```diff
@@ -125,21 +135,27 @@ pgmoneta_workers_initialize(int num, struct workers** workers)
...
-      t->function = function;
-      t->wc = wc;
+      task->function = function;
+      task->wc = wc;
 
-      queue_push(&workers->queue, t);
+      struct value_config config = {0};
+      config.destroy_data = destroy_task_wrapper;
+
+      pgmoneta_log_trace("Adding task to queue");
```
jesperpedersen, line 155: Can we add more information about what was added ?

src/libpgmoneta/workers.c
```diff
@@ -156,8 +172,9 @@ pgmoneta_workers_wait(struct workers* workers)
    {
       pthread_mutex_lock(&workers->worker_lock);
 
-      while (workers->queue.number_of_tasks || workers->number_of_working)
+      while (pgmoneta_deque_size(workers->queue) || workers->number_of_working)
```
jesperpedersen, line 175: Use `> 0` to make it explicit

src/libpgmoneta/workers.c
```diff
@@ -319,21 +338,21 @@ worker_do(struct worker* worker)
...
-            func_ref = t->function;
-            func_ref(t->wc);
+         task = (struct worker_task*)pgmoneta_deque_poll(workers->queue, NULL);
 
-            free(t);
+         if (task)
+         {
+            task->function(task->wc);
+            free(task);
+            pgmoneta_log_trace("Finished task");
```
jesperpedersen, line 355: Move up a bit and add a bit more information

src/libpgmoneta/workers.c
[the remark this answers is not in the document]
jesperpedersen: It shouldn't be necessary to call `_wait` inside `_destroy` if it is part of the public API

src/libpgmoneta/workers.c
```diff
@@ -511,10 +420,17 @@ static void
 semaphore_wait(struct semaphore* semaphore)
 {
    pthread_mutex_lock(&semaphore->mutex);
-   while (semaphore->value != 1)
+   while (semaphore->value == 0)
```
jesperpedersen, line 423: Is this correct ?

src/libpgmoneta/workers.c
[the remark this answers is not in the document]
dancingactor: got it, so I sperate them

dancingactor:
@jesperpedersen By the way, I really appreciate your time and feeback. you literally make my code better, and I have learned so much from it. I really appreciate your help, especially knowing you’re doing this with no pay and just to support the community.

jesperpedersen (changes requested):

src/libpgmoneta/workers.c
```diff
@@ -115,31 +124,39 @@ pgmoneta_workers_initialize(int num, struct workers** workers)
...
+      free(w->has_tasks);
       free(w);
    }
 
    return 1;
 }
 
 int
-pgmoneta_workers_add(struct workers* workers, void (*function)(struct worker_common*), struct worker_common* wc)
+pgmoneta_workers_add(struct workers* workers, void (*function)(struct worker_common*), struct worker_common* wc, const char* function_name)
```
jesperpedersen, line 136: Remove `const`

src/libpgmoneta/workers.c
```diff
@@ -115,31 +124,39 @@ pgmoneta_workers_initialize(int num, struct workers** workers)
...
-      t->wc = wc;
+      task->function = function;
+      task->wc = wc;
+      task->name = function_name;
 
-      queue_push(&workers->queue, t);
+      struct value_config config = {0};
+      config.destroy_data = destroy_task_wrapper;
+
+      pgmoneta_log_trace("Adding task to queue: %s", function_name);
```
jesperpedersen, line 156: Maybe add the task pointer too so `%s/%p`

src/libpgmoneta/workers.c
```diff
@@ -319,21 +337,21 @@ worker_do(struct worker* worker)
...
-         {
-            func_ref = t->function;
-            func_ref(t->wc);
+         task = (struct worker_task*)pgmoneta_deque_poll(workers->queue, NULL);
 
-            free(t);
+         if (task)
+         {
+            task->function(task->wc);
+            pgmoneta_log_trace("Finished task: %s", task->name);
```
jesperpedersen, line 353: Same here `%s/%p`

jesperpedersen (changes requested):

src/libpgmoneta/workers.c
```diff
@@ -115,31 +124,39 @@ pgmoneta_workers_initialize(int num, struct workers** workers)
...
-      t->wc = wc;
+      task->function = function;
+      task->wc = wc;
+      task->name = function_name;
 
-      queue_push(&workers->queue, t);
+      struct value_config config = {0};
+      config.destroy_data = destroy_task_wrapper;
+
+      pgmoneta_log_trace("Adding task to queue: %s (%p)", function_name, function);
```
jesperpedersen, line 156: Remove - it doesn't add anything really

src/libpgmoneta/workers.c
```diff
@@ -319,21 +337,21 @@ worker_do(struct worker* worker)
...
-         {
-            func_ref = t->function;
-            func_ref(t->wc);
+         task = (struct worker_task*)pgmoneta_deque_poll(workers->queue, NULL);
 
-            free(t);
+         if (task)
+         {
+            task->function(task->wc);
+            pgmoneta_log_trace("Finished task: %s (%p)", task->name, task->function);
```
jesperpedersen, line 353: Remove - it doesn't really add anything

src/libpgmoneta/workers.c
```diff
@@ -360,119 +378,17 @@ worker_destroy(struct worker* w)
 }
 
 static int
-queue_init(struct queue* queue)
+semaphore_init(struct semaphore* semaphore, int count)
```
jesperpedersen, line 381: Remove `int count` and init to `0`

jesperpedersen (changes requested):

jesperpedersen:
Please, look at `pgmoneta_workers_destroy` - it need A LOT of work

src/libpgmoneta/workers.c
```diff
@@ -115,31 +124,39 @@ pgmoneta_workers_initialize(int num, struct workers** workers)
...
       }
 
-      t->function = function;
-      t->wc = wc;
+      task->function = function;
+      task->wc = wc;
+      task->name = function_name;
 
-      queue_push(&workers->queue, t);
+      struct value_config config = {0};
```
jesperpedersen, line 153: Move to the beginning of the function

jesperpedersen (changes requested):

jesperpedersen:
Try and see if you can make `workers->number_of_alive` an atomic, then in *some* (not *all*) can remove the mutex locking

jesperpedersen:
And, `workers->number_of_working`

jesperpedersen:
Make `semaphore_init` an output parameter

dancingactor:
> Try and see if you can make `workers->number_of_alive` an atomic, then in _some_ (not _all_) can remove the mutex locking

Sure. but could I ask the reason, I am curious, Thanks

src/libpgmoneta/workers.c
jesperpedersen: Remove the TRACE

src/libpgmoneta/workers.c
jesperpedersen: Sure

src/include/workers.h
```diff
@@ -48,29 +49,17 @@ struct semaphore
...
-struct queue
-{
-   pthread_mutex_t rwmutex;     /**< The read/write mutex */
-   struct task* front;          /**< The first task */
-   struct task* rear;           /**< The last task */
-   struct semaphore* has_tasks; /**< Are there any tasks ? */
-   int number_of_tasks;         /**< The number of tasks */
+   void (*function)(struct worker_common*); /**< The task function */
+   struct worker_common* wc;                /**< Pointer to the common data */
+   const char* name;                        /**< The name of the function */
```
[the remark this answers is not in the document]
jesperpedersen: I don't like `const`

src/libpgmoneta/workers.c
jesperpedersen: Remove the TRACE

src/libpgmoneta/workers.c
jesperpedersen: Remove the TRACE

src/libpgmoneta/workers.c
jesperpedersen: Yes

src/libpgmoneta/workers.c
jesperpedersen: Remove the TRACE

jesperpedersen:
with atomic we don't need to take/release the mutex -- but it isn't super critical if you can't make it work

src/libpgmoneta/workers.c
dancingactor: But didn't you just tell me to add the log when adding and finishing the tasks

dancingactor:
> with atomic we don't need to take/release the mutex -- but it isn't super critical if you can't make it work

Got it, I will try it out

src/libpgmoneta/workers.c
dancingactor: https://github.com/pgmoneta/pgmoneta/pull/596#issuecomment-2812395278

jesperpedersen: Yes, but it didn't add anything useful in the end - so we can remove it

jesperpedersen:
Of course we can remove `function_name` too since we don't need it

jesperpedersen:
Maybe forward declare `struct workers` instead of `struct worker_common` because then `struct worker_common` can be the first `struct worker_X` defined

jesperpedersen:
We should do the standard function signature for all functions too, e.g.

```
int
pgmoneta_XYZ(input, output);
```

Since `_destroy()` should handle `NULL` it will always return `0`

jesperpedersen:
To make it easier just use `atomic_` all over

src/libpgmoneta/workers.c
dancingactor: got it

jesperpedersen:
@dancingactor Thanks, I'll work on it for a bit

dancingactor:
Sorry, I haven’t finished it yet. I’ll re-request the review again once I complete

jesperpedersen:
@dancingactor No worries, I'm working on your patch since we need it for 0.16

dancingactor:
> @dancingactor No worries, I'm working on your patch since we need it for 0.16

got it, so do I need to keep fixing the code

jesperpedersen:
@dancingactor I got it

dancingactor:
If there's anything I can help with, I am happy to do anytime!

jesperpedersen:
@dancingactor Feel free to find a task, and start on it

dancingactor:
Thanks! Now I am working on another issue https://github.com/pgmoneta/pgmoneta/issues/523

jesperpedersen:
@dancingactor Don't spend time on this - I have completely changed the API on my branch

dancingactor:
Oh okay, thanks for the heads-up!

jesperpedersen:
@dancingactor We are going with mostly your patch for now.

We can make improvements later

jesperpedersen:
Merged.

PR #598, opened 2025-04-15
temp

Mohab96:


Mohab96:
@jesperpedersen 

Can you please take a quick look at this and tell me whether I am on the right track or not?
Of course I will move everything out of `walinfo.c` when everything is good.

src/walinfo.c
```diff
@@ -92,6 +94,253 @@ usage(void)
...
+           h1->std.xlp_info == h2->std.xlp_info &&
+           h1->std.xlp_tli == h2->std.xlp_tli &&
+           h1->std.xlp_pageaddr == h2->std.xlp_pageaddr &&
+           h1->xlp_seg_size == h2->xlp_seg_size &&
+           h1->xlp_xlog_blcksz == h2->xlp_xlog_blcksz);
+}
+
+// Helper: Generic deque comparison with custom comparator
+static bool
+compare_deque(struct deque* dq1, struct deque* dq2, bool (*compare)(void*, void*))
```
jesperpedersen, line 162: This belongs in `deque.h`

src/walinfo.c
```diff
@@ -92,6 +94,253 @@ usage(void)
...
+{
+   if (dq1 == NULL || dq2 == NULL)
+   {
+      return (dq1 == dq2);
+   }
+
+   if (pgmoneta_deque_size(dq1) != pgmoneta_deque_size(dq2))
+   {
+      printf("Deque sizes mismatch: %u != %u\n", pgmoneta_deque_size(dq1), pgmoneta_deque_size(dq2));
```
jesperpedersen, line 171: Logging

jesperpedersen (changes requested):

jesperpedersen:
@Mohab96 Yes, you are on the right track.

However, you should open a discussion and point to a working branch instead of this.

So, I'll close it

src/walinfo.c
Jubilee101: I don't think we need a generic deque comparator. This looks like something very specific for walinfo at the moment.

Issue #414, opened 2024-11-12
walinfo: Support reading from a directory


jesperpedersen:
Support all WAL files in a directory, including split records across files

Holo-xy:
Hello, could i work on this one ?

jesperpedersen:
@sh-soltanpour Are you still working on this ?
@Holo-xy Feel free to reach out through email to us

sh-soltanpour:
I'm stuck at a point of it and need time to figure it out, I'd say if you can work on this it can free my time up to review other works and PRs.
@Holo-xy Just a note that this is harder than a pre-project in my opinion, so you may want to pick sth more straight-forward

jesperpedersen:
@Holo-xy Take a look at it - it is basically a streaming `void*` such that it keeps decoding when the information is in the next file. Generate 2+ WAL files and decode the entire directory as a tesst

Holo-xy:
Thank you i'll start looking into it now

sh-soltanpour:
Use this script that I got from @jesperpedersen to generate WAL files:


```
DROP TABLE IF EXISTS test;

CREATE OR REPLACE FUNCTION random_string(INTEGER)
RETURNS TEXT AS
$BODY$
SELECT array_to_string(
    ARRAY (
        SELECT substring(
            '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz'
            FROM (ceil(random()*62))::int FOR 1
        )
        FROM generate_series(1, $1)
    ), 
    ''
)
$BODY$
LANGUAGE sql VOLATILE;

CREATE TABLE test AS SELECT generate_series(1, 1000000) AS id, random_string(20) AS data FROM generate_series(1, 1);

-- GRANT EXECUTE ON FUNCTION pg_create_restore_point TO test;
-- pg_create_restore_point ( name text ) → pg_lsn

ANALYZE test;

```

sh-soltanpour:
> also is there a function that can help detect split records?

No there is no function. It happens in the `pgmoneta_wal_parse_wal_file` when `EOF` happens in the middle of parsing a record.

jesperpedersen:
It could also be a file that is "empty" at the end, and the next record is in the next file.

You need to see this as a byte stream of WAL records - with potential holes

Mohab96:
How is it going? @Holo-xy

Holo-xy:
[...]

PR #599, opened 2025-04-16
[#563] WAL testing infrastructure

### LINKED ISSUE #563: walfile: Generate WAL file
Create infrastructure in `walfile.h` such that we can generate a WAL file (PostgreSQL 13 - 17) in an easy way. The interface should be as simple as possible, so the complexity has to be "hidden" in the implementation.

Then create test cases that verifies basic 13->17 files using pgmoneta-walinfo

Mohab96:
Resolves #563


Mohab96:
There's a segmentation fault and I am working on it, but the rest of the patch is ready for review.

Mohab96:
> Magic number mismatch: 53526 != 1237236587

Seems like everything is broken and it is more than just a seg fault. This task might take from me more time because I realized that its harder than I thought.

jesperpedersen:
You can use `pgmoneta_backtrace()` to get a better idea -- or `gdb` of course

jesperpedersen:
@Mohab96 Maybe you need a `src/include/walfile/wal_utils.h` file to help you

Mohab96:
Sorry for the delay, I was stuck the last couple of days in debugging problems in this patch but I got wasn't able to do anything. I tried to put this code in `walinfo.c` to be able to debug it and inspect it easily and I was stuck at this problem `Magic number mismatch: 53526 != 1237236587` and `1237236587` changes on every run, so apparently the code tries to access a place in memory where `magic_number` is not there. I couldn't know why this behavior happens. 

Any hints? @jesperpedersen @sh-soltanpour

jesperpedersen:
@sh-soltanpour You got this, right ?

sh-soltanpour:
Hi,
I will take a look into it tomorrow night. Sorry I've been super busy in the past two days

sh-soltanpour:
I'm having this problem running the project on mac:
```
make[2]: *** No rule to make target `/opt/homebrew/Cellar/openssl@3/3.4.0/lib/libcrypto.dylib', needed by `src/libpgmoneta.0.16.0.dylib'.  Stop.
```
Not sure if this is a problem with my laptop or we've changed this in pgmoneta. Are you able to send me the WAL file that is created in the testsuite to test that directly? @Mohab96

jesperpedersen:
@sh-soltanpour We have Linux / MacOS / FreeBSD in CI - see https://github.com/pgmoneta/pgmoneta/blob/main/.github/workflows/ci.yml#L108 if that helps

Mohab96:
> Are you able to send me the WAL file that is created in the testsuite to test that directly? @Mohab96

I think it was this https://drive.google.com/file/d/1MJPBiCTDwCRDub5lyoDusDn9W5HTfbx7/view?usp=sharing

jesperpedersen:
That file only says
```
Incomplete | | | | | Skipped
```

jesperpedersen:
maybe we need a mode where the raw bytes of the WAL records are displayed like `pgmoneta_log_mem`

Mohab96:
> That file only says
> 
> ```
> Incomplete | | | | | Skipped
> ```

Yeah I know, the problem is I can't debug so know why this behavior is happening.

Mohab96:
> maybe we need a mode where the raw bytes of the WAL records are displayed like `pgmoneta_log_mem`

I think it would help, I'll try to do it in this patch after I complete at least one test case.

jesperpedersen:
@Mohab96 `hexdump` shows that the file is almost `0`s - we could look at a format like that with color coding of records

sh-soltanpour:
Yeah I think the problem is the file, not `walinfo`. This is the output of `pg_waldump` which can't parse this as well. How is this file generated?

```
pg_waldump 00000001000000000000001D
pg_waldump: error: could not find a valid record after 0/1D000000
```

jesperpedersen:
@sh-soltanpour Could it be a boundary issue ? Maybe we need #414 to investigate more... @Holo-xy ?

sh-soltanpour:
> @sh-soltanpour Could it be a boundary issue ? Maybe we need #414 to investigate more... @Holo-xy ?

Is the file generated by `pgmoneta`?

Mohab96:
Yeah I know that the file is not valid, I tried walinfo and waldump, the problem is that I don't understand why this behavior happens. This file is generated using the testcase in the current pr.

jesperpedersen:
Unless you are using `pg_ctl -m immediate ...` the WAL file should be valid... There is data in it as shown by `hexdump`

Mohab96:
`pg_ctl` has nothing to do with this file, this file is totally generated from the test case in this pr.

Mohab96:
While revisiting work I've done here, I noticed something. While designing the process in which I will test a specific type of record I tried to find the structure that the infrastructure reads the records into and then tried to create it by hand. The problem with this approach (I don't know yet whether it is a problem or not) is that for example if tried to test other type of records I don't know which part exactly in the generate function (which is `generate_xlog_checkpoint_shutdown` in out case) I will need to change. I don't know what in this implementation is very specific to this type of records.

If you have any insights on this please tell me. @jesperpedersen @sh-soltanpour 

Since you have a bit of experience with the test suite, please take a look on this patch and tell me if you have any notes on why this weird behavior occurs. @ashu3103

jesperpedersen:
@Mohab96 I don't exactly what you are saying, but we need functionality such that it is easier to generate a WAL record for a specific PostgreSQL version so `int pgmoneta_wal_XYZ(int version, int type, void** record)` could be the prototype signature, or create a .h file with specific signatures (`XYZ_v13(`)

Mohab96:
> @Mohab96 I don't exactly know what you are saying

What I was saying is I don't know yet what should be changed in `generate_xlog_checkpoint_shutdown` to test a different record type. (probably `main_data`, but I'm not sure though)

test/tsclient.c
```diff
@@ -292,3 +302,333 @@ get_configuration_path()
...
+      pgmoneta_free_walfile(wf);
+   if (read_wf != NULL)
+      pgmoneta_free_walfile(read_wf);
+   if (path != NULL)
+      free(path);
+   return 1;
+}
+
+static void
+pgmoneta_free_walfile(struct walfile* wf)
```
jesperpedersen, line 507: static doesn't have pgmoneta_ as prefix

jesperpedersen:
@Mohab96 Create a new .h/.c pair for the WAL testing - tsclient is for overall testing infrastructure

jesperpedersen:
@sh-soltanpour Thoughts on the setup ?

Mohab96:
I am working on the setup right now, once it is ok, I will get back to the problem we had earlier.

Mohab96:
After thinking about this for a while, this is my proposed structure to setup wal testing.

1. create list of available types of wal records: 
```c
    #define XLOG_CHECKPOINT_SHUTDOWN 0
    #define XLOG_CHECKPOINT_ONLINE 1

    struct wal_record {
        int version;
        int type;
        struct walfile* (*generate)(void);
    };

    const wal_record wal_record_types[] = {
        { 14, XLOG_CHECKPOINT_SHUTDOWN, pgmoneta_generate_xlog_checkpoint_shutdown_v14 },
        { 15, XLOG_CHECKPOINT_SHUTDOWN, pgmoneta_generate_xlog_checkpoint_shutdown_v15 },
        { 16, XLOG_CHECKPOINT_SHUTDOWN, pgmoneta_generate_xlog_checkpoint_shutdown_v16 },
        { 17, XLOG_CHECKPOINT_SHUTDOWN, pgmoneta_generate_xlog_checkpoint_shutdown_v17 },
    };
```
2. create a function that test all the wal records for all versions: 
`pgmoneta_wal_test_all_types()`, this function will iterate over the 
`wal_record_types` array and call `pgmoneta_test_wal_record(record)` for each record.

3. create a function that test specific wal record for a specific version: 
`pgmoneta_test_wal_record(int version, int type)`, this function will iterate over 
the `wal_record_types` array and call `pgmoneta_test_wal_record(record)` for each 
record that matches the version and type.

4. create a function to generate wal record for a specific type and specific 
version: `pgmoneta_generate_<type>_v<version>()`

5. create a function to use this generate function, write the wal record to 
disk, read it back and compare it with the original wal record: 
`pgmoneta_test_wal_record(wal_record* record)`

I am still confused though whether to choose 2 or 3.
Option 1: create 1 test suite that has one test case that runs `pgmoneta_wal_test_all_types()`. I have no problem with this approach other than it is not modularized.

Option 2: create 1 test suite that has test case per record type, in every test case it tests all versions for this type of record. My only problem with this approach is that I will need to loop through the whole list to find those with the target type and any version and I will do this for every single record type, which is a bit inefficient. 

Any thoughts? @jesperpedersen @sh-soltanpour

Mohab96:
And all of this code will be inside `wal_utils.h|c`

jesperpedersen:
@Mohab96 I would create a `.c` with all test cases for a specific PostgreSQL version.

Sure, we will test PostgreSQL 18 right now, and it may change until September. But, we know  that we will retire PostgreSQL 13 in November, and in that case we can just remove the `_v13.c` file

jesperpedersen:
@Mohab96 It could be a function pointer to a `_base.c` -- you have to look at what is the best solution

Mohab96:
I got your point. I will play around and come up with a patch tomorrow at most.

Mohab96:
I am working on this patch and the last update for today is that I almost finished all this structure I explained [here](https://github.com/pgmoneta/pgmoneta/pull/599#issuecomment-2874575805) but I made a last minute change. What I was doing is that I created a test case for all v17 records and inside it is only this
```c
START_TEST(pgmoneta_test_wal_v17)
{
   int rc = pgmoneta_wal_test(17);
   ck_assert_int_eq(rc, 0);
}
END_TEST
```

that iterates over all records and test every record with version 17 from this list
```c
    const wal_record wal_record_types[] = {
        { 14, XLOG_CHECKPOINT_SHUTDOWN, pgmoneta_generate_xlog_checkpoint_shutdown_v14 },
        { 15, XLOG_CHECKPOINT_SHUTDOWN, pgmoneta_generate_xlog_checkpoint_shutdown_v15 },
        { 16, XLOG_CHECKPOINT_SHUTDOWN, pgmoneta_generate_xlog_checkpoint_shutdown_v16 },
        { 17, XLOG_CHECKPOINT_SHUTDOWN, pgmoneta_generate_xlog_checkpoint_shutdown_v17 },
    };
```

so I noticed that when this test case fail it will only say that the test case for v17 records failed, nothing more. So I replaced it with a test case for every record but all of them inside the same testsuite like this
```c
START_TEST(check_point_shutdown_v17)
{
   int rc = pgmoneta_test_walfile(generate_check_point_shutdown_v17);
   ck_assert_int_eq(rc, 0);
}
END_TEST
```

and the rest of the records will follow in the same manner. It also contributes to the clarity of the file to help know which records exactly are tested in this test suite.

I will updated this draft PR with what I reached until now in an hour or two then I will update my work log in the discussions. 

Just to keep you in the loop. @jesperpedersen @sh-soltanpour

jesperpedersen:
@Mohab96 I think this is a good direction - make sure that it is very clear which test is failing so f.ex. 13/checkpoint

Mohab96:
I could do it in the previous way but it will be only the log, but now I it will be available directly in the console when the test fails.

jesperpedersen:
@Mohab96 We just need it to be clear, and easy to add test cases for records like 13 or 18 in the above case - so the array idea is good

Mohab96:
This is still draft PR but just to show you where I am now. I will continue tomorrow.

jesperpedersen:
@Mohab96 Yes, looks like the right direction - just call it `test/wal/`

Mohab96:
Can you explain to me the purpose of this assert? https://github.com/pgmoneta/pgmoneta/blob/5482ce93d758ab41d75b052e1044dc780986251d/src/libpgmoneta/walfile/wal_reader.c#L231

because it fails because it sees that that the server version is 0. Regardless of the reason why it is 0, I don't understand why would I need such an assert, if I am inspecting some WAL files (since this function is not used in any other place other than `walinfo.c`), why would I need a postgres server? the headers contain everything I need.

```
Running suite(s): pgmoneta_test1
 pgmoneta_test2
 pgmoneta_test3
 pgmoneta_test4
 pgmoneta_test5
pgmoneta_test: /home/mohab/Documents/Coding/Projects/pgmoneta/src/libpgmoneta/walfile/wal_reader.c:231: int pgmoneta_wal_parse_wal_file(char *, int, struct walfile *): Assertion `config->common.servers[server].version == magic_value_to_postgres_version(long_header->std.xlp_magic)' failed.
92%: Checks: 14, Failures: 0, Errors: 1
```

Mohab96:
I am currently investigating why it is 0. Seems like a problem with the configuration maybe.

jesperpedersen:
@Mohab96 I don't know if we need this `assert`, but we need to be able to compare a WAL stream against a cluster - think about stream from 16, shutdown 16, upgrade to 17, start streaming again.

That said - walfilter will be an offline tool so you won't have access to any PostgreSQL cluster. So it sounds like this functionality of checking should be split into multiple steps

Mohab96:
Should I add some option called `online` or so in the configuration so that I can control these asserts using it?

jesperpedersen:
@Mohab96 No, we have `offline` in `main.c` so that tries in into `struct common_configuration` and #523

jesperpedersen:
@Mohab96 Put your requirements on #523, and remove the `assert` so you can continue. All your read/write/test functionality need to be done in an offline envrionment

Mohab96:
@jesperpedersen @sh-soltanpour 

Can you please take a look at the output of running the test suite "locally" because I am facing a strange error that I cannot understand because I can't understand `decode_xlog_record` well.
image
image

jesperpedersen:
@Mohab96 Yes, seeing the same
@sh-soltanpour Can you take a look, please

sh-soltanpour:
So basically this means that a flag saying the data is true is set, but the data is not there (which can be caused by a bug in parser code or a corrupted file)
What are the steps to reproduce? or send me the walfile please

Mohab96:
Just run the testsuite and it will generate the file and parse it.

Mohab96:
You can take a look at `generate_check_point_shutdown_v17` before debugging to see if there's an obvious problem I missed.

Mohab96:
Seems like something wrong happens in the last iteration in this loop `while (remaining > datatotal)`
image

The problem happens when it reads `block_id` and finds that it is <= 32, and accordingly it gets interpreted as it is referencing some block (which is not the case in the type of records I am testing "`check_point_shutdown` doesn't have blocks") instead of getting one of these 4 values 
```
#define XLR_BLOCK_ID_DATA_SHORT    255
#define XLR_BLOCK_ID_DATA_LONG     254
#define XLR_BLOCK_ID_ORIGIN        253
#define XLR_BLOCK_ID_TOPLEVEL_XID  252
```

Please take a look at this part, especially the part of `xlp_xlog_blcksz` because when I initialized it by 8192 I ran into other problem that I fixed by making it anything below 64. I can reproduce it if you think it is important. @sh-soltanpour 

```
   long_phd->std.xlp_magic = 0xD116;
   long_phd->std.xlp_info = 0;
   long_phd->std.xlp_tli = 1;
   long_phd->std.xlp_pageaddr = 0;
   long_phd->xlp_seg_size = 16777216;
   long_phd->xlp_xlog_blcksz = 50;
   long_phd->std.xlp_rem_len = 0;
```

sh-soltanpour:
@Mohab96 you're going in the correct way to debug it. I know the debug process is hard but I don't know any better way, so here we go:

1- use `xxd` command to get a hex version of your WAL file. Let's say you know that WAL reader reads the file correctly until index `n`, so jump straight to that index in the file. 
2- Then you should go step by step in the code and in the file, to see what content of the file is parsed into which struct of the code. Logging (and only logging what matters )here is the most important thing, so I usually go like this:
```
if (file_index == n)
   printf("YOUR LOG HERE\n");
```
3- With doing this, you will finally find an index in which the file starts to be interpreted incorrectly. Then you need to go back to WAL generation code to see if that one is generating wrong or maybe the reader is reading the file wrong.

jesperpedersen:
Maybe we need some infrastructure around this based on `pgmoneta_log_mem()` - load the file into memory, jump to the index (or iterate through them), and dump out the byte stream...

@sh-soltanpour Can you help @Mohab96 write something like that - if it makes sense - ?

sh-soltanpour:
> Maybe we need some infrastructure around this based on `pgmoneta_log_mem()` - load the file into memory, jump to the index (or iterate through them), and dump out the byte stream...
> 
> @sh-soltanpour Can you help @Mohab96 write something like that - if it makes sense - ?

A visualization that shows what bytes are parsed into what struct is the most helpful thing that we can have, just dumping out the bytestream would be similar to what `xxd` does

Mohab96:
Okay I just need you to double check values in [this function](https://github.com/Mohab96/pgmoneta/blob/a20e37255e4e764ece665068d0187ff15aa6fda3/test/wal/wal_17.c#L59) so that I know for sure that the generation logic is fine to start questioning reading and writing processes, because now I have many parts that I need to debug so narrowing them down will help me a lot. @sh-soltanpour

Mohab96:
Now I have even more weird (and funny tho) situation.

When I run the testsuite, I get this
```
Running suite(s): pgmoneta_test1
 pgmoneta_test2
 pgmoneta_test3
 pgmoneta_test4
 pgmoneta_test5
PASSED FROM HERE 1
PASSED FROM HERE 2
/home/mohab/Documents/Coding/Projects/pgmoneta/src/libpgmoneta/walfile/wal_reader.c:354:30: runtime error: member access within null pointer of type 'struct partial_xlog_record'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /home/mohab/Documents/Coding/Projects/pgmoneta/src/libpgmoneta/walfile/wal_reader.c:354:30 
64%: Checks: 14, Failures: 4, Errors: 1
/home/mohab/Documents/Coding/Projects/pgmoneta/test/testcases/pgmoneta_test_1.c:39:P:Core:test_pgmoneta_backup:0: Passed
```

but when I use `pgmoneta-walinfo ./walfiles/00000001000000000000001D` (which basically uses the exact same APIs, but with different configurations than in the testsuite) I get this
```
XLOG | 0/0 | 0/1D000028 | 122 | 122 | 0 | redo 12345678/9ABCDEF0; tli 1; prev tli 1; fpw true; wal_level logical; xid 0:42; oid 100; multi 200; offset 300; oldest xid 400 in DB 500; oldest multi 600 in DB 700; oldest/newest commit timestamp xid: 900/1000; oldest running xid 0 shutdown 
```

which is what is expected actually. The problem is that the testsuite fails and I have no idea why this happen, so I think I will spend some time on checking the configurations in both environments and see what changes.

<details>
<summary>Based on these numbers</summary>

```c
struct walfile*
generate_check_point_shutdown_v17(void)
{
   struct walfile* wf = NULL;
   struct xlog_long_page_header_data* long_phd = NULL;
   struct deque* page_headers = NULL;
   struct deque* records = NULL;
   struct xlog_page_header_data* ph = NULL;
   struct decoded_xlog_record* rec = NULL;

   wf = (struct walfile*)malloc(sizeof(struct walfile));
   if (wf == NULL)
   {
      goto error;
   }

   wf->magic_number = 0xD116;

   long_phd = (struct xlog_long_page_header_data*)malloc(sizeof(struct xlog_long_page_header_data));
   if (long_phd == NULL)
   {
      goto error;
   }
   long_phd->xlp_sysid = 7503680172681168497;
   long_phd->std.xlp_pageaddr = 0x0000000017000000;
   long_phd->std.xlp_magic = 0xD116;
   long_phd->std.xlp_info = 0x0007;
   long_phd->std.xlp_tli = 1;
   long_phd->xlp_seg_size = 16777216;
   long_phd->xlp_xlog_blcksz = 8192;
   long_phd->std.xlp_rem_len = 0; // sizeof(struct xlog_record) + sizeof(struct check_point_v17);
   wf->long_phd = long_phd;

   if (pgmoneta_deque_create(false, &page_headers))
   {
      goto error;
   }
   wf->page_headers = page_headers;

   ph = (struct xlog_page_header_data*)malloc(sizeof(struct xlog_page_header_data));
   if (ph == NULL)
   {
      goto error;
   }
   ph->xlp_magic = 0xD116;
   ph->xlp_info = 0;
   ph->xlp_tli = 1;
   ph->xlp_pageaddr = 0;
   if (pgmoneta_deque_add(page_headers, NULL, (uintptr_t)ph, 0))
   {
      goto error;
   }

   if (pgmoneta_deque_create(false, &records))
   {
      goto error;
   }
   wf->records = records;

   rec = (struct decoded_xlog_record*)malloc(sizeof(struct decoded_xlog_record));
   if (rec == NULL)
   {
      goto error;
   }
   rec->header.xl_tot_len = sizeof(struct xlog_record) + sizeof(struct check_point_v17);
   rec->header.xl_xid = 0;
   rec->header.xl_prev = 0;
   rec->header.xl_info = XLOG_CHECKPOINT_SHUTDOWN;
   rec->header.xl_rmid = 0;
   rec->header.xl_crc = 0x457256A5;
   rec->main_data_len = sizeof(struct check_point_v17);
   rec->main_data = (char*)malloc(rec->main_data_len);
   if (rec->main_data == NULL)
   {
      goto error;
   }

   struct check_point_v17 cp;
   memset(&cp, 0, sizeof(cp));
   cp.redo = 0x123456789ABCDEF0ULL;
   cp.this_timeline_id = 1;
   cp.prev_timeline_id = 1;
   cp.full_page_writes = true;
   cp.wal_level = 2;
   cp.next_xid.value = 42;
   cp.next_oid = 100;
   cp.next_multi = 200;
   cp.next_multi_offset = 300;
   cp.oldest_xid = 400;
   cp.oldest_xid_db = 500;
   cp.oldest_multi = 600;
   cp.oldest_multi_db = 700;
   cp.time = 800;
   cp.oldest_commit_ts_xid = 900;
   cp.newest_commit_ts_xid = 1000;
   cp.oldest_active_xid = 1100;
   memcpy(rec->main_data, &cp, sizeof(cp));
   if (pgmoneta_deque_add(records, NULL, (uintptr_t)rec, 0))
   {
      goto error;
   }

   memset(rec->blocks, 0, sizeof(rec->blocks));
   rec->max_block_id = -1; // No block references

   return wf;

error:
   if (wf != NULL)
   {
      if (long_phd != NULL)
      {
         free(long_phd);
      }
      if (page_headers != NULL)
      {
         struct deque_iterator* iter = NULL;
         if (pgmoneta_deque_iterator_create(page_headers, &iter) == 0)
         {
            while (pgmoneta_deque_iterator_next(iter))
            {
               free((void*)iter->value->data);
            }
            pgmoneta_deque_iterator_destroy(iter);
         }
         pgmoneta_deque_destroy(page_headers);
      }
      if (records != NULL)
      {
         struct deque_iterator* iter = NULL;
         if (pgmoneta_deque_iterator_create(records, &iter) == 0)
         {
            while (pgmoneta_deque_iterator_next(iter))
            {
               struct decoded_xlog_record* r = (struct decoded_xlog_record*)iter->value->data;
               free(r->main_data);
               free(r);
            }
            pgmoneta_deque_iterator_destroy(iter);
         }
         pgmoneta_deque_destroy(records);
      }
      free(wf);
   }
   return NULL;
}
```
</details>

Mohab96:
Although, this happens:
```
pgmoneta@apollo:~$ /usr/lib/postgresql/17/bin/pg_waldump 00000001000000000000001D 
pg_waldump: error: could not find a valid record after 0/1D000000
```

FWIW, today I generated multiple WAL files from PostgreSQL cluster on my machine and then directly after getting generated I used `pg_waldump` and it gave me
```
pg_waldump: error: WAL segment size must be a power of two between 1 MB and 1 GB
```

so, in the future, I don't think we should rely on it as a single source of truth.

jesperpedersen:
@Mohab96 For test cases you should create - at least - 1MB empty WAL file (with `\0`) and then insert the records into that file.

We need to be able to use PostgreSQL to load these, and 16MB is standard size, so maybe use that

Mohab96:
I don't understand which comment you're replying to, but if you're talking about the file generated by pgmoneta, then yeah it is 16 mb, if you're talking about postgresql's generated files, then they're generated from postgres, I've nothing to do with them.

```
root@apollo:/var/lib/postgresql/17/main/pg_wal# ls -lh
total 141M
-rw------- 1 postgres postgres  16M Jun 13 04:54 000000010000000000000010
-rw------- 1 postgres postgres  16M Jun 13 04:54 000000010000000000000011
-rw------- 1 postgres postgres  16M Jun 13 04:54 000000010000000000000012
-rw------- 1 postgres postgres  16M Jun 13 04:54 000000010000000000000013
-rw------- 1 postgres postgres  16M Jun 13 04:54 000000010000000000000014
-rw------- 1 postgres postgres  16M Jun 13 04:54 000000010000000000000015
-rw------- 1 postgres postgres  16M Jun 13 18:12 000000010000000000000016
-rw------- 1 postgres postgres  16M Jun 13 18:12 000000010000000000000017
drwx------ 2 postgres postgres 4.0K May 13 16:16 archive_status
drwx------ 2 postgres postgres 4.0K May 13 16:16 summaries
-rw-r--r-- 1 root     root      13M Jun 13 05:37 temp
root@apollo:/var/lib/postgresql/17/main/pg_wal# ls -lh /home/mohab/Documents/Coding/Projects/pgmoneta/build/walfiles
total 17M
-rw-rw-r-- 1 mohab mohab 16M Jun 13 20:46 00000001000000000000001D
root@apollo:/var/lib/postgresql/17/main/pg_wal# 
```

jesperpedersen:
@Mohab96 Former - we should do 16MB by default

Mohab96:
I think it is the case currently. If you think it is not, what do you recommend me to do?

jesperpedersen:
@Mohab96 I was just making it an explicit requirement for the test cases

sh-soltanpour:
I have a problem with mac that's blocking me from running the testsuite, I'm getting this error:
```
/opt/homebrew/include/ev.h:159:10: fatal error: 'string.h' file not found
  159 | #include <string.h> /* for memmove */
      |          ^~~~~~~~~~
1 error generated.
```

and when I try to fix it using different methods, I face this:

```
make[2]: *** No rule to make target `/Library/Developer/CommandLineTools/SDKs/MacOSX15.1.sdk/usr/lib/libbz2.tbd', needed by `src/libpgmoneta.0.18.0.dylib'.  Stop.
```
@jesperpedersen do you if this is this already reported or may be my local issue?

sh-soltanpour:
@Mohab96 in your testsuite, what happens if you change it to generate the fileand pass it to `pg_waldump` directly instead of `walinfo`?

jesperpedersen:
@ashu3103 Ideas for Mac ?

jesperpedersen:
@sh-soltanpour Can you `find / -type f -name string.h` ?

sh-soltanpour:
> find / -type f -name string.h

```

/Library/Developer/CommandLineTools/usr/lib/clang/16/include/llvm_libc_wrappers/string.h
/Library/Developer/CommandLineTools/usr/lib/clang/17/include/llvm_libc_wrappers/string.h
/Library/Developer/CommandLineTools/usr/lib/tapi/16/include/llvm_libc_wrappers/string.h
/Library/Developer/CommandLineTools/usr/lib/tapi/17/include/llvm_libc_wrappers/string.h
/Library/Developer/CommandLineTools/SDKs/MacOSX15.2.sdk/usr/include/c++/v1/__fwd/string.h
/Library/Developer/CommandLineTools/SDKs/MacOSX15.2.sdk/usr/include/c++/v1/string.h
/Library/Developer/CommandLineTools/SDKs/MacOSX15.2.sdk/usr/include/string.h
/Library/Developer/CommandLineTools/SDKs/MacOSX15.2.sdk/System/Library/Frameworks/Kernel.framework/Versions/A/Headers/string.h
/Library/Developer/CommandLineTools/SDKs/MacOSX15.5.sdk/usr/include/c++/v1/__fwd/string.h
/Library/Developer/CommandLineTools/SDKs/MacOSX15.5.sdk/usr/include/c++/v1/string.h
/Library/Developer/CommandLineTools/SDKs/MacOSX15.5.sdk/usr/include/string.h
/Library/Developer/CommandLineTools/SDKs/MacOSX15.5.sdk/System/Library/Frameworks/Kernel.framework/Versions/A/Headers/string.h
/Library/Developer/CommandLineTools/SDKs/MacOSX14.5.sdk/usr/include/c++/v1/__fwd/string.h
/Library/Developer/CommandLineTools/SDKs/MacOSX14.5.sdk/usr/include/c++/v1/string.h
/Library/Developer/CommandLineTools/SDKs/MacOSX14.5.sdk/usr/include/string.h
/Library/Developer/CommandLineTools/SDKs/MacOSX14.5.sdk/System/Library/Frameworks/Kernel.framework/Versions/A/Headers/string.h
/Library/Developer/CommandLineTools/Library/PrivateFrameworks/LLDB.framework/Versions/A/Resources/Clang/include/llvm_libc_wrappers/string.h
```

jesperpedersen:
@sh-soltanpour @ashu3103 This is the one that need to be in the include and library path

```
/Library/Developer/CommandLineTools/SDKs/MacOSX15.2.sdk/usr/include/string.h
```

Mohab96:
> @Mohab96 in your testsuite, what happens if you change it to generate the fileand pass it to `pg_waldump` directly instead of `walinfo`?

```
pg_waldump: error: could not find a valid record after 0/1D000000
```

sh-soltanpour:
> This is the one that need to be in the include and library path

I still get the second error:

```
make[2]: *** No rule to make target `/Library/Developer/CommandLineTools/SDKs/MacOSX15.1.sdk/usr/lib/libbz2.tbd', needed by `src/libpgmoneta.0.18.0.dylib'.  Stop.
```

sh-soltanpour:
> > @Mohab96 in your testsuite, what happens if you change it to generate the fileand pass it to `pg_waldump` directly instead of `walinfo`?
> 
> ```
> pg_waldump: error: could not find a valid record after 0/1D000000
> ```

@Mohab96 so based on this, it means the problem is with the generator, not the parser.

Mohab96:
> @Mohab96 so based on this, it means the problem is with the generator, not the parser.

Should I add two records instead of one record?

sh-soltanpour:
> > @Mohab96 so based on this, it means the problem is with the generator, not the parser.
> 
> Should I add two records instead of one record?

If you say it doesn't print the parses version of the first record no, the problem is with the record itself as it cannot parse it

sh-soltanpour:
@Mohab96 btw is there any specific reason to limit the test suite for only Postgres 17? I have a fedora 41 and it apparently doesn't support Postgres 17

Mohab96:
> @Mohab96 btw is there any specific reason to limit the test suite for only Postgres 17? I have a fedora 41 and it apparently doesn't support Postgres 17

I am just trying to build the test suite for a specific version (as I mentioned previously, there'll be test suite for every version with test case per record type) so I am just trying to implement any version and then after that I just document the steps.

Mohab96:
This is the first version where the test case works.
PTAL @jesperpedersen

src/include/walfile/wal_reader.h
```diff
@@ -97,6 +97,17 @@ typedef int64_t timestamp_tz;
 #define MAXALIGNTYPE(LEN)          TYPEALIGN(MAXIMUM_ALIGNOF, (LEN))
 #define SHORTALIGN(LEN)            TYPEALIGN(ALIGNOF_SHORT, (LEN))
 
+/* When record crosses page boundary, set this flag in new page's header */
+#define XLP_FIRST_IS_CONTRECORD     0x0001
```
jesperpedersen: Align the constants

src/libpgmoneta/logging.c
```diff
@@ -392,6 +392,14 @@ pgmoneta_log_line(int level, char* file, int line, char* fmt, ...)
          else if (config->common.log_type == PGMONETA_LOGGING_TYPE_FILE)
          {
             buf[strftime(buf, sizeof(buf), config->common.log_line_prefix, tm)] = '\0';
+            if (log_file == NULL)
```
jesperpedersen: This must be a missing `pgmoneta_start_logging()` call

src/libpgmoneta/walfile.c
```diff
@@ -238,8 +297,10 @@ pgmoneta_write_walfile(struct walfile* wf, int server __attribute__((unused)), c
    {
       fclose(file);
    }
-   pgmoneta_deque_iterator_destroy(page_header_iterator);
-   pgmoneta_deque_iterator_destroy(record_iterator);
+   if (record_iterator)
```
jesperpedersen: You don't need to guard this

src/libpgmoneta/walfile/wal_reader.c
```diff
@@ -241,7 +241,7 @@ pgmoneta_wal_parse_wal_file(char* path, int server, struct walfile* wal_file)
    }
    else
    {
-      assert(config->common.servers[server].version == magic_value_to_postgres_version(long_header->std.xlp_magic));
+      // assert(config->common.servers[server].version == magic_value_to_postgres_version(long_header->std.xlp_magic));
```
jesperpedersen: Just remove

src/libpgmoneta/walfile/wal_reader.c
```diff
@@ -1447,10 +1450,85 @@ pgmoneta_wal_encode_xlog_record(struct decoded_xlog_record* decoded, uint16_t ma
    char* ptr = NULL;
    struct xlog_record record = decoded->header;
 
-   /* Compute total length required for the buffer */
-   total_length = record.xl_tot_len;
+   total_length = SIZE_OF_XLOG_RECORD;
+
+   // if (decoded->record_origin != INVALID_REP_ORIGIN_ID)
```
jesperpedersen: Can we remove this ?

test/wal/wal_17.c
```diff
@@ -0,0 +1,232 @@
...
+   memset(wf, 0, sizeof(struct walfile));
+
+   wf->long_phd = (struct xlog_long_page_header_data*)malloc(sizeof(struct xlog_long_page_header_data));
+   if (wf->long_phd == NULL)
+   {
+      goto error;
+   }
+
+   memset(wf->long_phd, 0, sizeof(struct xlog_long_page_header_data));
+   wf->long_phd->xlp_sysid = 7503680172681168497;
```
jesperpedersen: There should be a small description for this

jesperpedersen: Could we use some constants ?

test/wal/wal_17.c
```diff
@@ -0,0 +1,232 @@
...
+   {
+      goto error;
+   }
+
+   memset(rec, 0, sizeof(struct decoded_xlog_record));
+
+   struct check_point_v17 cp;
+   memset(&cp, 0, sizeof(cp));
+
+   cp.redo = 0x123456789ABCDEF0ULL;
```
jesperpedersen: Same

test/wal/wal_utils.c
```diff
@@ -0,0 +1,457 @@
...
+      return -1;
+   }
+
+   if (!compare_long_page_headers(wf1->long_phd, wf2->long_phd))
+   {
+      pgmoneta_log_error("Long page header mismatch\n");
+      return -1;
+   }
+
+   // if (!compare_deque(wf1->page_headers, wf2->page_headers, compare_xlog_page_header))
```
jesperpedersen: Remove ?

test/wal/wal_utils.c
```diff
@@ -0,0 +1,457 @@
...
+   destroy_walfile(wf);
+   destroy_walfile(read_wf);
+   free(path);
+   return 0;
+
+error:
+   destroy_walfile(wf);
+   destroy_walfile(read_wf);
+
+   if (path != NULL)
```
jesperpedersen: You don't need to guard

test/wal/wal_utils.c
Mohab96: yep, I'll do it.

src/libpgmoneta/walfile/wal_reader.c
Mohab96: I was testing something and forgot them, I'll think again and either remove or uncomment them.

test/wal/wal_17.c
Mohab96: These attributes have completely random values, so do we need to pick a random value and assign it to a constant? I think we don't need to, and probably will be forgotten, since when anyone tries to write a test case, it's easier to pick any random value that comes to their mind instead of searching through the codebase to find the proper constant (if they know it exists). What do you think?

Mohab96: Some of these attributes are pointless in this context, so I will simply remove them in order not to confuse whoever tries to write another testcase.

jesperpedersen: Just call the constant `RANDOM_BLAH` so we know it isn't important what the value actually is

Mohab96: I realized that we don't even have to put any values in most of these attributes, just being there is enough (e.g. for an int field, just allocating 4 bytes would be enough to test the parsing, and we don't have to put any values inside). But I'll leave them for illustration only so that it will be easier for the rest of the test cases in the future.

Mohab96:
If you think the documentation needs more details on what to consider while writing the generate function, please let me know. @jesperpedersen

src/libpgmoneta/logging.c
```diff
@@ -201,7 +201,7 @@ pgmoneta_start_logging(void)
 
       if (!log_file)
       {
-         printf("Failed to open log file %s due to %s\n", strlen(config->common.log_path) > 0 ? config->common.log_path : "pgmoneta.log", strerror(errno));
+         pgmoneta_log_error("Failed to open log file %s due to %s\n", strlen(config->common.log_path) > 0 ? config->common.log_path : "pgmoneta.log", strerror(errno));
```
jesperpedersen: But, logging won't work at this point

src/libpgmoneta/walfile.c
```diff
@@ -149,87 +148,147 @@ pgmoneta_write_walfile(struct walfile* wf, int server __attribute__((unused)), c
...
-      fseek(file, page_number * wf->long_phd->xlp_xlog_blcksz, SEEK_SET);
-      fwrite(page_header, sizeof(struct xlog_page_header_data), 1, file);
-      page_number++;
-   }
-
-   pgmoneta_deque_iterator_destroy(page_header_iterator);
-
-   page_number = 0;
-   fseek(file, SIZE_OF_XLOG_LONG_PHD, SEEK_SET);
+   int current_page = 0;
```
jesperpedersen: Try and do variables at the beginning of the function

src/libpgmoneta/walfile.c
```diff
@@ -149,87 +148,147 @@ pgmoneta_write_walfile(struct walfile* wf, int server __attribute__((unused)), c
...
    {
       char* encoded_record = NULL;
-      struct decoded_xlog_record* record = (struct decoded_xlog_record*) record_iterator->value->data;
+      struct decoded_xlog_record* record = (struct decoded_xlog_record*)record_iterator->value->data;
+
+      /* Encode the record */
       encoded_record = pgmoneta_wal_encode_xlog_record(record, wf->long_phd->std.xlp_magic, encoded_record);
+      if (!encoded_record)
+      {
+         pgmoneta_log_error("Failed to encode WAL record");
```
jesperpedersen, line 184: can't we log more information ?

src/libpgmoneta/walfile.c
```diff
@@ -149,87 +148,147 @@ pgmoneta_write_walfile(struct walfile* wf, int server __attribute__((unused)), c
...
-            /* Add padding until MAXALIGN */
-            int padding = MAXALIGN(ftell(file)) - ftell(file);
-            if (padding > 0)
+            char zero_padding[8] = {0};
+            if (fwrite(zero_padding, 1, padding, file) != padding)
             {
-               char zero_padding[padding];
-               memset(zero_padding, 0, padding);
-               fwrite(zero_padding, 1, padding, file);
+               pgmoneta_log_error("Failed to write alignment padding");
```
jesperpedersen: Same

test/include/wal_utils.h
```diff
@@ -0,0 +1,101 @@
...
+#include <getopt.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <sys/mman.h>
+#include <sys/types.h>
+
+#define WAL_CHECKPOINT_SHUTDOWN 0
```
jesperpedersen: Try and do

```
#define A  0
#define AB 1
```

such the constants are aligned

jesperpedersen (changes requested):

jesperpedersen:
These are general comments - there are multiple places

src/libpgmoneta/walfile.c
Mohab96: The only case this condition will be true is when `buffer = malloc(total_length);` fails in `pgmoneta_wal_encode_xlog_record`, so I'll add a log in `pgmoneta_wal_encode_xlog_record`.

jesperpedersen:
Merged.

Issue #600, opened 2025-04-17
Build: cmake 3.31+


jesperpedersen:
```
CMake Warning (dev) at doc/CMakeLists.txt:15 (add_custom_command):
  The following keywords are not supported when using
  add_custom_command(TARGET): OUTPUTS.

  Policy CMP0175 is not set: add_custom_command() rejects invalid arguments.
  Run "cmake --help-policy CMP0175" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at doc/CMakeLists.txt:15 (add_custom_command):
  Exactly one of PRE_BUILD, PRE_LINK, or POST_BUILD must be given.  Assuming
  POST_BUILD to preserve backward compatibility.

  Policy CMP0175 is not set: add_custom_command() rejects invalid arguments.
  Run "cmake --help-policy CMP0175" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at doc/CMakeLists.txt:22 (add_custom_command):
  The following keywords are not supported when using
  add_custom_command(TARGET): OUTPUTS.

  Policy CMP0175 is not set: add_custom_command() rejects invalid arguments.
  Run "cmake --help-policy CMP0175" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at doc/CMakeLists.txt:22 (add_custom_command):
  Exactly one of PRE_BUILD, PRE_LINK, or POST_BUILD must be given.  Assuming
  POST_BUILD to preserve backward compatibility.

  Policy CMP0175 is not set: add_custom_command() rejects invalid arguments.
  Run "cmake --help-policy CMP0175" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at doc/CMakeLists.txt:29 (add_custom_command):
  The following keywords are not supported when using
  add_custom_command(TARGET): OUTPUTS.

  Policy CMP0175 is not set: add_custom_command() rejects invalid arguments.
  Run "cmake --help-policy CMP0175" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.
[...]

PR #604, opened 2025-04-21
[#600] Build: cmake 3.31+

Userfrom1995:
New policy rules
https://cmake.org/cmake/help/latest/policy/CMP0175.html#policy:CMP0175

As per the new CMake policy CMP0175, I've updated the project to comply with the latest standards. The old policy is being deprecated and will soon be retired, so it’s important to transition to the new one

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -1,6 +1,7 @@
 # man target
 add_custom_target(man ALL)
 
+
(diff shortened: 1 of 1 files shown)


jesperpedersen:
@Userfrom1995 Please, port to pgagroal and pgexporter

jesperpedersen:
Merged.

Issue #612, opened 2025-04-28
backup.sha512 for an entire backup


jesperpedersen:
Create a `backup.sha512` file that is generated last of all files in a backup.

Ideally the format should be able to be read by `sha512sum`

jesperpedersen:
@shikharish Can you take this one ?

shikharish:
Yes.

PR #613, opened 2025-04-29
[#612] Create backup.sha512 for entire backup

shikharish:
cc: @jesperpedersen

### DIFF
--- a/src/libpgmoneta/wf_sha512.c
+++ b/src/libpgmoneta/wf_sha512.c
@@ -1,0 +1,355 @@
+/*
+ * Copyright (C) 2025 The pgmoneta community
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
+ * are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this list
+ * of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice, this
+ * list of conditions and the following disclaimer in the documentation and/or other
+ * materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its contributors may
+ * be used to endorse or promote products derived from this software without specific
+ * prior written permission.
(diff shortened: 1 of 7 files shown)


shikharish:
I have simply used the code from `wf_sha256.c`. It is mostly the same except the format. 512 follows `sha512sum` format(tested 👍🏻). Maybe we can combine the two if and when needed.

jesperpedersen:
@shikharish Did you check if the file changes when a retain / expunge command is executed ? Due to backup.info...

shikharish:
No. Adding that.

shikharish:
done.

src/libpgmoneta/wf_sha512.c
```diff
@@ -0,0 +1,355 @@
...
+#include <assert.h>
+#include <dirent.h>
+#include <errno.h>
+#include <unistd.h>
+
+static char* sha512_name(void);
+static int sha512_execute(char*, struct art*);
+
+static int write_backup_sha512(char* root, char* relative_path);
+int pgmoneta_update_sha512(char* root_dir, char* filename);
```
jesperpedersen, line 46: This is a public function

jesperpedersen (changes requested):

jesperpedersen:
@shikharish Can you add some documentation in the user / advanced guide on how to use the file with `sha512sum` ?

jesperpedersen:
```
... backup primary
... cd ...

sha512sum --check backup.sha512 
./backup.manifest: OK
./backup.sha512: FAILED
./backup.info: FAILED
./data/backup_label: OK
./data/pg_hba.conf.zstd.aes: OK
```

We should probably not put `backup.sha512` in the file

jesperpedersen:
Well, ideally we would have `backup.sha512` in there...

shikharish:
`backup.sha512` can't really be there i think. Practically a file cannot contain its owh hash..

jesperpedersen:
Merged.

Issue #623, opened 2025-05-06
test suite: Development servers


jesperpedersen:
```
./testsuite.sh: line 167: [: 18devel: integer expression expected
```

ashu3103:
> [@ashu3103](https://github.com/ashu3103) Can you take this ?

Sure

PR #624, opened 2025-05-06
[#623] test suite: Development servers

ashu3103:
The recent postgres versions which are either the beta versions or the pre-release version have versioning something like `18devel` or `18beta1` . To parse the version and compare it with our requirement in testsuite, we read only the numeric part of the version string using regex.
`s/^([0-9]+(\.[0-9]+)?).*/\1/`.

Note: It will however drop the patch version.

@jesperpedersen PTAL

### DIFF
--- a/test/testsuite.sh
+++ b/test/testsuite.sh
@@ -161,7 +161,7 @@ check_psql() {
 }
 
 check_postgres_version() {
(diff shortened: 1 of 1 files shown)


jesperpedersen:
Merged.

jesperpedersen:
@ashu3103 Please, port as necessary

PR #641, opened 2025-05-25
[#640]Fix Hash validation

### LINKED ISSUE #640: [Bug]hash mismatch logic
While reviewing the code in verify.c, I noticed a logic error involving the use of strcmp when comparing hash values. The current implementation uses:
```c
if (!strcmp(hash, calculated_hash))
{
  // hash mismatch
}
```

This condition is  incorrect if the intention is to detect mismatched hashes, since strcmp(a, b) returns 0 when strings are equal.

jchrys:
fixes #640

### DIFF
--- a/src/libpgmoneta/verify.c
+++ b/src/libpgmoneta/verify.c
@@ -406,7 +406,7 @@ pgmoneta_sha512_verification(char** argv)
                goto cleanup;
             }
 
-            if (!strcmp(hash, calculated_hash))
+            if (strcmp(hash, calculated_hash) != 0)
             {
                pgmoneta_log_error("Verification: Server %s / Hash mismatch for %s | Expected: %s | Got: %s",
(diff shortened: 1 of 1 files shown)


src/libpgmoneta/verify.c
```diff
@@ -406,7 +406,7 @@ pgmoneta_sha512_verification(char** argv)
                goto cleanup;
             }
 
-            if (!strcmp(hash, calculated_hash))
+            if (strcmp(hash, calculated_hash))
```
jesperpedersen, line 409: Lets do `!= 0`

jesperpedersen (changes requested):

src/libpgmoneta/verify.c
jchrys: Sure :D

jesperpedersen:
Merged.

Issue #638, opened 2025-05-22
Block Reference Table impl


ashu3103:
**Is your feature request related to a problem? Please describe.**

**Problem Statement:** To enable block-level incremental backups from a base reference backup we need to have some information about the blocks that have been modified since the base backup. This feature aims to cater that issue.

**Describe the solution you'd like**

**Solution:** We will fetch the block information from the archived WAL segments. For a range of LSNs (typically from REDO[n-1] record to REDO[n] record), we will generate summaries about the modified blocks for each relation file fork and store that info in a data structure - `BlockReferenceTable`. 

**Additional information**

Some key features related to this issue:

- Each file is divided into chunks of 2^16 blocks and an appropriate representation of a chunk in memory is decided based on the number of modified blocks in a chunk (the representation can be an array or a bitmap).
- The reference table is itself a key-value data structure, where key is/is derived from the `relfilelocator` of the file and the `forknumber` of the file while the value contains an array of chunks which further contains modified blocks info, it also provides an idea of `limit_block` (shortest known length of the relation in blocks within the LSN range covered by a particular block reference table)
- Finally we have some APIs to serialize this in-memory data structure into summary files on disk (will be used later by **Phase-2**).

ps: Many of the features are inspired directly from the [postgres](https://github.com/postgres/postgres).

PR #646, opened 2025-06-02
[#638] block reference table impl

ashu3103:
### Tasks done

- Block Reference Table Data Structure impl - **Manipulation** and **Serialization** APIs
- Tests to write the data structure to a file and read from a file.


src/include/value.h
```diff
@@ -64,6 +64,7 @@ enum value_type {
    ValueARTRef,
    ValueRef,
    ValueMem,
+   ValueBRTEntry,
```
Jubilee101: We would like value types being restricted only to primitive types and core data structures. You don't have to create a new value type here. For ART take a look at `pgmoneta_art_insert_with_config`. It allows you to put the value into the ART with a configuration specifying how to destroy the value and/or convert the value to string. 

You can also do similar thing in deque, with `pgmoneta_deque_add_with_config`. The usage of config is identical, search for the function in code if you need examples.

src/include/brt.h
```diff
@@ -0,0 +1,265 @@
...
+ * .....
+ * .....
+ * rlocatorN | forknumN | limit_blockN | nchunksN | entryN_chunk_usage | entryN_chunk_data | 0 | 0 | 0 | 0 |
+ * The last serialized entry is all zeros and denote a termination
+ * @param brt The block reference table
+ * @param file The file stream
+ * @return 0 if success, otherwise failure
+ */
+int
+pgmoneta_brt_write(block_ref_table* brt, FILE* file);
```
Jubilee101: Would it be better if input is dir path instead?

src/include/brt.h
```diff
@@ -0,0 +1,265 @@
...
+
+/**
+ * Read the contents of the summary file and create a block reference table from it
+ * @param file The file stfream
+ * @param [out] brt The block reference table
+ * @return 0 if success, otherwise failure
+ */
+int
+pgmoneta_brt_read(FILE* file, block_ref_table** brt);
```
Jubilee101: Same comment as above.

src/include/brt.h
```diff
@@ -0,0 +1,265 @@
...
+ * 'chunk_usage' is an array storing the number of elements used in each chunk. If that value is less
+ * than MAX_ENTRIES_PER_CHUNK, the corresponding chunk is used as an array; else the corresponding
+ * chunk is used as a bitmap. When used as a bitmap, the least significant bit of the first array element
+ * is the status of the lowest-numbered block covered by this chunk.
+ *
+ * 'chunk_data' is the array of chunks.
+ */
+typedef struct block_ref_table_entry
+{
+   block_ref_table_key key;     /**< The key used to search for the block entry in the ART */
```
Jubilee101: Don't have to do it now. But eventually you need to align the comments.

src/include/brt.h
```diff
@@ -0,0 +1,265 @@
...
+                       enum fork_number forknum, block_number* limit_block);
+
+/**
+ * Get block numbers from a table entry.
+ * @param entry pointer to the brt entry
+ * @param start_blkno start processing from this block number
+ * @param stop_blkno stop processing, beyond this block number
+ * @param blocks [out] Block array which contain all the modified blocks in the brt entry between [start_blkno, stop_blkno)
+ * @param nblocks The lower bound on the number of blocks that can be stored
+ * @return The return value is the number of block numbers actually written.
```
Jubilee101: Returning results from output parameter is better. Reserve the return value for error code.

src/include/brt.h
```diff
@@ -0,0 +1,265 @@
...
+ *
+ * 'chunk_data' is the array of chunks.
+ */
+typedef struct block_ref_table_entry
+{
+   block_ref_table_key key;     /**< The key used to search for the block entry in the ART */
+   block_number limit_block;          /**< The limit block for the relation fork */
+   uint32_t nchunks;          /**< The number of chunks for the relation fork */
+   uint16_t* chunk_size;           /**< The size of each chunk in the relation fork */
+   uint16_t* chunk_usage;           /**< The number of used entries in each chunk */
```
Jubilee101: This only means something when data is in array representation right? You can make it clearer.

Jubilee101:
I think this is good direction.

src/include/brt.h
ashu3103: Okay, will do it with the file path

src/include/brt.h
ashu3103: Sure, will make the changes accordingly

src/include/value.h
ashu3103: Hmm that's a better way without including additional code.

Nice!

ashu3103:
@Jubilee101 PTAL

src/libpgmoneta/value.c
```diff
@@ -27,6 +27,7 @@
  */
 /* pgmoneta */
 #include <art.h>
+#include <brt.h>
```
jesperpedersen: *Cough*

jesperpedersen:
@ashu3103 Rebase, and run uncrustify

src/libpgmoneta/value.c
ashu3103: Oopss

jesperpedersen:
Merged.

Issue #649, opened 2025-06-05
WAL Summarizer


ashu3103:
**Is your feature request related to a problem? Please describe.**

Since now we have a infrastructure to store the modified blocks for a range of LSNs. The next task is to parse WAL within that range and detect the changed blocks from the archived WAL records.

For this we need:

- Archiving, parsing and filtering of wal (fortunately, this is already there)
- a separate wal summarizer process that read the wal segment and generate summary files.

## Solution:

A WAL summarizer API in `pgmoneta` that takes `timeline`, `startLSN` and `endLSN` as input, parse and summarize all the WAL records in that timeline between [`startLSn`, `endLSN`) and write the summary (information about the modified blocks in that LSN range) into the file `{timeline}.{startLSN}.{endLSN}.summary.

Currently there are two design choices circling my head :-

### Design-1

Have a separate background process in the main, which runs till the lifetime of the main process and summarize the WAL records from the WAL achive as and when possible and save them into summary files (used later by incremental backup workflows).

Here the `startLSN` would be the `endLSN` of the latest summary file (if summary files present) or the starting of the oldest WAL segment present in all of the timelines (if summary file not present), the `endLSN` can be the LSN of the checkpoint done in `pg_backup_start()` while initiating incremental backups.

#### Challenges:
- While using summarizer API what if timeline changes
- Since you have to provide startLSN and endLSN to the summarizer API, it can happen that startLSN and endLSN bekongs to different WAL segments.
- Think for the retention of the summary files.

### Design-2

Summarize only the wal records needed for incremental backups. No additional process, just call the summarize API in the incremental backup workflow. WALl summary file generation can be optional as the block reference table created after the summarization phase can directly be used by the next phase of the workflow.

Here the `startLSN` would be the `stopLSN` of the base backup and `endLSN` will be the LSN of the checkpoint done in `pg_backup_start()` while initiating incremental backups.

The wal summary files generated during this time can be discarded.

#### Challenges:
- While using summarizer API what if timeline changes
- Since you have to provide startLSN and endLSN to the summarizer API, it can happen that startLSN and endLSN bekongs to different WAL segments.
- Maintain a large set of WAL records, which is probably not scalable, as Robert Haas said (atleast upto the oldest backup)

**Note:** Implementation of WAL summarizer API is independent of the design choice (in progress).

@jesperpedersen, @Jubilee101 PTAL

jesperpedersen:
[...]

Issue #255, opened 2024-04-09
Server streams WAL segments lazily


Jubilee101:
We've noticed for quite a while that server sometimes does not sends us WAL segment eagerly - it just idles and sends status report request to keep the connection open, even when it has more segments not yet streamed to us.

We'll try to force server do CHECKPOINT, so that all the records on WAL get applied to disk, and see if that triggers streaming.

jesperpedersen:
We only require a `REPLICATION` user so we don't have access to executing the `CHECKPOINT` command

jesperpedersen:
For PostgreSQL 15+ we could add the `pg_checkpoint` role, and update our requirements for `pg_hba.conf` to allow non-replication based login.

So, it has to be an optional operation

jesperpedersen:
Unless we can come up with something else

Jubilee101:
> For PostgreSQL 15+ we could add the pg_checkpoint role, and update our requirements for pg_hba.conf to allow non-replication based login.
>
> So, it has to be an optional operation

..That doesn't seem ideal. I'll think about it.

bassamadnan:
Would influencing [wal_writer_delay](https://postgresqlco.nf/doc/en/param/wal_writer_delay/#:~:text=Time%20between%20WAL%20flushes%20performed,by%20an%20asynchronously%20committing%20transaction.) have an impact on this? i've been [reading](https://demirhuseyinn-94.medium.com/optimizing-postgresql-performance-the-impact-of-adjusting-commit-delay-and-wal-writer-delay-0f4dd0402cca) on some material but most of these come from configuring the database itself and would have tradeoffs.

hzhang-vibrant:
> Would influencing [wal_writer_delay](https://postgresqlco.nf/doc/en/param/wal_writer_delay/#:~:text=Time%20between%20WAL%20flushes%20performed,by%20an%20asynchronously%20committing%20transaction.) have an impact on this? i've been [reading](https://demirhuseyinn-94.medium.com/optimizing-postgresql-performance-the-impact-of-adjusting-commit-delay-and-wal-writer-delay-0f4dd0402cca) on some material but most of these come from configuring the database itself and would have tradeoffs.

[...]

PR #663, opened 2025-06-23
[#649] WAL Summarizer

ashu3103:
## WORK IN PROGRESS | DRAFT

A simple API that summarizes all the WAL records in the range (start_lsn, end_lsn) in a single summary file (for now flushing the contents of `BRT` into a file, but its optional as we will be requiring the in memory data structure only once for the incremental backup).

### Pseudo code

- Opens, Reads and Parses each `walfile` in the WAL directory (similar to `pgmoneta_describe_walfiles_in_directory`)
- For each file, iterate through each parsed WAL record.
- If the parsed WAL record `LSN` is not within the range return and continue with next record, otherwise proceed further.
- If the parsed WAL record belongs to operations that don't modify a block like `REDO`, `CHECKPOINT SHUTDOWN` etc. return and continue with next record, otherwise proceed further.
- Special handling of WAL record that can change the `limit_block` of a relation file like those belonging to `{"info":"XLOG_SMGR_TRUNCATE"}` (triggered for truncate command).
- Final step - for every `block_number` from 0 to `max_block_number` (extracted from WAL record `main_data`), check if this `block_number` is been referenced by a relation file locator, if yes, that means this block is definitely modified. (execute `brt_mark_block_modified` for this block number)

### Assumptions

This API assumes the following
- The WAL records from `start_lsn` to `end_lsn` is present in WAL archive
- `start_lsn` and `end_lsn` are always the starting of a WAL record and not the middle of WAL record.
- `wal_level` is greater than `WAL_LEVEL_MINIMAL`

### TODO

- [x] Add testcases (tests with multiple timelines and partial records).
- [ ] For now, the API do all the task in one go, there can be an issue if it has to summarize a huge pile of WAL records, maybe we can add some fallback mechanisms where we write small summaries, such that whenever anything fails we can continue from the last summarized point.

**Disclaimer:** This draft is just for reference or to provide the idea of my implementation. It may contain memory bugs or runtime bugs (Testing ongoing!).

@sh-soltanpour PTAL at the logic of `pgmoneta_summarize_walfile` (As per my understanding it should be able to accommodate cases like timeline switches and partial records (records extending beyond the WAL segment)).


src/libpgmoneta/walfile/wal_reader.c
```diff
@@ -1159,6 +1175,95 @@ pgmoneta_wal_record_display(struct decoded_xlog_record* record, uint16_t magic_v
...
+int
+pgmoneta_wal_record_summary(struct decoded_xlog_record* record, uint64_t start_lsn, uint64_t end_lsn, struct block_ref_table* brt)
+{
+   struct rel_file_locator rlocator;
+   enum fork_number forknum;
+   block_number blocknum;
+   uint8_t info;
+
+   /* Simply return if record is partial or not in range */
+   if (record->partial || !is_included(NULL, NULL, record->header.xl_prev, start_lsn,
```
ashu3103: @sh-soltanpour PTAL at the `is_included` logic -
```
if (record->partial || !is_included(NULL, NULL, record->header.xl_prev, start_lsn,
                                       record->lsn, end_lsn, record->header.xl_xid, NULL, NULL, NULL))
   {
      return 0;
   }
```

I am still dicey about it and think this -
```
if (record->partial || !is_included(NULL, NULL, record->lsn, start_lsn,
                                       record->next_lsn, end_lsn, record->header.xl_xid, NULL, NULL, NULL))
   {
      return 0;
   }
```
would be appropriate. Maybe my understanding is skewed.

jesperpedersen:
@Mohab96 Cc:

Mohab96:
I modified this function previously but I have no idea what is the difference between these LSNs, all I know is that they're used in walinfo when the user want to specify the target LSN range so consider this case and take it from `walinfo.c` and see what they mean.

src/include/walfile.h
```diff
@@ -164,4 +164,26 @@ pgmoneta_describe_walfiles_in_directory(char* dir_path, enum value_type type, ch
...
+/**
+ * Summarize WAL files in a directory
+ * @param dir_path The path to the WAL files directory
+ * @param start_lsn The start LSN
+ * @param end_lsn The end LSN
+ * @param brt The block reference table
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_summarize_walfiles_in_directory(char* dir_path, uint64_t start_lsn, uint64_t end_lsn, struct block_ref_table* brt);
```
Jubilee101: `pgmoneta_summarize_walfiles` will suffice I presume.

src/libpgmoneta/walfile.c
```diff
@@ -502,4 +503,158 @@ pgmoneta_describe_walfiles_in_directory(char* dir_path, enum value_type type, ch
...
+      // functions delete the source file
+      pgmoneta_copy_file(wal_path, tmp_wal, NULL);
+      copy = false;
+
+      pgmoneta_strip_extension(basename(wal_path), &decrypted_file_name);
+
+      free(wal_path);
+      wal_path = NULL;
+
+      wal_path = pgmoneta_format_and_append(wal_path, "/tmp/%s", decrypted_file_name);
```
Jubilee101: There is a `workspace` configuration, by default it's `/tmp/`. You can see its usage in `pgmoneta_extract_backup_file`. I think perhaps we can have another `pgmoneta_extract_wal_file`. @jesperpedersen  Thoughts?

src/libpgmoneta/walfile/wal_reader.c
```diff
@@ -1159,6 +1175,95 @@ pgmoneta_wal_record_display(struct decoded_xlog_record* record, uint16_t magic_v
...
+    */
+   for (int block_id = 0; block_id <= record->max_block_id; block_id++)
+   {
+      if (!xlog_record_get_block_tag(record, block_id, &rlocator, &forknum, &blocknum))
+      {
+         continue;
+      }
+
+      /*
+       * As we do elsewhere, ignore the FSM fork, because it's not
```
Jubilee101: I know the idea is largely inspired and borrowed from Postgres. But this comment seems to be directly copied from postgres, which for me isn't exactly very appropriate.

src/libpgmoneta/walfile/wal_summary.c
```diff
@@ -0,0 +1,150 @@
...
+ * Handles summarization of arbitary range in a timeline only
+ * (It is the job of upper layers to keep the timeline intact)
+ *
+ * Assumptions this function makes:
+ * - The WAL records from start_lsn to end_lsn is present in WAL archive
+ * - wal_level is greater than WAL_LEVEL_MINIMAL
+ * - start_lsn and end_lsn are always the starting of a WAL record and not the middle of WAL record
+ */
+int
+pgmoneta_summarize_wal(int srv, uint64_t start_lsn, uint64_t end_lsn)
```
Jubilee101: Just an idea, is it really necessary for us to write the summary to disk? We only seem to use the summary for one time. Unless of course you plan to reuse some of the summaries in the future somehow.

Jubilee101: ...Ignore this if you need it in #673

src/libpgmoneta/walfile/wal_reader.c
ashu3103: I'll update it

src/libpgmoneta/walfile/wal_summary.c
ashu3103: As I mentioned above, it is optional to save the contents of `BRT` to the disk (as it is one time), we can directly return the data structure to the caller.

> Unless of course you plan to reuse some of the summaries in the future somehow.

As of now there is no such plan but we can definitely ponder over it. Shall I change the function signature in that case?

Jubilee101: > As of now there is no such plan but we can definitely ponder over it. Shall I change the function signature in that case?

Nah, current is fine. We can always circle back.

src/libpgmoneta/walfile.c
ashu3103: Sounds good to me but we then have to make similar changes (maybe in terms of function signature or some refactoring) in `pgmoneta_describe_walfiles_in_directory` and `pgmoneta_describe_walfile` as well.

Because for now, I have tried to make it consistent with `pgmoneta_describe_walfile`.

jesperpedersen: Refactoring is ok

ashu3103: Okay so just to reiterate the implementation -

- We'll have a function `pgmoneta_extract_wal_file` similar to `pgmoneta_extract_backup_file` but instead of taking server index as an argument (and deriving wal directory inside this func) we'll directly pass the wal directory (otherwise we'll have to modify the signatures of `pgmoneta_describe_walfile` and `pgmoneta_summarize_wal`)
- Fitting `pgmoneta_extract_wal_file` in `pgmoneta_describe_walfile` and `pgmoneta_summarize_wal` with some refactoring.

jesperpedersen: You can refactor code - use of static functions; e.g. not public API is best - but just change function signatures as you see fit. Code duplication isn't good

ashu3103: Just realized changing function signatures would be a complete mess because `pgmoneta_extract_backup_file` is backup specific and need to have a server id associated with it while the use case of `pgmoneta_extract_wal_file` is not server dependent as we may require any arbitary wal file (at any location in file system) to be extracted (extracted here means - decrypted and decompressed to prepare it for read).

An alternative would be to create a public API in `utils.h` named `pgmoneta_copy_and_extract_file(char* from, char* to)` - this way we will take out the logic of copying and extracting a file from src to dst from `pgmoneta_extract_backup_file` function.

jesperpedersen: compression.h/pgmoneta_decompress ?

ashu3103: `pgmoneta_decompress` alone won't be enough, `pgmoneta_copy_and_extract_file` will contain the logic for -

- copying src file to dst file
- check if the file is encrypted (using extension) -if yes decrypt it
- check if the file is compressed (using extension) - if yes decompress it

This draft implements it by doing each of these tasks sequentially (similar to one implemented in `pgmoneta_describe_walfile`) but we were trying to do all these task in a wrapper function (similar to `pgmoneta_extract_backup_file`).

jesperpedersen: well, the functions are there - you just need to implement a wrapper, and check existing usage - like pgmoneta_decrypt_file - if it can use the new function instead

ashu3103: @jesperpedersen @Jubilee10, PTAL

`pgmoneta_summarize_walfile()` ->
for now I have made changes such that, the wal file will be extracted in `/tmp/` directory (writing tests is easier this way), later (while writing incremental backup workflow) we can pass the server and label to this function and find the destination using `workspace` configuration.

src/libpgmoneta/walfile/rm_xact.c
```diff
@@ -372,6 +372,30 @@ pgmoneta_wal_parse_xl_xact_parsed_abort_v15(struct xl_xact_parsed_abort* wrapper
...
+}
+
+void
+pgmoneta_wal_parse_commit_record_v15(uint8_t info, struct xl_xact_commit* xlrec, struct xl_xact_parsed_commit_v15* parsed)
+{
+   parse_commit_record_v15(info, xlrec, parsed);
+}
+
+void
+pgmoneta_wal_parse_abort_record_v14(uint8_t info, struct xl_xact_abort* xlrec, struct xl_xact_parsed_abort_v14* parsed)
```
Jubilee101: I know that v14 works for <14 as well, but the naming is a bit confusing. I know it's because it was named that way initially. But now that you have made the API public, try to make the naming clearer. That being said, I don't have good ideas rn..

test/testcases/pgmoneta_test_5.c
```diff
@@ -0,0 +1,58 @@
...
+#include <tsclient.h>
+#include <wal.h>
+#include <walfile/wal_reader.h>
+
+#include "pgmoneta_test_5.h"
+
+START_TEST(test_pgmoneta_summarize_wal)
+{
+    int found = 0;
+    found = !pgmoneta_tsclient_summarize_wal("primary", 19671232, 33554680);
```
Jubilee101: How do you know the start LSN and end LSN here?

test/testcases/pgmoneta_test_5.c
```diff
@@ -0,0 +1,58 @@
...
+ *
+ */
+
+#include <tsclient.h>
+#include <wal.h>
+#include <walfile/wal_reader.h>
+
+#include "pgmoneta_test_5.h"
+
+START_TEST(test_pgmoneta_summarize_wal)
```
Jubilee101: You may also need to test if the the summary file can be parsed back.

src/libpgmoneta/walfile/wal_reader.c
```diff
@@ -55,6 +61,16 @@ static bool is_included(char* rm, struct deque* rms,
                         uint64_t e_lsn, uint64_t end_lsn,
                         uint32_t xid, struct deque* xids, char** included_objects, char* rm_desc);
 
+static bool xlog_record_get_block_tag(struct decoded_xlog_record* record, int block_id,
+                                      struct rel_file_locator* rlocator, enum fork_number* frk,
+                                      block_number* blkno);
+
+static bool summarize_xlog_record(uint8_t info);
+static void summarize_xact_record_14(struct decoded_xlog_record* record, block_ref_table* brt);
```
Jubilee101: Maybe `summarize_xact_record_l15` and `summarize_xact_record_ge15`, just an idea

src/libpgmoneta/walfile/wal_summary.c
```diff
@@ -0,0 +1,172 @@
...
+   free(wal_dir);
+   free(summary_dir);
+   free(summary_filename);
+   pgmoneta_brt_destroy(brt);
+   pgmoneta_stop_logging();
+   return 1;
+}
+
+static char*
+summary_file_name(uint64_t s_lsn, uint64_t e_lsn)
```
Jubilee101, line 160: What about timeline info?

Jubilee101:
Only a few small issues, this is good as long as you have tested the functionalities.

test/testcases/pgmoneta_test_5.c
ashu3103: Glad you raised it, I was about to take it to the discussions. Actually these LSNs were there in my local snapshots of client wal stream (on `pgmoneta` side) after executing testcase 1 to testcase 4, and I have tested the whole functionality using those wal files.

However for testsuite, I was thinking, if we could have a public API, that returns a list of all previous `CHECKPOINT`s we'll just randomly choose any two and put them as arguments here.

test/testcases/pgmoneta_test_5.c
ashu3103: Whether the summary file can be successfully parsed back or not is included in testcase 4. Maybe we can merge testcase 4 and 5.

src/libpgmoneta/walfile/wal_summary.c
ashu3103: I think timeline info will only make sense if we are making different summaries for different timelines (a new summary file after each timeline switch), but that's not the case here. Also I couldn't think of any case where we want to access timeline info from summary files (Do let me know if its relevant for our future milestones).

test/testcases/pgmoneta_test_5.c
Jubilee101: Sorry for the late reply. Hardcoding is probably not ok... 
And I don't think you should rely on records generated by previous tests. Perhaps you can a checkpoint, generate some logs(do an insert), and checkpoint again and summarize between those two checkpoints. The idea is to have the logs more controllable so that you can do more detailed unit testing

test/testcases/pgmoneta_test_5.c
Jubilee101: I see. Nah, in this case just leave them separated. If You know what relations it's going to touch you can also verify that. Ignore If it's too troublesome.

src/libpgmoneta/walfile/wal_summary.c
Jubilee101: Hmm, you are right. I was thinking about avoiding file name conflict, but lsn is still incremental.

test/testcases/pgmoneta_test_5.c
ashu3103: Hi @Jubilee101,

I am moving ahead with this approach and by using `pgmoneta_query_execute` API in the testsuite, I am performing checkpoint (using command: `CHECKPOINT;`). But this API does not handle the response type from checkpoint command that is `CommandComplete`.

```
CommandComplete (B)

    Byte1('C')

        Identifies the message as a command-completed response.
    Int32

        Length of message contents in bytes, including self.
    String

        The command tag. This is usually a single word that identifies which SQL command was completed.
```

Shall we include the support for this response type or something like we detect this response type, set response as `NULL` and doesn't throw an error (typically that's what I need)?

jesperpedersen: @ashu3103 Yes, we can include support for `('C')` in the query infrastructure. You can just return the command tag as the result. We can merge this separate from this PR.

It is ok to use `CHECKPOINT` in the testsuite as we can execute as `SUPERUSER` - for users we have the pgmoneta_extension

ashu3103:
@Jubilee101 @jesperpedersen, PTAL 

For testsuite I'm doing ->

- Checkpoint - (`start_lsn` is obtained from here)
- execute `pgbench` to generate some WAL records
- Checkpoint - (`end_lsn` is obtained from here)
- summarize between `start_lsn` and `end_lsn`

Implementation is working fine but the either the WAL records (from `start_lsn` to `end_lsn`) are not streamed to `main` or they are in partial segments.

Jubilee101:
> Implementation is working fine but the either the WAL records (from start_lsn to end_lsn) are not streamed to main or they are in partial segments.

So does the test pass? We just finished investigating the lazy streaming problem, see https://github.com/pgmoneta/pgmoneta/issues/255#issuecomment-3092593508 for details. So it's unlikely that the records don't get streamed to our side. Try parsing the partial segment with pgmoneta-walinfo to confirm that's the issue. If that's indeed the case then just leave it as @Mohab96 is currenlty investigating it. And move on to the next task of yours. We can come back to this later.

test/tsclient.c
```diff
@@ -310,6 +468,68 @@ pgmoneta_tsclient_read(block_ref_table** brt)
...
+   }
+
+   /* get the starting lsn for summary */
+   if (do_checkpoint_and_get_lsn(srv, &s_lsn))
+   {
+      goto error;
+   }
+
+   /* generate wal records using pgbench */
+   if (pgmoneta_tsclient_execute_pgbench(srv, false, 0, 0, 0))
```
Jubilee101: It's better to just insert a record into a table in my opinion. You'll have a clearer understanding whether the summarizer is working as expected -- by seeing if it picks up the changes on the table relation. pg_bench can be used if we want to stress test the summarizer to see its performance.

Jubilee101:
> So it's unlikely that the records don't get streamed to our side. 

That being said sometimes it's still likely that the records take a while to get here. Try sleeping a little bit (10-20 seconds should be enough) to wait for that.

test/tsclient.c
ashu3103: Okay, will try that

ashu3103:
Hi @Jubilee101 

The testcase now do - 

- Checkpoint
- Create a Table and Insert some tuples to it
- Checkpoint
- Sleep for 5 seconds
- summarize

> Try parsing the partial segment with pgmoneta-walinfo to confirm that's the issue.

I have tried parsing the partial segment (by removing the `.partial` suffix from the wal segment) using walinfo and found that the records always appear in the partial segments.

Maybe one thing we can do is do a `pg_switch_wal()` before summarizing to make it a full segment.

jesperpedersen:
@ashu3103 @Mohab96 is working on the .partial part - for now I think it is ok to go with `pg_switch_wal()`

Jubilee101:
> The testcase now do -

Yup that looks nice. 

> I have tried parsing the partial segment (by removing the .partial suffix from the wal segment) using walinfo and found that the records always appear in the partial segments.

Do you have to remove the suffix for it to work? What happens if you just feed it to walinfo with the original name?

> Maybe one thing we can do is do a pg_switch_wal() before summarizing to make it a full segment.

If the problem is with walinfo then this is the right way to go at the moment. Otherwise something may be wrong with the streamer.

ashu3103:
> Do you have to remove the suffix for it to work? What happens if you just feed it to walinfo with the original name?

Yes, otherwise it just says that the wal segment is partial and moves ahead

ashu3103:
@Jubilee101, @jesperpedersen PTAL

I have added a `pg_switch_wal()` just before summarization and now we are getting the desired output.

One more thing, the clang Release build for CI is again and again failing while it is working fine locally. On inspecting the logs, I found that the backup command for testcase 2 is failing but have no idea why since the logs are clean (no `ERROR`s and no `FATAL`s)

src/include/walfile/wal_summary.h
```diff
@@ -0,0 +1,49 @@
...
+ * @param wal_dir The directory to the wal segments (Optional and is used for testing)
+ * @param start_lsn The start lsn is the point at which we should start summarizing. If this
+ * value comes from the end LSN of the previous record as returned by the xlogreader machinery,
+ * 'exact' should be true; otherwise, 'exact' should be false, and this function will search
+ * forward for the start of a valid WAL record.
+ * @param end_lsn identifies the point beyond which we can't count on being able to read any
+ * more WAL.
+ * @return 0 is success, otherwise failure
+ */
+int pgmoneta_summarize_wal(int srv, char* wal_dir, uint64_t start_lsn, uint64_t end_lsn);
```
jesperpedersen: Formatting

src/libpgmoneta/message.c
```diff
@@ -1079,10 +1079,12 @@ pgmoneta_query_execute(SSL* ssl, int socket, struct message* msg, struct query_r
       }
       else if (status == MESSAGE_STATUS_ZERO)
       {
+         printf("status not success -> 1");
```
jesperpedersen: Remove

src/libpgmoneta/message.c
```diff
@@ -1079,10 +1079,12 @@ pgmoneta_query_execute(SSL* ssl, int socket, struct message* msg, struct query_r
...
          SLEEP(1000000L);
       }
       else
       {
+         printf("status not success -> 2");
```
jesperpedersen: Remove

src/libpgmoneta/walfile.c
```diff
@@ -304,80 +303,30 @@ pgmoneta_describe_walfile(char* path, enum value_type type, FILE* out, bool quie
...
    if (!pgmoneta_is_file(path))
    {
       pgmoneta_log_fatal("WAL file at %s does not exist", path);
       goto error;
    }
 
-   wal_path = pgmoneta_append(wal_path, path);
+   from = pgmoneta_append(from, path);
+   /* Extract the wal file in /tmp/ */
+   to = pgmoneta_append(to, "/tmp/");
```
jesperpedersen, line 380: Is `/tmp` safe ?

src/libpgmoneta/walfile.c
```diff
@@ -304,80 +303,30 @@ pgmoneta_describe_walfile(char* path, enum value_type type, FILE* out, bool quie
...
-
-      wal_path = pgmoneta_format_and_append(wal_path, "/tmp/%s", decompressed_file_name);
-      free(decompressed_file_name);
-
-      if (pgmoneta_decompress(tmp_wal, wal_path))
-      {
-         pgmoneta_log_fatal("Failed to decompress WAL file at %s", path);
-         goto error;
-      }
+      pgmoneta_log_fatal("Failed to extract WAL file from %s to %s", from, to);
```
jesperpedersen: FATAL means `exit(1)` so it is maximum an ERROR

src/libpgmoneta/walfile.c
```diff
@@ -488,3 +437,106 @@ pgmoneta_describe_walfiles_in_directory(char* dir_path, enum value_type type, FI
...
+{
+   struct walfile* wf = NULL;
+   struct deque_iterator* record_iterator = NULL;
+   struct decoded_xlog_record* record = NULL;
+   char* from = NULL;
+   char* to = NULL;
+
+   if (!pgmoneta_is_file(path))
+   {
+      pgmoneta_log_fatal("WAL file at %s does not exist", path);
```
jesperpedersen: Same

src/libpgmoneta/walfile.c
```diff
@@ -488,3 +437,106 @@ pgmoneta_describe_walfiles_in_directory(char* dir_path, enum value_type type, FI
...
+   }
+
+   from = pgmoneta_append(from, path);
+   /* Extract the wal file in /tmp/ */
+   to = pgmoneta_append(to, "/tmp/");
+   to = pgmoneta_append(to, basename(path));
+
+   if (pgmoneta_copy_and_extract_file(from, &to))
+   {
+      pgmoneta_log_fatal("Failed to extract WAL file from %s to %s", from, to);
```
jesperpedersen: Same

src/libpgmoneta/walfile.c
```diff
@@ -488,3 +437,106 @@ pgmoneta_describe_walfiles_in_directory(char* dir_path, enum value_type type, FI
...
+      goto error;
+   }
+
+   /* Read and Parse the WAL records of this WAL file */
+   if (pgmoneta_read_walfile(-1, to, &wf))
+   {
+      pgmoneta_log_fatal("Failed to read WAL file at %s", path);
```
jesperpedersen: Same

src/libpgmoneta/walfile.c
```diff
@@ -488,3 +437,106 @@ pgmoneta_describe_walfiles_in_directory(char* dir_path, enum value_type type, FI
...
+      goto error;
+   }
+
+   if (pgmoneta_deque_iterator_create(wf->records, &record_iterator))
+   {
+      pgmoneta_log_fatal("Failed to create deque iterator");
```
jesperpedersen: Same

jesperpedersen: Should we add information to DEVELOPERS.d about logging levels ?

src/libpgmoneta/walfile.c
```diff
@@ -488,3 +437,106 @@ pgmoneta_describe_walfiles_in_directory(char* dir_path, enum value_type type, FI
...
+      goto error;
+   }
+
+   /* Iterate each record */
+   while (pgmoneta_deque_iterator_next(record_iterator))
+   {
+      record = (struct decoded_xlog_record*)record_iterator->value->data;
+      if (pgmoneta_wal_record_summary(record, start_lsn, end_lsn, brt))
+      {
+         pgmoneta_log_fatal("Failed to summarize the WAL record at %s", pgmoneta_lsn_to_string(record->lsn));
```
jesperpedersen: Same

src/libpgmoneta/walfile/wal_summary.c
```diff
@@ -0,0 +1,172 @@
...
+
+   partial_record = malloc(sizeof(struct partial_xlog_record));
+   partial_record->data_buffer_bytes_read = 0;
+   partial_record->xlog_record_bytes_read = 0;
+   partial_record->xlog_record = NULL;
+   partial_record->data_buffer = NULL;
+   /* Look upon the WAL archive directory and summarize the WAL records in the range [start_lsn, end_lsn) */
+   if (pgmoneta_summarize_walfiles(wal_dir, start_lsn, end_lsn, brt))
+   {
+      pgmoneta_log_fatal("Error while reading/describing WAL directory");
```
jesperpedersen: Same

src/libpgmoneta/walfile/wal_summary.c
```diff
@@ -0,0 +1,172 @@
...
+   }
+   free(partial_record);
+
+   // /* Generate the WAL summaries */
+   summary_dir = pgmoneta_get_server_summary(srv);
+   if (!pgmoneta_ends_with(summary_dir, "/"))
+   {
+      summary_dir = pgmoneta_append_char(summary_dir, '/');
+   }
+   pgmoneta_log_debug("summary dir: %s", summary_dir);
```
jesperpedersen, line 112: What if there are multiple clusters running summary ?

src/walinfo.c
```diff
@@ -164,6 +164,7 @@ main(int argc, char** argv)
       {"v", "verbose", false},
       {"S", "summary", false},
       {"V", "version", false},
+      {"S", "summary", false},
```
jesperpedersen: Have you rebased ?

jesperpedersen:
@ashu3103 Lets start with those
@Mohab96 Can you look at your part of this patch ?

src/libpgmoneta/walfile.c
ashu3103: I don't see an issue with `/tmp/`, since it is present on most of Unix-like systems (the systems `pgmoneta` is [tested on](https://github.com/pgmoneta/pgmoneta?tab=readme-ov-file#tested-platforms) do have `/tmp` directory by default).

I didn't really get the context of `safe` here? @jesperpedersen

src/libpgmoneta/walfile.c
ashu3103: Okay. So shall I replace `fatal` with `error` in my PR? Or just ones you have explicitly marked for change

src/walinfo.c
ashu3103: Will do that

src/libpgmoneta/walfile.c
ashu3103: > Should we add information to DEVELOPERS.d about logging levels ?

it's better if we have it.

src/libpgmoneta/walfile/wal_summary.c
ashu3103: If you are worried about the conflict between the summary directory for two server instances, that won't happen because the summary directory is defined over a server and is unique for each server instance.

src/libpgmoneta/walfile.c
jesperpedersen: I mean access and enough space - anyway, we can improve later

src/libpgmoneta/walfile.c
jesperpedersen: all of them

ashu3103:
> @ashu3103 Lets start with those @Mohab96 Can you look at your part of this patch ?

@jesperpedersen PTAL

test/runner.c
```diff
@@ -30,8 +30,9 @@
 
 #include "testcases/pgmoneta_test_1.h"
 #include "testcases/pgmoneta_test_2.h"
-#include "testcases/pgmoneta_test_3.h"
+// #include "testcases/pgmoneta_test_3.h"
```
jesperpedersen: What is wrong with 3 ?

test/include/tsclient.h
```diff
@@ -100,6 +104,30 @@ pgmoneta_tsclient_execute_restore(char* server, char* backup_id, char* position)
...
+ * Execute a pgbench command for a set of instructions, assuming we are connecting to the 1st server
+ * @param srv server index
+ * @param select_only true if we are only doing selects
+ * @param client_count number of clients
+ * @param thread_count number of threads
+ * @param transaction_count number of transactions
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_tsclient_execute_pgbench(int srv, bool select_only, int client_count, int thread_count, int transaction_count);
```
jesperpedersen: Do we really need this ?

jesperpedersen (changes requested):

test/runner.c
ashu3103: The TC-3 has some `heap-use-after-free` errors which was due to some concurrency issues. This would require changes in the current echo server implementation, so I thought it's better to comment the TC until @bassamadnan resolves it.

Refer [this](https://github.com/pgmoneta/pgmoneta/issues/683#issuecomment-3062626125) for details.

test/include/tsclient.h
ashu3103: As of now, **Not required**.

This is a port from `pgagroal` and sooner or later we may need it.

jesperpedersen: Just rip it out for now then

ashu3103: Done

jesperpedersen:
@Mohab96 PTAL
@Jubilee101 PTAL
@bassamadnan Please, bump the priority of the echo server, so we can get test3 back

Mohab96:
> @Mohab96 PTAL @Jubilee101 PTAL @bassamadnan Please, bump the priority of the echo server, so we can get test3 back

I am on it

bassamadnan:
>  @bassamadnan Please, bump the priority of the echo server, so we can get test3 back

Yup, shifting to this right now.

Jubilee101:
@ashu3103 The patch looks good to me. @bassamadnan is working on the tests and @Mohab96 is double checking the WAL parser part. I'll merge the patch once both are ok. In the meantime you may start on the next steps. I think you still need to get the backup start and end LSN, and also be able to stream the files to our side, starting on either one is ok for me. As for the task 2 of this PR, I think we can do it as an enhancement later, what do you think?

test/runner.c
bassamadnan: @ashu3103 im abit confused here, ever since [this](0c839b9ee60538f65748a09e8879a96cd74ffd1d) commit, the test cases for HTTP have passed. And they pass up to the latest commit as well, do you have any idea why they are failing?

bassamadnan: also i tried restoring the test suite for tc -3 [here](https://github.com/bassamadnan/pgmoneta/tree/refs/heads/pr-663) the test cases pass just fine ([link](https://github.com/bassamadnan/pgmoneta/actions/runs/16539122227)). This was done on top of your PR.

ashu3103: > @ashu3103 im abit confused here, ever since [this](0c839b9ee60538f65748a09e8879a96cd74ffd1d) commit, the test cases for HTTP have passed. And they pass up to the latest commit as well, do you have any idea why they are failing?

@bassamadnan, at the surface they appear to be passing, but if you check the details of the CI. Check `build-linux (clang, Debug)` for the [link](https://github.com/bassamadnan/pgmoneta/actions/runs/16539122227/job/46777850366) you have provided. you will find something like this ->

```
pgmoneta_test3
=================================================================
==11905==ERROR: AddressSanitizer: heap-use-after-free on address 0x503000002120 at pc 0x55ded5b88c69 bp 0x7f8837682b70 sp 0x7f8837682b68
READ of size 1 at 0x503000002120 thread T1
=================================================================
==11908==ERROR: AddressSanitizer: heap-use-after-free on address 0x503000002120 at pc 0x55ded5b88c69 bp 0x7f8837682b70 sp 0x7f8837682b68
READ of size 1 at 0x503000002120 thread T1
=================================================================
==11911==ERROR: AddressSanitizer: heap-use-after-free on address 0x503000002120 at pc 0x55ded5b88c69 bp 0x7f8837682b70 sp 0x7f8837682b68
READ of size 1 at 0x503000002120 thread T1
=================================================================
==11914==ERROR: AddressSanitizer: heap-use-after-free on address 0x503000002120 at pc 0x55ded5b88c69 bp 0x7f8837682b70 sp 0x7f8837682b68
READ of size 1 at 0x503000002120 thread T1
=================================================================
==11917==ERROR: AddressSanitizer: heap-use-after-free on address 0x503000002120 at pc 0x55ded5b88c69 bp 0x7f8837682b70 sp 0x7f8837682b68
READ of size 1 at 0x503000002120 thread T1
```

ashu3103: > do you have any idea why they are failing?

There were some synchronization issues with the echo server take a look at [this](https://github.com/pgmoneta/pgmoneta/issues/683#issuecomment-3062544176) .

ashu3103:
> I'll merge the patch once both are ok.

Cool

> In the meantime you may start on the next steps. I think you still need to get the backup start and end LSN, and also be able to stream the files to our side, starting on either one is ok for me

I would like to go ahead with the streaming of the data files to our side part.

> As for the task 2 of this PR, I think we can do it as an enhancement later, what do you think?

Yes, I think that would be better to take it later when everything is in place.

Jubilee101:
Hi @ashu3103 , can you rebase? I think the test should have been fixed.

ashu3103:
> Hi @ashu3103 , can you rebase? I think the test should have been fixed.

Yeah, I have enabled TC-3 and it's working fine, but I was facing some errors in TC-5 while parsing walfile after the [PR](https://github.com/pgmoneta/pgmoneta/pull/705).

@Mohab96 I was getting following errors:-
```
pgmoneta_test5
/home/ashu3103/Desktop/pgmoneta/src/libpgmoneta/walfile/wal_reader.c:549:13: runtime error: member access within null pointer of type 'struct decoded_xlog_record'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /home/ashu3103/Desktop/pgmoneta/src/libpgmoneta/walfile/wal_reader.c:549:13 
```

The line from which this error originates is -
```
537:   if (pgmoneta_deque_iterator_create(wal_file->records, &iter))
538:   {
539:      pgmoneta_log_error("Failed to create deque iterator");
540:      goto error;
541:   }
542:
543:   while (pgmoneta_deque_iterator_next(iter))
544:   {
545:      decoded = (struct decoded_xlog_record*) iter->value->data;
546:      decoded->next_lsn = lsn_array[idx++];
547:   }
548:
549:   decoded->next_lsn = 0; // Handle last record
550:
551:   pgmoneta_deque_iterator_destroy(iter);
```

At line number `549`, the program crashes when it encounters a `decoded == NULL`.

To resolve it, I have just added a security check for `decoded` here like:
```
if (decoded != NULL)
{
    decoded->next_lsn = 0;
}
```
@Mohab96 PTAL

jesperpedersen:
@ashu3103 Are you using an empty WAL file ?

ashu3103:
> @ashu3103 Are you using an empty WAL file ?

I don't think this is the case since I can see WAL records in every WAL segment (not sure how our current parser implementation handles partial files -> maybe it treats partial file as empty).

jesperpedersen:
@ashu3103 Attach the file that is failing

ashu3103:
[000000010000000000000003.zip](https://github.com/user-attachments/files/21453091/000000010000000000000003.zip)

@jesperpedersen, this is the file from the tests which was failing and yes this is an empty file.
Ignore my previous comment.

Mohab96:
> [000000010000000000000003.zip](https://github.com/user-attachments/files/21453091/000000010000000000000003.zip)

@ashu3103 
this file is giving me this error
```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==84494==ERROR: AddressSanitizer: SEGV on unknown address (pc 0x7a912ea44688 bp 0x7ffec83d93e0 sp 0x7ffec83d93a0 T0)
==84494==The signal is caused by a READ memory access.
==84494==Hint: this fault was caused by a dereference of a high value address (see register values below).  Disassemble the provided pc to learn which register was used.
    #0 0x7a912ea44688 in bool __sanitizer::atomic_compare_exchange_strong<__sanitizer::atomic_uint8_t>(__sanitizer::atomic_uint8_t volatile*, __sanitizer::atomic_uint8_t::Type*, __sanitizer::atomic_uint8_t::Type, __sanitizer::memory_order) ../../../../src/libsanitizer/sanitizer_common/sanitizer_atomic_clang.h:81
    #1 0x7a912ea44688 in __asan::Allocator::AtomicallySetQuarantineFlagIfAllocated(__asan::AsanChunk*, void*, __sanitizer::BufferedStackTrace*) ../../../../src/libsanitizer/asan/asan_allocator.cpp:668
    #2 0x7a912ea44688 in __asan::Allocator::Deallocate(void*, unsigned long, unsigned long, __sanitizer::BufferedStackTrace*, __asan::AllocType) ../../../../src/libsanitizer/asan/asan_allocator.cpp:724
    #3 0x7a912eafc49b in free ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:53
    #4 0x7a912dc321b6 in pgmoneta_destroy_walfile /home/mohab/Documents/Coding/Projects/pgmoneta/src/libpgmoneta/walfile.c:298
    #5 0x7a912dc30195 in pgmoneta_read_walfile /home/mohab/Documents/Coding/Projects/pgmoneta/src/libpgmoneta/walfile.c:123
    #6 0x7a912dc327ad in pgmoneta_describe_walfile /home/mohab/Documents/Coding/Projects/pgmoneta/src/libpgmoneta/walfile.c:380
    #7 0x5e849887f5f3 in main /home/mohab/Documents/Coding/Projects/pgmoneta/src/walinfo.c:621
    #8 0x7a912cc2a1c9 in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58
    #9 0x7a912cc2a28a in __libc_start_main_impl ../csu/libc-start.c:360
    #10 0x5e849887b8c4 in _start (/home/mohab/Documents/Coding/Projects/pgmoneta/build/src/pgmoneta-walinfo+0xb8c4) (BuildId: 4fda39a4d6852d915b926c25f1a4db2997d0505c)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV ../../../../src/libsanitizer/sanitizer_common/sanitizer_atomic_clang.h:81 in bool __sanitizer::atomic_compare_exchange_strong<__sanitizer::atomic_uint8_t>(__sanitizer::atomic_uint8_t volatile*, __sanitizer::atomic_uint8_t::Type*, __sanitizer::atomic_uint8_t::Type, __sanitizer::memory_order)
==84494==ABORTING
➜  build git:(main) 
```

please provide the file that causes [this](https://github.com/pgmoneta/pgmoneta/pull/663#issuecomment-3124226751) issue.

Mohab96:
@ashu3103 

Where did you get that file from? Because, as far as I know, even if the file has no records, it should have a long header and then be filled with zeros to fit 16mb
```
➜  build git:(main) ls -lh /home/mohab/Downloads/000000010000000000000003                                              
-rw------- 1 mohab mohab 0 Jul 27 12:30 /home/mohab/Downloads/000000010000000000000003
```

ashu3103:
> @ashu3103
> 
> Where did you get that file from? Because, as far as I know, even if the file has no records, it should have a long header and then be filled with zeros to fit 16mb
> 
> ```
> ➜  build git:(main) ls -lh /home/mohab/Downloads/000000010000000000000003                                              
> -rw------- 1 mohab mohab 0 Jul 27 12:30 /home/mohab/Downloads/000000010000000000000003
> ```

@Mohab96, this file is generated when we run the `testsuite` inside the local test environment. In the `testsuite.sh`, comment the `clean` command inside `run_tests` and you will be able to find this inside -
`<path/to/build>/build/pgmoneta-testsuite/primary/wal`

ashu3103:
@jesperpedersen @Jubilee101 any more comments on this?

I am planning to start with the phase-2 of the project.

jesperpedersen:
@ashu3103 Yes, go ahead - we will merge @Mohab96's patch first then yours

Jubilee101:
Hi @ashu3103 , can you rebase?

ashu3103:
> Hi @ashu3103 , can you rebase?

Done

jesperpedersen:
@ashu3103 Can you look at CI ?

ashu3103:
Hi @Jubilee101, For the testcase corresponding to this PR, I want to grant `repl` with `pg_checkpoint` and so I am just adding `GRANT pg_checkpoint TO PG_REPL_USER_NAME` to `setup.sql`.

But it doesn't seem to be working, can you help me with it?

Jubilee101:
Hi @ashu3103 , can you elaborate how it's not working? There is a log under /tmp/pgmoneta-test/pg_log/ that'll show you if postgres is granting the role. 

Another thing you can do is to manually build and start the container. `cd` into test/postgresql/src/postgresql17/, run `make all` to rebuild the image. Start the pod. Then as long as you have psql installed locally you are able to connect to the server and open the sql terminal. From there you'll be able to manually run the sql command and see what's wrong. Refer to https://github.com/pgmoneta/pgmoneta/blob/main/test/postgresql/src/postgresql17/README.md

ashu3103:
> Hi @ashu3103 , can you elaborate how it's not working? There is a log under /tmp/pgmoneta-test/pg_log/ that'll show you if postgres is granting the role.
> 
> Another thing you can do is to manually build and start the container. `cd` into test/postgresql/src/postgresql17/, run `make all` to rebuild the image. Start the pod. Then as long as you have psql installed locally you are able to connect to the server and open the sql terminal. From there you'll be able to manually run the sql command and see what's wrong. Refer to https://github.com/pgmoneta/pgmoneta/blob/main/test/postgresql/src/postgresql17/README.md

Everything was fine, I just didn't realized that I have to rebuild the image. Thanks for the help

ashu3103:
@Jubilee101 @jesperpedersen,

Fixed the `test_brt_io` and `test_wal_summary` test suites 
Removed `test_wal_utils` from runner until its fixed

PTAL

Jubilee101:
> Removed test_wal_utils from runner until its fixed

Please don't remove this, @Mohab96 is fixing the issue. You only need to fix your own test cases.

test/include/tsclient.h
```diff
@@ -73,6 +74,17 @@ pgmoneta_tsclient_delete(char* server, char* backup_id);
...
+/**
+ * Execute an SQL query on postgres database
+ * @param srv The server index
+ * @param query The query to be executed
+ * @param is_dummy Use dummy user and password
+ * @param qr [out] The query response
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_tsclient_execute_query(int srv, char* query, bool is_dummy, struct query_response** qr);
```
Jubilee101: tsclient is for sending commands to pgmoneta (like CLI). Move this to tscommon

test/check.sh
```diff
@@ -295,13 +295,21 @@ export_pgmoneta_test_variables() {
 
   echo "export PGMONETA_TEST_RESTORE_DIR=$RESTORE_DIRECTORY"
   export PGMONETA_TEST_RESTORE_DIR=$RESTORE_DIRECTORY
+
+  echo "export PGMONETA_TEST_USER_CONF=$CONFIGURATION_DIRECTORY/pgmoneta_users.conf"
+  export PGMONETA_TEST_USER_CONF=$CONFIGURATION_DIRECTORY/pgmoneta_users.conf
+
+  echo "export PGMONETA_TEST_WAL_DIR=$BACKUP_DIRECTORY/primary/wal"
```
Jubilee101: I don't think you need to export this. See `pgmoneta_get_server_wal` and `PRIMARY_SERVER`

test/libpgmonetatest/tscommon.c
```diff
@@ -89,7 +102,14 @@ pgmoneta_test_environment_create(void)
...
     memcpy(TEST_BASE_DIR, base_dir, strlen(base_dir));
 
+    wal_dir = getenv(ENV_VAR_WAL_DIR);
+    assert(wal_dir != NULL);
+    memcpy(TEST_WAL_DIR, wal_dir, strlen(wal_dir));
+
     pgmoneta_start_logging();
+
+    // Try reading the users configuration path
+    assert(!pgmoneta_read_users_configuration(shmem, user_conf_path));
```
Jubilee101, line 104: Is this the only place we are using the user_conf_path? If so then we should not make this global, reading it from env and loading the config is enough.

test/postgresql/src/postgresql17/conf/setup.sql
```diff
@@ -27,4 +27,8 @@
 CREATE ROLE PG_USER_NAME WITH LOGIN PASSWORD 'PG_USER_PASSWORD';
 CREATE DATABASE PG_DATABASE WITH OWNER PG_USER_NAME TEMPLATE template0 ENCODING UTF8;
 
-CREATE ROLE PG_REPL_USER_NAME WITH LOGIN REPLICATION PASSWORD 'PG_REPL_PASSWORD';
\ No newline at end of file
+CREATE ROLE PG_REPL_USER_NAME WITH LOGIN REPLICATION PASSWORD 'PG_REPL_PASSWORD';
+
+-- GRANT REPL_USER priviledge to perform CHECKPOINT and pg_switch_wal
+GRANT pg_checkpoint TO PG_REPL_USER_NAME;
+GRANT EXECUTE ON FUNCTION pg_catalog.pg_switch_wal() TO PG_REPL_USER_NAME;
```
Jubilee101, line 34: This is ok in test, but I hope we don't have to do it in production.

test/runner.c
```diff
@@ -51,14 +52,16 @@ main(int argc, char* argv[])
...
+   // wal_utils_suite = pgmoneta_test_wal_utils_suite();
+   brt_io_suite = pgmoneta_test_brt_io_suite();
    wal_summary_suite = pgmoneta_test_wal_summary_suite();
 
    sr = srunner_create(backup_suite);
    srunner_add_suite(sr, restore_suite);
    srunner_add_suite(sr, delete_suite);
    srunner_add_suite(sr, http_suite);
-   srunner_add_suite(sr, wal_utils_suite);
+   // srunner_add_suite(sr, wal_utils_suite);
```
Jubilee101: Enable wal_utils_suite. We know which tests are failing~

test/testcases/test_wal_summary.c
```diff
@@ -98,58 +117,56 @@ pgmoneta_test_wal_summary_suite()
...
-   return pgmoneta_get_server(PRIMARY_SERVER);
+    char* chkpt_lsn = NULL;
+    uint32_t chkpt_hi = 0;
+    uint32_t chkpt_lo = 0;
+    struct query_response* qr = NULL;
+
+    /* Assuming the user associated with this server has 'pg_checkpoint' role */
+    if (pgmoneta_tsclient_execute_query(srv, "CHECKPOINT;", 0, &qr))
+    {
+        goto error;
```
Jubilee101: You should not goto error nor return error code. Imagine this function errors out, it is more informative if it immediately complains in unit test. It saves you the trouble of guessing which step goes wrong. So use assert instead. Unit test is not about handling error, instead it's about reporting as soon as possible.

test/postgresql/src/postgresql17/conf/setup.sql
ashu3103: Yeah, we don't need it for production

test/libpgmonetatest/tscommon.c
ashu3103: Okkay

ashu3103:
@Jubilee101 PTAL

test/libpgmonetatest/tscommon.c
```diff
@@ -161,4 +173,69 @@ void
 pgmoneta_test_teardown(void)
 {
     pgmoneta_memory_destroy();
-}
\ No newline at end of file
+}
+
+int
+pgmoneta_tsclient_execute_query(int srv, char* query, bool is_dummy, struct query_response** qr)
```
Jubilee101: Change name to `pgmoneta_test_execute_query`

test/libpgmonetatest/tsclient.c
```diff
@@ -33,6 +33,7 @@
 #include <configuration.h>
 #include <json.h>
 #include <management.h>
+#include <memory.h>
```
Jubilee101: Do you still need this? Otherwise remove.

test/libpgmonetatest/tscommon.c
ashu3103: Done

Jubilee101:

Issue #635, opened 2025-05-15
walinfo: Summary


jesperpedersen:
Create a summary for a WAL file or a WAL directory where the number of each WAL record is displayed, and the total.

This should be

```
pgmoneta-walinfo -S
```

jesperpedersen:
@Holo-xy Can you take this one ?

jesperpedersen:
@Holo-xy Lets just start with a simple summary of each record type, then we can do a `-S -v` format after

Holo-xy:
> [@Holo-xy](https://github.com/Holo-xy) Lets just start with a simple summary of each record type, then we can do a `-S -v` format after

So currently the outpout looks something like this
``` 
Heap | 0/2FF4F88 | 0/2FF4FD8 | 80 | 80 | 753 | off 93 flags 0x00 blkref #0: rel 1663/16389/16392 forknum 0 blk 2155
```
if we use -S 
should it look like this 
```
1 | Heap 
```
where 1 means this is the first record?
and maybe a line at the end saying something like ``` Total number of recrods: X"?

jesperpedersen:
yeah, maybe something like

```
Number | Record
---------------
     1 | Heap
---------------
     1 | Total
```

So, right aligned in first column, and left in the other

jesperpedersen:
@Holo-xy Run `pgbench` for a while, that should give you a good mix for testing the output

jesperpedersen:
@Holo-xy Yeah, it is a little bit of a PITA to find the maximum number, and maximum name of record in order to get the number of `-` correct. So, maybe start with something simpler...

Just work on it, and think about `-F` and `-o`

Holo-xy:
> [@Holo-xy](https://github.com/Holo-xy) Yeah, it is a little bit of a PITA to find the maximum number, and maximum name of record in order to get the number of `-` correct. So, maybe start with something simpler...
> 
> Just work on it, and think about `-F` and `-o`

WDYM by the maximum number and maximum name of record?

jesperpedersen:
```
1000000000 | Very loooooooooooooong record name
```

Holo-xy:
> ```
> 1000000000 | Very loooooooooooooong record name
> ```

I mean assuming that record names have a pre defined set of names then it should be easy 
and an upper limit can easily be set by dividing the file length by 24 (the length of a log header) 
but I guess starting simple in general isn't a bad idea.

jesperpedersen:
[...]

PR #673, opened 2025-06-29
[#635] walinfo: Summary

ashu3103:
#### Changes

- Introduced a new flag `-S` for walinfo summary
- Introduced a new global data structure: `RmgrTableSummary` - a mapping of resource manager and their count in parsed WAL records.
- Before the output stream is created inside the `pgmoneta_describe_walfile`, this PR would create the output stream on the top level (because we want to flush summary at the top level) in `walinfo` process.
- With summary (if `-S`) is enabled only filters accepted are `start_lsn` (`-s`) and `end_lsn` (`-e`), rest all filter flags will be ignored.

#### Usage

`./pgmoenta-walinfo -S </path/to/wal_file>`
or
`./pgmoenta-walinfo -S </path/to/wal_dir>`

@jesperpedersen PTAL

### DIFF
--- a/doc/DEVELOPERS.md
+++ b/doc/DEVELOPERS.md
@@ -420,6 +420,7 @@ Options:
   -x,  --xid         Filter on an XID
   -l,  --limit       Limit number of outputs
   -v,  --verbose     Output result
+  -S,  --summary     Show a summary of WAL record counts grouped by resource manager
   -V,  --version     Display version information
   -m,  --mapping     Provide mappings file for OID translation
   -t,  --translate   Translate OIDs to object names in XLOG records
--- a/src/include/walfile/rmgr.h
+++ b/src/include/walfile/rmgr.h
@@ -58,6 +58,7 @@ extern "C" {
 #include <stdint.h>
 
 #define PG_RMGR(symname, name, desc) {name, desc},
+#define PG_RMGR_SUMMARY(symname, name, number_of_records) {name, number_of_records},
 #define RM_MAX_ID           UINT8_MAX
 
 /**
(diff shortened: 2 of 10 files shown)


doc/DEVELOPERS.md
```diff
@@ -420,6 +420,7 @@ Options:
   -x,  --xid         Filter on an XID
   -l,  --limit       Limit number of outputs
   -v,  --verbose     Output result
+  -S,  --sumary      Show a summary of WAL record counts grouped by resource manager
```
jesperpedersen: summary

src/include/walfile/rmgr.h
```diff
@@ -75,33 +76,25 @@ struct rmgr_data
...
+   char* name;             /**< The name of the resource manager */
+   int number_of_records;  /**< The number of records of a specific type */
 };
 
+/**
+ * Table of resource managers. Each entry corresponds to a specific RMGR identified by an ID.
+ */
+extern struct rmgr_data RmgrTable[RM_MAX_ID + 1];
+extern struct rmgr_summary RmgrSummaryTable[RM_MAX_ID + 1];
+
```
jesperpedersen, line 97: Why did you move these ? And, no camel-case -> rmgr_table + rmgr_summary_table

src/libpgmoneta/walfile.c
```diff
@@ -390,31 +389,33 @@ pgmoneta_describe_walfile(char* path, enum value_type type, char* output, bool q
       goto error;
    }
 
-   if (output == NULL)
+   if (out)
```
jesperpedersen: Always use explicit `!= NULL` and `== NULL` for pointers

src/libpgmoneta/walfile/rmgr.c
```diff
@@ -0,0 +1,79 @@
...
+   PG_RMGR_SUMMARY(RM_GIN_ID, "Gin", 0)
+   PG_RMGR_SUMMARY(RM_GIST_ID, "Gist", 0)
+   PG_RMGR_SUMMARY(RM_SEQ_ID, "Sequence", 0)
+   PG_RMGR_SUMMARY(RM_SPGIST_ID, "SPGist", 0)
+   PG_RMGR_SUMMARY(RM_BRIN_ID, "BRIN", 0)
+   PG_RMGR_SUMMARY(RM_COMMIT_TS_ID, "CommitTs", 0)
+   PG_RMGR_SUMMARY(RM_REPLORIGIN_ID, "ReplicationOrigin", 0)
+   PG_RMGR_SUMMARY(RM_GENERIC_ID, "Generic", 0)
+   PG_RMGR_SUMMARY(RM_LOGICALMSG_ID, "LogicalMessage", 0)
+};
```
jesperpedersen, line 79: Yes, I don't think this

src/walinfo.c
```diff
@@ -622,6 +641,94 @@ main(int argc, char** argv)
...
+   {
+      int max_rmgr_name_length = 6; // Corresponds to the length of string `Record`
+      int max_digit_length = 6; // Corresponds to the length of string `Number`
+      int total_records = 0;
+      char* number_string = NULL;
+      char* rmgr_name = NULL;
+
+      for (int i = 0; i < RM_MAX_ID + 1; i++)
+      {
+         if (!RmgrSummaryTable[i].number_of_records) continue;
```
jesperpedersen: Formatting - run uncrustify

src/walinfo.c
```diff
@@ -622,6 +641,94 @@ main(int argc, char** argv)
...
+            rmgr_name = RmgrSummaryTable[i].name;
+            fprintf(out, "%*s%s | %s%*s\n", max_digit_length - (int)strlen(number_string), " ", 
+                  number_string, rmgr_name, max_rmgr_name_length - (int)strlen(rmgr_name), " ");
+            free(number_string);
+            number_string = NULL;
+
+            /* Revert the state of RmgrSummaryTable */
+            RmgrSummaryTable[i].number_of_records = 0;
+         }
+
```
jesperpedersen, line 723: Print a horizontal line above `Total`

jesperpedersen (changes requested):

src/include/walfile/rmgr.h
ashu3103: > Why did you move these ?

The data structures inside `rmgr.h` are included in multiple `.c` files so if we have `extern` for `rmgr_data` and `rmgr_summary` to avoid multiple definition problem thrown by linker. To mitigate this I have to declare them in `rmgr.h` and define them in some root file (why not root file `rmgr.c`)

> And, no camel-case -> rmgr_table + rmgr_summary_table

No specific reason to do so, `RmgrTable` was present from starting so I thought it might be the convention for this file, anyways I'll revert it.

src/libpgmoneta/walfile/rmgr.c
ashu3103: Can you please elaborate a bit..

jesperpedersen: Yes, but we have `#ifndef PGMONETA_RMGR_H` - it is a problem of circular dependencies ?

jesperpedersen: Did you try a `<walfile/rm_defs.h` with the `struct`s and include that file along with the others in `rmgr.h`

jesperpedersen: Or go in a `static int rmgr_init(...)` route

ashu3103: > Yes, but we have `#ifndef PGMONETA_RMGR_H` - it is a problem of circular dependencies ?

No a case of multiple definitions, since `walinfo.c` and `wal_reader.c` will include `walfile/rmgr.h`, they will load/define the struct independently and hence multiple definitions

ashu3103: > Or go in a `static int rmgr_init(...)` route

No, I haven't, let me check

ashu3103: > Or go in a `static int rmgr_init(...)` route

I have checked that this would not share states across files (sharing states across files is we need)

jesperpedersen: Ok - let me think - assume we are going in the `extern` direction

ashu3103: > Ok - let me think - assume we are going in the `extern` direction

Sure

ashu3103: Hi @jesperpedersen 

Any updates?

jesperpedersen:
Rebased, and merged.

PR #697, opened 2025-07-24
[#688] Fixed the first record's offset calculation

### LINKED ISSUE #688: walinfo: "Invalid" partial WAL file
walinfo can't parse the attached file - pg_waldump can

Mohab96:
Resolves #688 

This patch should fix the calculation of the offset of the first record after the long header correctly when there are remaining bytes from the previous segment.


Mohab96:
PTAL @jesperpedersen @sh-soltanpour

jesperpedersen:
Merged.

PR #707, opened 2025-07-26
Use select for HTTP test suite

bassamadnan:
`accept` was running into race issues.  This uses `select` with some break mechanism and fixes the issues. Addressing #706


bassamadnan:
cc: @ashu3103

jesperpedersen:
Merged.

Issue #709, opened 2025-07-30
Request a relation file from the server


ashu3103:
**Is your feature request related to a problem? Please describe.**

Assuming we have our infrastructure for detecting the changes/modifications in the cluster during a range of `LSN`s with us (Refer #649). We now want to extract those changes (either the complete changed file or better only the changed blocks of a file) from the `postgres` server. 

**Describe the solution you'd like**

A clear solution path is yet to be decided but here are some breadcrumbs for some potential approaches:

- Get the database cluster directory and using `relfilenode` information copy the associated file. [**require SUPERUSER role**]
- Leverage the `pgmoneta_ext` functions to receive a file. [**require SUPERUSER permissions**]
- Perform a base backup, keep it as a  local temporary copy and then extract relevant files and blocks out of it and prune the local base backup copy. [**slow compute**]

### Approach-1

Use the admin function `pg_read_binary_file()` which is used to provide native access to files on the machine hosting the server. [**Some [sources](https://pgpedia.info/p/pg_read_binary_file.html) claim it works only for local cluster, have to check**]

**Function Signature:**

`pg_read_binary_file(filename text [, offset bigint, length bigint ] [, missing_ok boolean ] ) → bytea`

**Limitations:**

- Only files within the database cluster directory and the `log_directory` can be accessed.
- By default, only `SUPERUSER` can execute this, but some non-root user with role `pg_read_server_files` can also perform this command.
- Grant users the `EXECUTE` privilege on `pg_read_binary_file()`, when granting privilege on these functions, note that the table entries showing optional parameters are mostly implemented as several physical functions with different parameter lists. Privilege must be granted separately on each such function, if it is to be used (therefore: `EXECUTE` on `pg_read_binary_file(text)` is different from `EXECUTE` on `pg_read_binary_file(text, bigint, bigint, boolean)`)

**Highlights:**
- Since the function can accept `offset` and `length` as parameters, we can request a block (modified) from the relation file. [**Most Desired**]
- Don't really need the absolute path to database cluster, relative path inside the database cluster can be used to make a query for example `pg_read_binary_file('base/1/112')` will work.

**A glimpse:**

- GRANT the required roles and permissions to `repl` user.
```
GRANT pg_read_server_files TO repl;
GRANT EXECUTE on function pg_read_binary_file(text, bigint, bigint, boolean) to repl;
GRANT EXECUTE on function pg_read_binary_file(text) to repl;
```

[...]

PR #716, opened 2025-08-09
[#709] Request a relation file from the postgres server

ashu3103:
Feature #709 

This patch contains the API to fetch the binary content of a relation file from `postgres` server.

@jesperpedersen  @Jubilee101 PTAL

### DIFF
--- a/src/libpgmoneta/server.c
+++ b/src/libpgmoneta/server.c
@@ -53,6 +53,9 @@ static int get_checksums(SSL* ssl, int socket, bool* checksums);
 static int get_segment_size(SSL* ssl, int socket, size_t* segsz);
 static int get_block_size(SSL* ssl, int socket, size_t* blocksz);
 static int get_summarize_wal(SSL* ssl, int socket, bool* sw);
+static int has_user_role(SSL* ssl, int socket, char* usr, char* role, bool* has_role);
+static int has_execute_privilege(SSL* ssl, int socket, char* usr, char* func_name, bool* has_privilege);
+static int transform_hex_bytea_to_binary(char* hex_bytea, uint8_t** out, int* len);
 static int process_server_parameters(int server, struct deque* server_parameters);
 
 static bool is_valid_response(struct query_response* response);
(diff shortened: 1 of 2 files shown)


src/libpgmoneta/server.c
```diff
@@ -261,6 +265,120 @@ pgmoneta_server_verify_connection(int srv)
...
+      pgmoneta_log_error("Unexpected number of columns in query response");
+      goto error;
+   }
+
+   memset(bytea_data_buffer, 0, DEFAULT_BURST);
+
+   /* Note: we get data in hex format */
+   snprintf(bytea_data_buffer, DEFAULT_BURST, "%s", response->tuples->data[0]);
+
+   /* Transform it to binary */
```
ashu3103, line 360: TODO: can add support for legacy escape bytea type

src/libpgmoneta/server.c
```diff
@@ -261,6 +265,120 @@ pgmoneta_server_verify_connection(int srv)
...
+   if (response->number_of_columns != 1)
+   {
+      pgmoneta_log_error("Unexpected number of columns in query response");
+      goto error;
+   }
+
+   memset(bytea_data_buffer, 0, DEFAULT_BURST);
+
+   /* Note: we get data in hex format */
+   snprintf(bytea_data_buffer, DEFAULT_BURST, "%s", response->tuples->data[0]);
```
Jubilee101, line 358: Is there a way we can get binary data directly?

src/libpgmoneta/server.c
```diff
@@ -760,6 +878,203 @@ is_valid_response(struct query_response* response)
...
+   *len = binary_len;
+   return 0;
+error:
+
+   free(binary_out);
+   return 1;
+}
+
+static int
+hex_to_int(char c)
```
Jubilee101: Try sscanf?

src/libpgmoneta/server.c
Jubilee101, line 358: And are we sure we get all the data in one burst?

Jubilee101:
Should we profile it? How long does it take to stream a file?

src/libpgmoneta/server.c
ashu3103: Let me check...

ashu3103: I was assuming that we will be requesting the server in a block-by-block fashion (8KB at a time), so the `DEFAULT_BURST` shall always accommodate the requested size.

ashu3103:
> Should we profile it? How long does it take to stream a file?

Sure, it would be of great help, if u could tell me how to perform profiling, what tools to use etc

jesperpedersen:
@ashu3103 You can use `perf` and https://github.com/brendangregg/FlameGraph

src/libpgmoneta/server.c
ashu3103: @Jubilee101, the target function returns `bytea` - which is the hex representation of the underlying binary data. I don't think there is a way to get binary data directly (we can however change the `bytea` format to either `hex` or `escape`, but directly getting binary from this func is simply not possible).

src/libpgmoneta/server.c
ashu3103: Done PTAL

Jubilee101:
> Should we profile it? How long does it take to stream a file?

Or simply get the start time and end time before and after. We are just looking for a ballpark figure.

src/libpgmoneta/server.c
Jubilee101: I see, my concern is the time. If the translation takes an acceptable amount of time then we are good.

Jubilee101:
This looks good to me @jesperpedersen

Jubilee101:
Merged @ashu3103 , thanks for your contribution! Note that your previous patch still has to wait for a bit, we are still adjusting the testing framework and fixing some bugs. Once we are done you can rebase and refactor your tests. Thanks for your patience!

ashu3103:
> we are still adjusting the testing framework and fixing some bugs. Once we are done you can rebase and refactor your tests. 

Sure, tag me if any help is needed from my end

> Thanks for your patience!

Not an issue, Thanks

Issue #735, opened 2025-08-30
Get the `endLSN` for WAL Summarizer


ashu3103:
**Is your feature request related to a problem? Please describe.**

Our [WAL Summarizer](https://github.com/pgmoneta/pgmoneta/issues/649) needs two LSNs to summarize WAL records in that range. In this issue, our focus will be to procure the endLSN for the summarizer which is nothing but a `CHECKPOINT LSN` that occur after starting incremental backup workflow.

Note: We are emphasizing on `CHECKPOINT` because that way summarization would be more reliable as we can be sure that every change before this `LSN` is persistent.

**Describe the solution you'd like**

After getting inside the incr backup workflow, we are left with two choices to get CHECKPOINT LSN:

### Active Approach

Force a `CHECKPOINT` in the cluster using `CHECKPOINT;` command or using the extension function: `pgmoneta_ext_checkpoint()`

However the only downside of this approach is that we need special privileges to execute it:

- We need `pg_checkpoint` role to execute it for PostgreSQL +15
- We need `SUPERUSER` role to execute it for PostgreSQL < 15

### Passive Approach

Wait for the checkpoint, we can track the change of `checkpoint_lsn` using `pg_control_checkpoint()` function

The only downside of this approach is that checkpoint timeout can stall the workflow for a long time. The default `checkpoint_timeout` is 5mins and changing the checkpoint_timeout to a lower value will compromise performance (since there will frequent file syncs).

p.s. Just some thoughts, I have not given any thought to the appropriate solution or a workaround yet

@Jubilee101 @jesperpedersen PTAL

jesperpedersen:
@ashu3103 I think we have to go the active route and use the extension to force the checkpoint - it isn't ideal, but that is the state of the union. So, go that route please

ashu3103:
> [@ashu3103](https://github.com/ashu3103) I think we have to go the active route and use the extension to force the checkpoint - it isn't ideal, but that is the state of the union. So, go that route please

So, shall we relax the privilege levels for `pgmoneta_ext_checkpoint` or use some custom roles so that we don't need to grant SUPERUSER to `repl`

jesperpedersen:
@ashu3103 For PostgreSQL 13/14 we will require `SUPERUSER`, and PostgreSQL 15+ its `pg_checkpoint`. So, that is a documentation thing plus a guard for the incremental workflow.

[...]

PR #740, opened 2025-09-02
[#735] Get endLSN of WAL Summarizer

ashu3103:
### Work In Progress

For the endLSN of WAL Summarizer, we force a checkpoint into the server at the time of starting of incremental backup workflow (This checkpoint LSN is our `endLSN`).

#### TODO

- [ ] Refactor: The logic is duplicated in the `test_wal_summary` testsuite
- [ ] Documentation: document the role required to perform checkpoint and security guard applied


jesperpedersen:
Yes, keep going in this direction

Jubilee101:
> For the endLSN of WAL Summarizer, we force a checkpoint into the server at the time of starting of incremental backup workflow (This checkpoint LSN is our endLSN).

@ashu3103  I think this may be able to do the trick but have you checked https://www.interdb.jp/pg/pgsql10/01.html#1012-pg_backup_stop-ver14-or-earlier-pg_stop_backup? Any particular reason why we can't just insert a log and switch wal instead of issueing a checkpoint?

ashu3103:
> > For the endLSN of WAL Summarizer, we force a checkpoint into the server at the time of starting of incremental backup workflow (This checkpoint LSN is our endLSN).
> 
> 
> 
> @ashu3103  I think this may be able to do the trick but have you checked https://www.interdb.jp/pg/pgsql10/01.html#1012-pg_backup_stop-ver14-or-earlier-pg_stop_backup? Any particular reason why we can't just insert a log and switch wal instead of issueing a checkpoint?

Hi @Jubilee101,

The way I was thinking is that while performing backup we will be fetching blocks from the postgres server, which we need to be flushed and updated before the backup start LSN (so a checkpoint before backup start)

I think you are confusing between the endLSN here with the stop LSN of the backup (This the endLSN of the WAL summarizer)

Jubilee101:
> The way I was thinking is that while performing backup we will be fetching blocks from the postgres server, which we need to be flushed and updated before the backup start LSN (so a checkpoint before backup start)
> I think you are confusing between the endLSN here with the stop LSN of the backup (This the endLSN of the WAL summarizer)

I see. Yes this sounds right, thanks for clarifying things. I think you should directly get backup start LSN and use that as your summarizer end location, since the same REDO point is used for both tasks. Tread carefully here as some other things seem to be going on during this process: https://www.interdb.jp/pg/pgsql10/01.html#1011-pg_backup_start-ver14-or-earlier-pg_start_backup

ashu3103:
> I think you should directly get backup start LSN and use that as your summarizer end location, since the same REDO point is used for both tasks

Got it

ashu3103:

src/libpgmoneta/server.c
```diff
@@ -398,6 +399,135 @@ pgmoneta_server_read_binary_file(int srv, SSL* ssl, char* relative_file_path, in
...
+   }
+q:
+   pgmoneta_query_execute(ssl, socket, query_msg, &response);
+
+   if (!is_valid_response(response))
+   {
+      pgmoneta_free_query_response(response);
+      response = NULL;
+
+      SLEEP(5000000L);
```
jesperpedersen, line 467: ?

jesperpedersen (changes requested):

jesperpedersen:
@ashu3103 Have tried these queries with a user that doesn't have the `SUPERUSER` bit ?

src/libpgmoneta/server.c
ashu3103: This is just our standard retry logic. In case there is any network hiccups or server congestion.

ashu3103:
> @ashu3103 Have tried these queries with a user that doesn't have the `SUPERUSER` bit ?

Yeah, I have tested it for version 17, and its working fine

jesperpedersen:
@ashu3103 I meant a PostgreSQL 13 role - where the user skipped the documentation and didn't apply the necessary changes to the role

ashu3103:
> @ashu3103 I meant a PostgreSQL 13 role - where the user skipped the documentation and didn't apply the necessary changes to the role

I'll do that

ashu3103:
> @ashu3103 I meant a PostgreSQL 13 role - where the user skipped the documentation and didn't apply the necessary changes to the role

@jesperpedersen, Done with the testing with versions 13 and 15, working as expected!

jesperpedersen:
Merged.

Issue #746, opened 2025-09-10
Memory Leak while converting bytea to binary for a chunk not present in file


ashu3103:
**Describe the bug**

Missed handling of the case when requesting a chunk of a relation file from the server which is not present or out of bounds f.ex. if the file size is 8192 and we are requesting 100 bytes at offset 8192. Ideally this is not an error, but still needs special handling to avoid allocating a memory buffer.

PR #747, opened 2025-09-10
[#746] Memory Leak while converting bytea to binary for a chunk not present

ashu3103:
Fix: #746 

Added an additional check for hex length, so if no data is received just return. This avoids calling the `malloc` which will then cause leaks.


jesperpedersen:
Merged.

PR #749, opened 2025-09-10
[#748] Add physical replication slot creation for replication user

### LINKED ISSUE #748: Server `Online: false` while running inside docker
When running postgres cluster using docker as explained [here](https://github.com/pgmoneta/pgmoneta/blob/main/test/postgresql/src/postgresql17/README.md) and then try
```
sudo -u pgmoneta pgmoneta-cli -c "/etc/pgmoneta/pgmoneta.conf" status
```

I get this 
```
➜  postgresql17 git:(walfilter) ✗ sudo -u pgmoneta pgmoneta-cli -c "/etc/pgmoneta/pgmoneta.conf" status
Header:
  ClientVersion: 0.19.0
  Command: status
  Compression: none
  Encryption: none
  Output: text
  Timestamp: 20250910184259
Outcome:
  Status: true
  Time: 00:00:0.0004
Request: {}
Response:
  FreeSpace: 21.76GB
  NumberOfServers: 1
  ServerVersion: 0.19.0
  Servers:
    - Checksums: true
      HotStandbySize: 0.00B
      NumberOfBackups: 0
      Online: false
      Primary: true
      RetentionDays: 7
      RetentionMonths: Unspecified
      RetentionWeeks: Unspecified
      RetentionYears: Unspecified
      Server: primary
      ServerSize: 16.19MB
      Workers: 0
      WorkspaceFreeSpace: 0.00B
  TotalSpace: 144.50GB
  UsedSpace: 16.19MB
  Workers: 0
```

where you can see clearly the server's `Online: false`

Mohab96:
Resolves #748

The problem was that the line responsible for creating the replication slot was missing.


jesperpedersen:
Merged.

Mohab96:
Another thing I noticed while working on this PR, if someone works mainly with pgmoneta on their own postgres cluster, everything is good, but if you use the cluster from the docker container, the pgmoneta server wouldn't know, but the postgres cluster will start throwing stuff like this
```
2025-09-10 23:22:06 ERROR message.c:186 error response message: requested starting point 0/B000000 is ahead of the WAL flush position of this server 0/1B9E828 (SQLSTATE code: XX000)
2025-09-10 23:22:06 DEBUG message.c:201 error response field type: R, message: 
2025-09-10 23:22:06 DEBUG message.c:201 error response field type: V, message: ERROR
2025-09-10 23:22:06 DEBUG message.c:201 error response field type: F, message: walsender.c
2025-09-10 23:22:06 DEBUG message.c:201 error response field type: L, message: 969
2025-09-10 23:22:06 DEBUG message.c:201 error response field type: R, message: StartReplication
```

where the WAL stream in `/wal` ends at some LSN different than what postgres expects, so to avoid these kinds of problems, you should clear `/wal` and `/backup` directories to have a fresh start.

jesperpedersen:
@Mohab96 Yes, the docker setup could use an update - lets revisit once PostgreSQL 18 is out next week

Mohab96:
I'd suggest using docker compose to start both containers for pgmoneta and postgres to have some consistency.

Jubilee101:
> where the WAL stream in /wal ends at some LSN different than what postgres expects, so to avoid these kinds of problems, you should clear /wal and /backup directories to have a fresh start.

That's why we always use /tmp as a clean base directory when running tests. And don't forget that you are not supposed to treat test postgres container as the primary server. Instead you should configure another server such as [test], so that the wal ends up there.

And any particular reason you are starting the container and interacting with it instead of using the unit test infra? Feel free to make suggestions as we are still working on improving it.

> I'd suggest using docker compose to start both containers for pgmoneta and postgres to have some consistency.

Are you suggesting moving pgmoneta into a container for testing? We deliberately avoid that for faster test and rebuild, and easier debugging. Because check doesn't work well inside the container. You can't print out things to see what's wrong. And we were having a hard time generating logs and moving it out of the pod. I would keep things as is for now and improve the current infra.

Mohab96:
> That's why we always use /tmp as a clean base directory when running tests. And don't forget that you are not supposed to treat test postgres container as the primary server. Instead you should configure another server such as [test], so that the wal ends up there.

Got that.

> And any particular reason you are starting the container and interacting with it instead of using the unit test infra? Feel free to make suggestions as we are still working on improving it.

If you're asking why I used the container instead of my main postgres cluster for testing my patch, that's because to test the patch, I want to start a cluster after replacing its wal stream, which means if I did something wrong, I'll have to set it up from scratch again, so it was time-consuming. Turned out that using docker was even more time-consuming, and I tested on my main cluster at the end ^_^.

> Are you suggesting moving pgmoneta into a container for testing? We deliberately avoid that for faster test and rebuild, and easier debugging. Because check doesn't work well inside the container. You can't print out things to see what's wrong. And we were having a hard time generating logs and moving it out of the pod. I would keep things as is for now and improve the current infra.

When I used postgres from inside docker, pgmoneta didn't know anything about this and continued to work normally, although it contains WAL stream and backups for a different cluster, so either start pgmoneta and postgres inside containers to ensure a fresh start for both, or put some warning in the docs about this scenario so it doesn't corrupt the postgres cluster or pgmoneta server in any way.

Issue #743, opened 2025-09-08
Serialization of modified blocks in incremental backup files


ashu3103:
**Is your feature request related to a problem? Please describe.**

Once we have all the modified block data of all the relation files in a cluster, we now have to serialize/store them in some format so that they can be used along with base backups for restoration.

**Describe the solution you'd like**

We can mimic the `INCREMENTAL` file format used by PostgreSQL +17 to store modified blocks. The idea is that it will then be compatible with our restoration logic (we can reuse most of the restore logic).

### Contents of the file

The file contains a header block at the starting followed by modified blocks, each block is 8KB (by default).

**Header Block**
- Magic Number: The header begins with “0xd3ae1f0d” to identify it as an incremental backup file.
- num_incremental_block: The number of modified blocks
- truncation_block_length: generally, the total number of blocks in the relation
- list of modified block numbers: A list of modified block numbers

Padded with `0s` to make header a multiple of `block_size`.

**Modified Blocks**

The actual data blocks of a relation file requested from the server.

### Convention of filename

For the nomenclature of an incremental backup file (for a relation) simply prefix the name of the base relation file with `INCREMENTAL.`; For instance, assuming the relation file for Table t1 (`OID=16551`) is stored in the full backup and the incremental backup file named ‘INCREMENTAL.16551’ is in the incremental backup

## Implementation Notes

- The block number from the BRT may belong to a different segment of a relation file `<relationfileoid>.<segno>`. This is important to find the filename when we are requesting a chunk from file
- It might happen that a relation file is truncated after initiation of incremenal backup, but we still get the block numbers from the summarizer (due to obvious reasons), in that case, don't panic, just pad the incremental file with zero upto the expected size promised to the caller (`expected_size = num_of_blocks * block_size`).

@Jubilee101 @jesperpedersen PTAL

ashu3103:
Hi @jesperpedersen @Jubilee101,

The draft for this issue is almost ready, I however have some doubts (also currently I am ignoring tablespaces)

- Except `base/` what all files/directories inside the cluster we have to target like do we need an incremental backup for `global/` directory.
- Same for fork numbers, which one to include for incremental (others for full/no backup).

Jubilee101:
> (also currently I am ignoring tablespaces)

That is ok. 

> Except base/ what all files/directories inside the cluster we have to target like do we need an incremental backup for global/ directory.

[...]

PR #751, opened 2025-09-16
[#743] Serialization of modified blocks in incremental backup files

ashu3103:
### DRAFT | Don't Merge

This draft contain the helper function for the workflow `write_incremental_file` to serialize the modified blocks. This draft is not supposed to be merged until the workflow is in place.

@jesperpedersen, @Jubilee101 Please review the static helper function only (rest of the code is just there to build the project).

### DIFF
--- a/src/libpgmoneta/wf_backup_incremental.c
+++ b/src/libpgmoneta/wf_backup_incremental.c
@@ -1,0 +1,360 @@
+/*
+ * Copyright (C) 2025 The pgmoneta community
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
+ * are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this list
+ * of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice, this
+ * list of conditions and the following disclaimer in the documentation and/or other
+ * materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its contributors may
+ * be used to endorse or promote products derived from this software without specific
+ * prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+ * OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+ * THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT
+ * OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR
+ * TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/* pgmoneta */
+#include <pgmoneta.h>
+#include <backup.h>
+#include <logging.h>
+#include <network.h>
+#include <security.h>
+#include <server.h>
+#include <utils.h>
+#include <walfile/wal_reader.h>
+#include <workflow.h>
+
+/* system */
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+/* fetch/compute these from server configuration inside create workflow */
+size_t block_size;     // size of each block (default 8KB)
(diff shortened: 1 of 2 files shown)


src/libpgmoneta/wf_backup_incremental.c
```diff
@@ -0,0 +1,360 @@
...
+    wf = (struct workflow*)malloc(sizeof(struct workflow));
+
+    if (wf == NULL)
+    {
+        return NULL;
+    }
+
+    wf->name = &incr_backup_name;
+    wf->setup = &pgmoneta_common_setup;
+    wf->execute = &incr_backup_execute;
```
Jubilee101, line 86: Currently support for incremental backup for postgres 17+ is inside `pgmoneta_create_basebackup`. Do you think you can move that part to pgmoneta_create_incremental_backup now that you introduced it?

src/libpgmoneta/wf_backup_incremental.c
```diff
@@ -0,0 +1,360 @@
...
+    }
+
+    block_size = config->common.servers[server].block_size;
+    segment_size = config->common.servers[server].segment_size;
+    rel_seg_size = config->common.servers[server].relseg_size;
+
+    /* TODO: start backup */
+
+    /* write the INCREMENTAL file */
+    write_incremental_file(server, ssl, socket, filename, rlocator, frk,
```
Jubilee101, line 152: Just one file?

src/libpgmoneta/wf_backup_incremental.c
ashu3103: @Jubilee101, yes once the basic outline of the new workflow is ready, I am planning to move that part to this workflow.

src/libpgmoneta/wf_backup_incremental.c
ashu3103: @Jubilee101 @jesperpedersen,

The BRT contains entries specific to (`rlocator`, `fork_number`) pair (which simply denotes a relation file in a cluster), so, for a particular WAL summarization ->
- go through each entry
- create the file path, where we have to write the backup data (full/incr)
- pass the above calculated file path into `write_incremental_file()` [we can however, resolve the file path inside this function, since we are passing (`rlocator`, `fork_number`) pair, but by providing filename, this function don't have to worry about file/path existence, permissions etc]
- Additionally, we can use manifest of the previous backup and create symlink for the files taht were not modified [Work of Later Phase].

This was my game plan. Any thoughts or suggestions?

Jubilee101: > Additionally, we can use manifest of the previous backup and create symlink for the files taht were not modified [Work of Later Phase].

You are still creating incremental files when the files are not changed. They just won't contain any blocks. For files not in incremental directory, wf_link will handle the symlink creation. You will find it much easier this way when implementing the logic, and aligning to our current incremental restore logic.

Other steps look alright to me~

ashu3103: Hmm, actually what I was thinking is that first we go through all brt entries (**Note:** BRT will only contain blocks from incremented files) so we can only create incremental files for modified relation files (This will include `all the modified files`+`any additional file added after preceding backup`). Once that is done, we will create symlinks (by comparing manifest) inside the incr backup for only those files that were not changed.

But as you pointed out comparing manifest and making symlinks is handled by a separate workflow therefore I will have to omit this step
|-> `Additionally, we can use manifest of the previous backup and create symlink for the files taht were not modified [Work of Later Phase].`

ashu3103:
@Jubilee101, I had a doubt, is it okay if we incrementally backup a new file (which was created after preceding backup and before current backup)?

Jubilee101:
Hi @ashu3103, excuse me for the late reply. 

> But as you pointed out comparing manifest and making symlinks is handled by a separate workflow therefore I will have to omit this step

Correction here: we are skipping symlinking incremental files only, for full files under incremental directory we are still symlinking them.
 
Yes, symlinks will only work on non-incremental files. We actually carefully thought this through when designing this part -- inside the incremental directory, most files will be different from their predecessors, since they are by nature recording incremental changes, so for those we can't symlink them anyways. The only exceptions are those files not changed at all, but then the incremental files will only contain very limited meta datas, and we can afford that. 

The reason we are skipping symlinking incremental file is that: imaging you have a chain of 3 backups, and inside there is a chain of 3 files: full_file -> incr_file_1 -> incr_file_2. When we delete the full backup, we do a process called "backup rollup", essenstially merging the full backup upwards to incr backup 1, making incr backup 1 the new full backup, so that we don't break the entire incremental backup chain. And as you can imagine during that process all the incr files in incr backup 1 are gone, turned into full files. So if incr_file_2 here is a symlink of incr_file_1, it will point to nothing.

And don't forget symlink chain and incremental chain are two different concepts. Users can do full backup - incremental backup - incremental backup - full backup - full backup again - incremental backup. So there are actually 3 chains of backups, but maybe only one symlink chain. Think about how you are going to manage the conflicts and overlaps among those two types of chains and you'll see why we are not symlinking the incremental files -- Does it make more sense to symlink a full backup to the previous full backup, which is maybe 10 incremental backups behind, or to the preceding incremental backup? What if the incremental backup gets rolled up, to another full or incremental backup?

And you may wonder, what if the files are not changed in the first place, then why couldn't I just symlink it, not as an incremental file, just a full file? You may, but don't forget the file not changed this time could have previously already been kept as an incremental file, and then we fall back to the previous question.

> I had a doubt, is it okay if we incrementally backup a new file (which was created after preceding backup and before current backup)?

I don't know how your wal summary is going to pick the new files up, I thought the summary ends on the checkpoint made before the backup starts. Anyways, I think it's fine if something is created after the backup process starts, and we failed to pick them up. Since when starting on a backup, postgres always goes through the WAL segments and redo the changes until end LSN. So anything missing will eventually be consistent. But that is also why the backup start/end lsn is so important.

jesperpedersen:
@ashu3103 Is this still a draft ?

ashu3103:
> @ashu3103 Is this still a draft ?

This feature is included in #782, you might close it

Issue #753, opened 2025-09-18
Add a utils function to translate string lsn to integer lsn


ashu3103:
Tired of writing the logic for translating a `char*` lsn to `uint64_t` :)

PR #754, opened 2025-09-18
[#753] Add a utils function to translate string lsn to integer lsn

ashu3103:
The new workflow requires this util function a couple of times so I feel the need to incorporate it.


jesperpedersen:
Merged.

Issue #755, opened 2025-09-18
Refactor WAL Summarize API


ashu3103:
Currently the API serializes the summarization output to the disk, but for the purpose of incremental backup we don't need the overhead of first writing and then reading the BRT in the same workflow. So, we can make it hybrid (support for both serialize to disk & return the BRT pointer to the caller).

PR #756, opened 2025-09-18
[#755] Refactor WAL Summarizer

ashu3103:
Refactor #755 

### DIFF
--- a/src/libpgmoneta/walfile/wal_summary.c
+++ b/src/libpgmoneta/walfile/wal_summary.c
@@ -49,13 +49,9 @@ static char* summary_file_name(uint64_t s_lsn, uint64_t e_lsn);
  * - start_lsn and end_lsn are always the starting of a WAL record and not the middle of WAL record
  */
 int
-pgmoneta_summarize_wal(int srv, char* dir, uint64_t start_lsn, uint64_t end_lsn)
+pgmoneta_summarize_wal(int srv, char* dir, uint64_t start_lsn, uint64_t end_lsn, block_ref_table** b)
 {
-   char* summary_dir = NULL;
-   char* summary_filename = NULL;
    char* wal_dir = NULL;
-   char tmp_file[MAX_PATH] = {0};
-   char file[MAX_PATH] = {0};
    block_ref_table* brt = NULL;
 
    /**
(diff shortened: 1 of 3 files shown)


src/libpgmoneta/walfile/wal_summary.c
```diff
@@ -104,43 +104,50 @@ pgmoneta_summarize_wal(int srv, char* dir, uint64_t start_lsn, uint64_t end_lsn)
    free(partial_record);
    partial_record = NULL;
 
-   // /* Generate the WAL summaries */
-   summary_dir = pgmoneta_get_server_summary(srv);
-   if (!pgmoneta_ends_with(summary_dir, "/"))
+   if (to_disk)
```
Jubilee101: I would rather have this part separated out to a `pgmoneta_wal_summary_save()`. Unless any particular reason you are refactoring this way?

ashu3103: > Unless any particular reason you are refactoring this way

Not really, just trying to put new changes into existing ones.

> I would rather have this part separated out to a `pgmoneta_wal_summary_save()`

That's a good idea, we can separate this logic out (anyways we won't be using it).

ashu3103:
@Jubilee101 PTAL

src/libpgmoneta/walfile/wal_summary.c
```diff
@@ -104,7 +100,27 @@ pgmoneta_summarize_wal(int srv, char* dir, uint64_t start_lsn, uint64_t end_lsn)
...
+   return 0;
+
+error:
+   free(wal_dir);
+   pgmoneta_brt_destroy(brt);
+   return 1;
+}
+
+int
+pgmoneta_wal_summary_save(int srv, uint64_t s_lsn, uint64_t e_lsn, block_ref_table* brt)
```
Jubilee101, line 116: Is it a good idea that `s_lsn` and `e_lsn` are two attributes of the `block_ref_table`? We set it when generating the table and then we don't have to pass them in again when saving it.

ashu3103: Okay, but it only make sense when we use BRT only for WAL sumarization (becoz BRT was supposed to be independent from WAL summarization). However, we can achieve the above if its important.

Jubilee101: I see. Let's keep it the way it is then.

Jubilee101:

Issue #757, opened 2025-09-18
Store the information about maximum block number modified in a BRT entry


ashu3103:
The issue is, once we have `BRT` in our incremental workflow, we iterate its entries to get to the relation file, but there is a catch, a relation file can be present in the form of segments inside the data directory like a relation file with `oid = 516` can be present as -

`516`, `516.1`, `516.2`.... (depending on the segment size, default `1GB`).

Back to our workflow, if we iterate each `brt` entry, we will get results from all the segments, therefore to find a particular segment or a bound (max which segment is present) we need to have an information about maximum block number modified for that relation file.

So, the idea is to maintain a field inside `brt` entry to track this **maximum block number modified** which will get its value at the time of `BRT` creation.

PR #759, opened 2025-09-19
[#757] Store the information about maximum block number modified

ashu3103:
Need this field to get an idea of total number of segments for a relation file using the formula:
```
segments = max_block_number / rel_seg_size + 1
```

If this field was not present, we have to iterate through the chunks in the BRT entry.

Note: we are not serializing this field and the reason is, on reading a serialized BRT, the field is recalculated on the fly.

### DIFF
--- a/src/libpgmoneta/brt.c
+++ b/src/libpgmoneta/brt.c
@@ -93,6 +93,7 @@ pgmoneta_brt_set_limit_block(block_ref_table* brt, const struct rel_file_locator
    {
       /* No existing entry, set the limit block for this relation block and initialize other fields */
       brt_entry->limit_block = limit_block;
+      brt_entry->max_block_number = -1;
       brt_entry->nchunks = 0;
       brt_entry->chunk_size = NULL;
       brt_entry->chunk_usage = NULL;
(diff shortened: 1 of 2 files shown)


src/libpgmoneta/brt.c
```diff
@@ -93,6 +93,7 @@ pgmoneta_brt_set_limit_block(block_ref_table* brt, const struct rel_file_locator
    {
       /* No existing entry, set the limit block for this relation block and initialize other fields */
       brt_entry->limit_block = limit_block;
+      brt_entry->max_block_number = -1;
```
Jubilee101: `block_number` is `uint`, it is a little weird to set it to -1. Instead you can have a constant `InvalidBlockNumber` set to `0xFFFFFFFF`. Even though I guess the outcome is the same.

ashu3103: My bad, I'll fix that

Jubilee101:
LGTM, can you rebase? @ashu3103

ashu3103:
> LGTM, can you rebase? @ashu3103 

@Jubilee101 PTAL

Jubilee101:

Issue #758, opened 2025-09-19
Fetch metadata of a file from the postgres server


ashu3103:
**Is your feature request related to a problem? Please describe.**

While calculating `truncation_block_length` for an `INCREMENTAL` file, we need the knowledge of file's size at that moment.

**Describe the solution you'd like**

To get the size along with the metadata, we can provide an API in `server.h` which would be a wrapper around [pg_stat_file()](https://www.postgresql.org/docs/9.5/functions-admin.html#FUNCTIONS-ADMIN-GENFILE-TABLE)

We can execute this function if the user has `EXECUTE` privileges of this function or `SUPERUSER`

In general,
```
truncation_block_length = size_of_file / rel_seg_size
```

@jesperpedersen, @Jubilee101 PTAL

PR #760, opened 2025-09-21
[#758] Fetch metadata of a file from the postgres server

ashu3103:
feature: #758 

### DIFF
--- a/src/include/server.h
+++ b/src/include/server.h
@@ -36,6 +36,14 @@ extern "C" {
 #include <pgmoneta.h>
 
 #include <stdlib.h>
+#include <time.h>
+
+struct file_stats
+{
+   size_t size;             /**< The size of the file */
+   bool is_dir;             /**< Is the file a directory */
+   struct tm timetamps[4];  /**< array of timestamps in the order: access time, modification time, change time and creation time */
+};
 
 /**
  * Get the information for a server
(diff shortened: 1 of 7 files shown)


ashu3103:

src/include/server.h
```diff
@@ -36,6 +36,13 @@ extern "C" {
 #include <pgmoneta.h>
 
 #include <stdlib.h>
+#include <time.h>
+
+struct file_stats {
```
jesperpedersen: uncrustify

jesperpedersen (changes requested):

src/include/server.h
ashu3103: Done

src/libpgmoneta/server.c
```diff
@@ -525,6 +525,113 @@ pgmoneta_server_checkpoint(int srv, SSL* ssl, int socket, uint64_t* c_lsn)
...
+{
+   char* user = NULL;
+   char* cell_output = NULL;
+   int ret;
+   int q = 0;
+   bool has_privilege = false;
+   char query[MISC_LENGTH];
+   struct message* query_msg = NULL;
+   struct query_response* response = NULL;
+   struct file_stats stat;
```
Jubilee101, line 540: Are you trying to return a memory on stack? file_stats will be reclaimed after it's out of scope... Instead you should do malloc.

Jubilee101 (commented):
While you are at it, can you write a simple unit test for this?

Jubilee101:
Also looks like the query exectuion has a lot of duplicated code. Can you do something about it? Maybe extract the shared part to a static function or maybe even a util function in a query.h?

src/libpgmoneta/server.c
ashu3103: Yeah its correct that its allocated on stack but I am not returning it I am actually copying this stack allocated `stat` to a separate location in the caller. (by passing a pointer of `file_stats` to this function)

ashu3103:
> Also looks like the query exectuion has a lot of duplicated code. Can you do something about it? Maybe extract the shared part to a static function or maybe even a util function in a query.h?

Can I open a separate PR for this. I will take this after completing the incremental backup workflow, if that's okay?

Jubilee101:
> Can I open a separate PR for this. I will take this after completing the incremental backup workflow, if that's okay

It's ok for me, put a pin in this and let's look at it later. Focus on your priority.

src/libpgmoneta/server.c
Jubilee101: Hmm, my mistake. Yes this'll work.

ashu3103:
> While you are at it, can you write a simple unit test for this?

@Jubilee101 PTAL

Jubilee101:

Issue #763, opened 2025-09-21
Broken CI


Jubilee101:
Hi @Mohab96 , can you take a look at this https://github.com/pgmoneta/pgmoneta/actions/runs/17898148803/workflow?

Jubilee101:
@ashu3103 Looks like the tests are failing, can you take a look? CI was broken and disabled so we didn't notice.

Mohab96:
Removing this should fix it

<img width="356" height="360" alt="Image" src="https://github.com/user-attachments/assets/00111362-0e9c-44af-bf77-c96784ac789c" />

But I don't understand how #765 fixed it without doing this.

ashu3103:
@Mohab96, #765 doesn't fixed it.

The solution you are proposing is already pushed by @jesperpedersen, refer: https://github.com/pgmoneta/pgmoneta/commit/467ee8f9afe7e9e4ea5ef719d2f1cf946dda5eba

Mohab96:
Okay then, this issue should be closed

PR #765, opened 2025-09-22
[#763] Broken CI | Fix test_wal_summary test case

ashu3103:
@jesperpedersen @Jubilee101 This would resolve the test fails issue.
PTAL


ashu3103:
@Mohab96, can you a look at the macOs build in the CI?

Seems like the header libraries - `<nmmintrin.h>` and `<wmmintrin.h>` are not supported `arm` arch

Mohab96:
I don't understand how the CI is working, although you didn't make the changes I expected in `ci.yaml`?

> Seems like the header libraries - `<nmmintrin.h>` and `<wmmintrin.h>` are not supported `arm` arch

Do this and it should work
```c
#ifdef HAVE_CRC32_SSE42
#include <nmmintrin.h>
#include <wmmintrin.h>
#endif
```

ashu3103:
> I don't understand how the CI is working, although you didn't make the changes I expected in `ci.yaml`?
> 

No, I just fixed my testcase memory leaks.

> > Seems like the header libraries - `<nmmintrin.h>` and `<wmmintrin.h>` are not supported `arm` arch
> 
> Do this and it should work
> 
> ```c
> #ifdef HAVE_CRC32_SSE42
> #include <nmmintrin.h>
> #include <wmmintrin.h>
> #endif
> ```

Okay, thanks!

jesperpedersen:
We need the last part

jesperpedersen:
@Mohab96 

```
/Users/runner/work/pgmoneta/pgmoneta/src/walfilter.c:690:8: error: call to undeclared function 'chdir'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
[ 87%] Linking C executable pgmoneta-walinfo
  690 |    if (chdir(target_pg_wal_dir))
```

jesperpedersen:
@ashu3103 @Mohab96 Its `#include <unistd.h>`. Please, work together and finish this

ashu3103:
@jesperpedersen PTAL

jesperpedersen:
@ashu3103 @Mohab96 This is wrong - look at the `#ifdef`'s

ashu3103:
> @ashu3103 @Mohab96 This is wrong - look at the `#ifdef`'s

Hi all, I have one doubt:-

Read somewhere that `HAVE_CRC32_SSE42` is set by `Cmake`, so it can be defined either true or false. Now, is it right to use `ifdef HAVE_CRC32_SSE42`/`if defined(HAVE_CRC32_SSE42)` because that flag would always be defined. Rather, I think it should be only `#if HAVE_CRC32_SSE42`?

Thoughts? @Mohab96 @jesperpedersen

jesperpedersen:
@ashu3103 @Mohab96 If we have something optional then we need to use `cmake` to set that variable - so in this case it is `#ifdef HAVE_CRC32_SSE42 ... #endif`

jesperpedersen:
So, `#ifdef __x86_64__` looks wrong

Mohab96:
I don't see where the problem is; the CI is passing, and it compiles on my machine.

Issue #766, opened 2025-09-26
Refactor HTTP API


bassamadnan:
Clean up the code to make it more reusable, something like this is what im thinking:

```c
struct http_connection
{
   int socket;
   SSL* ssl;
   char* hostname;
   int port;
   bool secure;
};

struct http_response
{
   int status_code;
   char* headers;
   char* body;
   size_t body_size;
};

struct http_request
{
   int method;
   char* path;
   char* headers;
   void* data;
   size_t data_size;
};

```

This way connections will be reusable accross functions, then for s3 like cases we can have something like this:

```c
struct http_request* pgmoneta_http_request_create(int method, char* path);
int pgmoneta_http_request_add_header(struct http_request* req, char* name, char* value);
int pgmoneta_http_request_set_data(struct http_request* req, void* data, size_t size);
int pgmoneta_http_request_set_file(struct http_request* req, FILE* file, size_t size);
int pgmoneta_http_send(struct http_connection* conn, struct http_request* req, struct http_response** response);
```
and ofcourse functions for GET/POST/PUT/PUT FILE

jesperpedersen:
@bassamadnan Do a

```
struct http_common
{
   char* headers;
   void* data;
   size_t data_size;
}
```

Maybe even use `deque` for headers...

Just use `struct http` for `struct http_connection`.

And

```
struct http_request* pgmoneta_http_request_create(int method, char* path); // Always use output parameter
int pgmoneta_http_request_add_header(struct http_request* req, char* name, char* value); // Remove `_request_`, and add `_update_` and `_remove_`
int pgmoneta_http_request_set_data(struct http_request* req, void* data, size_t size); // Ok
int pgmoneta_http_request_set_file(struct http_request* req, FILE* file, size_t size); // Remove -- that is a layer above
int pgmoneta_http_send(struct http_connection* conn, struct http_request* req, struct http_response** response); // `_send` -> `_invoke`
```

jesperpedersen:
You can likely remove `_request_` if they are unique

jesperpedersen:
You put `struct http_common` as the first element in `_request` and `_response`

bassamadnan:
@jesperpedersen  The only problem is for put method, we are reading entire files, i think curl probably hands this better, although i dont know how much of a problem it is though it depends on the files we are going to send for backup.

jesperpedersen:
Yeah, but in the layer above we can convert `FILE*` into `void*`

[...]

PR #767, opened 2025-09-28
[#766] Refactor HTTP API

bassamadnan:


jesperpedersen:
@bassamadnan I'm thinking

bassamadnan:
If its of any help - I think looking at the test cases for HTTP will be the simplest way to check the reuasbility of the API's.

src/include/http.h
```diff
@@ -46,112 +46,125 @@ extern "C" {
...
- * Read response data directly from a socket
- * @param ssl SSL connection (can be NULL for non-secure connections)
- * @param socket The socket to read from
- * @param response_text Pointer to store the response text
- * @return MESSAGE_STATUS_OK on success, MESSAGE_STATUS_ERROR otherwise
+ * @return PGMONETA_HTTP_STATUS_OK upon success, otherwise PGMONETA_HTTP_STATUS_ERROR
  */
 int
-pgmoneta_http_read(SSL* ssl, int socket, char** response_text);
+pgmoneta_http_add_header(struct http_request* request, char* name, char* value);
```
jesperpedersen: I think we should add

* `pgmoneta_http_get_header`
* `pgmoneta_http_update_header`
* `pgmoneta_http_remove_header`

jesperpedersen (changes requested):

src/include/http.h
jesperpedersen: The rest of the .h looks good

bassamadnan: Hi, i did not add update/remove as I wasnt sure how we would be needing them

bassamadnan: Will add

jesperpedersen: Consider `struct deque` or `struct art` for `headers`

bassamadnan: used a `deque` to handle requests.

src/include/http.h
```diff
@@ -33,6 +33,7 @@
 extern "C" {
 #endif
 
+#include <deque.h>
```
jesperpedersen: Sorting

src/include/http.h
```diff
@@ -46,112 +47,155 @@ extern "C" {
...
+   void* data;        /**< Message data */
+   size_t data_size;  /**< Size of message data */
+};
+
+/** @struct http_request
+ * Defines a HTTP request
+ */
+struct http_request
+{
+   struct deque* headers;  /**< Request headers */
```
jesperpedersen: You need to use `http_payload`

bassamadnan: The problem is , for response we are not sure about the incoming format in which the responses come in,  creating a deque for response isn't trivial, we would have to parse it back by handling missing lines, split on `:` etc. Should I implement that too?

bassamadnan: @jesperpedersen

jesperpedersen: The deque is only for the headers - so yes, do a bit of parsing of the response

src/libpgmoneta/se_s3.c
```diff
@@ -406,149 +417,122 @@ s3_send_upload_request(char* local_root, char* s3_root, char* relative_path)
...
    {
-      free(date_region_service_key_hmac);
+      goto error;
    }
 
-   if (date_region_key_hmac != NULL)
+   if (response->status_code >= 200 && response->status_code < 300)
    {
-      free(date_region_key_hmac);
+      pgmoneta_log_info("Successfully uploaded file to S3: %s", s3_path);
```
jesperpedersen: Logging

src/libpgmoneta/se_s3.c
```diff
@@ -406,149 +417,122 @@ s3_send_upload_request(char* local_root, char* s3_root, char* relative_path)
...
+      pgmoneta_log_info("Object URL: https://%s/%s", s3_host, s3_path);
    }
-
-   if (date_key_hmac != NULL)
+   else
    {
-      free(date_region_key_hmac);
+      char* headers_str = response->payload.headers ? pgmoneta_deque_to_string(response->payload.headers, FORMAT_TEXT, NULL, 0) : NULL;
```
jesperpedersen: Same...

jesperpedersen (changes requested):

src/libpgmoneta/se_s3.c
bassamadnan: i should remove the logs? These were in sync with what curl used to print to `stdout`

jesperpedersen: Just make sure there is only one logging statement per upload/failed to upload/...

src/libpgmoneta/http.c
```diff
@@ -36,256 +36,50 @@
...
 /* system */
 #include <errno.h>
+#include <string.h>
 #include <unistd.h>
 #include <openssl/err.h>
 
-static int http_build_header(int method, char* path, char** request);
-static int http_extract_headers_body(char* response, struct http* http);
+static int http_read_response(SSL* ssl, int socket, char** response);
+int http_parse_response(char* response, struct http_response* http_response);
```
jesperpedersen: This should be static

src/include/http.h
```diff
@@ -46,112 +48,162 @@ extern "C" {
...
- * @param http The HTTP structure
+ * @param request The HTTP request
  * @param name The header name
  * @param value The header value
- * @return 0 upon success, otherwise 1
+ * @return PGMONETA_HTTP_STATUS_OK upon success, otherwise PGMONETA_HTTP_STATUS_ERROR
  */
 int
-pgmoneta_http_add_header(struct http* http, char* name, char* value);
+pgmoneta_http_add_header(struct http_request* request, char* name, char* value);
```
jesperpedersen: We could these `pgmoneta_http_request_add_header` and so on, and add a `pgmoneta_http_response_get_header`

src/libpgmoneta/http.c
bassamadnan: i've kept a test case for it, should i remove it?

jesperpedersen: We are testing the public API

jesperpedersen:
Rebased, and merged.

jesperpedersen:
@bassamadnan Please, port to the projects necessary

bassamadnan:
> @bassamadnan Please, port to the projects necessary

Yes, this is needed in pgexporter as well. Will do in due time.

Issue #797, opened 2025-11-03
Fetch the list of files under a dirctory from postgres server


ashu3103:
**Is your feature request related to a problem? Please describe.**

A tree of directories of postgres data directory is useful for many tasks like -
- Creating manifest file
- finding the unchanged files by comparing with the WAL summarizer (to write empty incremental files)

ashu3103:
Refer last comment: #798

PR #798, opened 2025-11-03
[#797] Fetch the list of files under a dirctory from postgres server

ashu3103:
Capture all the valid oids (both files and sub-directories) inside a server data directory

@jesperpedersen @Jubilee101, I am planning to create a local snapshot of data directory of the server with corresponding metadata inside the incr backup

PTAL


ashu3103:
This this actually not required since this can be achieved by `pgmoneta_ext_get_files()` in the extension and since we have no choice but to use extension in our incremental workflow for some things we don't be needing this and hence I'm closing this.

Apologies for misleading!

Issue #771, opened 2025-09-30
`NODE_LABEL` missing during WAL shipping to SSH storage engine


jchrys:
**Describe the bug**

WAL shipping crashes on the SSH storage engine when running a debug build, and the production build logs look off. `wal.c` only assigns `NODE_SERVER_ID` to the workflow nodes, but `se_ssh.c` always expects both NODE_SERVER_ID and `NODE_LABEL`.

https://github.com/pgmoneta/pgmoneta/blob/31645860ef9aa1502c4a4cddc4a3b8ce007cdf1a/src/libpgmoneta/wal.c#L166C1-L191C5

https://github.com/pgmoneta/pgmoneta/blob/31645860ef9aa1502c4a4cddc4a3b8ce007cdf1a/src/libpgmoneta/se_ssh.c#L147

**To Reproduce**
setup ssh storage config like below
https://github.com/pgmoneta/pgmoneta/blob/31645860ef9aa1502c4a4cddc4a3b8ce007cdf1a/doc/SSH.md

**Version**
main branch

**OS**
Linux z13 6.16.3 #1-NixOS SMP PREEMPT_DYNAMIC Sat Aug 23 14:49:42 UTC 2025 x86_64 GNU/Linux

jchrys:
Would it make sense to add a fixed `NODE_LABEL = 'wal_streaming'` in `wal.c`?

jesperpedersen:
@jchrys I would just start with looking at the requirements of `se_ssh.c`. See, that nodes are available, and which you require.

Assume that `se_ssh.c` need a lot of changes, so feel free to change things

Pranav-error:
This looks fixed on `main` already, so it may be closeable rather than needing a change.

The three functions on the WAL shipping path — `ssh_storage_setup`, `ssh_storage_wal_shipping_execute` and `ssh_storage_wal_shipping_teardown` — now assert only on `NODE_SERVER_ID` and treat a missing label as expected:

```c
assert(pgmoneta_art_contains_key(nodes, NODE_SERVER_ID));

server = (int)pgmoneta_art_search(nodes, NODE_SERVER_ID);
label = (char*)pgmoneta_art_search(nodes, NODE_LABEL);

pgmoneta_log_debug("SSH storage engine (WAL shipping/execute): %s/%s",
                   config->common.servers[server].name,
                   label != NULL ? label : "(wal)");
```

[...]

PR #813, opened 2025-11-17
[#771] add missing  NODE_LABEL for WAL shipping to ssh storage engine to

jchrys:
properly identify workflow type

### DIFF
--- a/src/libpgmoneta/wal.c
+++ b/src/libpgmoneta/wal.c
@@ -173,6 +173,11 @@ pgmoneta_wal(int srv, char** argv)
       goto error;
    }
 
+   if (pgmoneta_art_insert(nodes, NODE_LABEL, (uintptr_t)"wal", ValueString))
+   {
+      goto error;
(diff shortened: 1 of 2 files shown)


src/libpgmoneta/wal.c
```diff
@@ -173,6 +173,11 @@ pgmoneta_wal(int srv, char** argv)
       goto error;
    }
 
+   if (pgmoneta_art_insert(nodes, NODE_LABEL, (uintptr_t)"wal", ValueString))
```
jesperpedersen, line 176: This should use a constant, and it is isolated so difficult figure out.

Stick it in the main patch instead

jesperpedersen:
@jchrys If you need something for SSH that isn't the `NODE_LABEL` then add a new node definition.

So, we will just close this one

Jubilee101:
NODE_LABEL is what we use to identify backups. We probably should start documenting these parameters.

Issue #718, opened 2025-08-12
hash_elapsed_time


jesperpedersen:
Implement `hash_elapsed_time` for `backup.info` to track how much time is used on creating the `backup.sha512` file

sksingh2005:
@jesperpedersen sir is this issue open ? And can I try to look into this sir?

jesperpedersen:
@sksingh2005 Go for it ! Try and see how the other elapsed are implemented - it should be straight forward. Remember to read the developer documentation.

Thanks !

PR #816, opened 2025-12-09
[#718] Hash Elapsed Time for backup.info

sksingh2005:
## Problem
Implement `hash_elapsed_time` for `backup.info` to track how much time is used on creating the `backup.sha512` file

## Changes Made

- Extended the `backup` structure and `backup.info` persistence logic to include a new `HASH_ELAPSED` field.

- Instrumented the `SHA512` workflow execution to measure the duration of the hashing process using the monotonic clock, consistent with other workflow timers.

### Fixes- [#718 ]

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -34,3 +34,4 @@ Mingzhuo Yin <yinmingzhuo@gmail.com>
 Vanes Angelo <k124k3n@gmail.com>
 Bassam Adnan <mailbassam@gmail.com>
 Sara Nabih <nabihsara8@gmail.com>
+Shashank Singh <shashanksgh3@gmail.com>
--- a/doc/manual/en/97-acknowledgement.md
+++ b/doc/manual/en/97-acknowledgement.md
@@ -42,6 +42,7 @@ Mingzhuo Yin <yinmingzhuo@gmail.com>
 Vanes Angelo <k124k3n@gmail.com>
 Bassam Adnan <mailbassam@gmail.com>
 Sara Nabih <nabihsara8@gmail.com>
+Shashank Singh <shashanksgh3@gmail.com>
 ```
 
 ## Committers
--- a/src/include/info.h
+++ b/src/include/info.h
@@ -57,6 +57,7 @@ extern "C" {
 #define INFO_END_TIMELINE              "END_TIMELINE"
 #define INFO_END_WALPOS                "END_WALPOS"
 #define INFO_EXTRA                     "EXTRA"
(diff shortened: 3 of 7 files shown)


jesperpedersen:
@sksingh2005 Please, take a look at the CI

jesperpedersen:
@sksingh2005 Remember to squash your commits.

Our code looks like

```
  Variable defs
  Success
  if (failure)
  {
     goto error;
  }
  Success

  return 0;

error:

  Cleanup
  return 1;
```

sksingh2005:
@jesperpedersen sir squashed the commits as you instructed sir, from next time i will surely take care. Did some changes any change you would suggest sir ?

jesperpedersen:
@sksingh2005 Check if you need to do something with SHA-256 code paths

sksingh2005:
To check if the root directory is NULL and jumps to the error label if it is (Adding a NULL check for the root path in  wf_sha256.c
 to match the safety improvements in  wf_sha512.c.) ? Is i am thinking in right direction sir ?

jesperpedersen:
@sksingh2005 Yes. You can check if we can actually replace all SHA-256 usage with SHA-512 - that would make it a lot easier

sksingh2005:
Breaking Change for Archives/Incremental Backups: Switching  `se_ssh.c` to use `backup.sha512` means it will look for `backup.sha512` in remote backups for deduplication.
If older backups only have `backup.sha256`, deduplication against them will fail (falling back to full copy).
Since  `wf_backup` currently generates both, reasonably recent backups should already have `backup.sha512`, mitigating this risk.

The `backup_manifest` file checksum (the checksum of the manifest itself) will change from `SHA-256` to `SHA-512`. Any external tools relying on it being `SHA-256` will need updates. What should I do sir should we replace them ?

jesperpedersen:
@sksingh2005 That doesn't matter. The SSH module isn't fully functional yet. And, it was made before we had a `backup.sha512` - maybe we can remove sha256.c ? Making the code simpler would be good.

sksingh2005:
I checked se_s3.c and se_azure.c, and both require SHA-256 to generate authentication signatures for AWS and Azure APIs. Eliminating SHA-256 code completely would break cloud storage integration. Is i am write i analyzed it by going through them but have remove that from other. What to do sir ?

jesperpedersen:
@sksingh2005 Ok, just keep the SHA-256 functionality in place then.

jesperpedersen:
Merged. Welcome to the community !

Issue #607, opened 2025-04-26
CLI: conf get|set postgresql


jesperpedersen:
```
conf get postgresql <server> <key>
conf set postgresql <server> <key> <value> [reload]
```

In order to set a PostgreSQL (postgresql.conf) GUC. User must have access to 'postgres'

sommye-ctr:
@jesperpedersen If this is still open can I have a go at it?

jesperpedersen:
@sommye-ctr Go for it !

sommye-ctr:
@jesperpedersen How should I go about to handle the `set` command. The get command is fairly straightforward but for set, how would I send SQL queries over a physical replication connection? Using libpq?

jesperpedersen:
@sommye-ctr Just start with the `get` part. As you said `set` is a lot harder since it is over the standard protocol - https://www.postgresql.org/docs/18/protocol-message-formats.html - and functionality depends on the privilege of the connecting user. So, there is also a lot of documentation around for this.

It could be an idea to depend on pgmoneta-ext for some of this, so start by installing the extension and looking at its functionality

jesperpedersen:
Rejected

PR #817, opened 2025-12-11
[#607] Get set postgres GUC variables

sommye-ctr:
Relates #607

- [x] Implement `get-postgresql` command
- [x] Implement `set-postgresql` command on pgmoneta
- [x] Add `reload` option for the `set-postgresql` command
- [ ] https://github.com/pgmoneta/pgmoneta_ext/pull/33

## Semantics
An alternative command spelling (get-postgresql) is currently used to avoid ambiguity (possible confusion on user side). This can be aligned with existing `get <object>` semantics if preferred.

### Get-postgresql command (similar response to get)
`get-postgresql <server>` -> returns all the guc param values
`get-postgresql <server> <guc_param>` -> returns the value only of the guc_param

### Set-postgresql command (similar response to set)
`set-postgresql <server> <guc_param> <guc_value>` -> sets the value to guc parameter

To get over the limitation of not able to send set command over the message protocol, the [pgmoneta_ext](https://github.com/pgmoneta/pgmoneta_ext) will be used to get the response from server.

## Constraints

The `set` command will require the installation of the `pgmoneta_ext` and `EXECUTE` privilege on the function (to be created) `pgmoneta_ext_set_guc ` be granted to the pgmoneta role.

### Requirements from the extension function
- Be a `SECURITY DEFINER` function
- Raise errors when running on standby, unknown guc param, and return any error that postgres raises
- Return the response in such format
`
{
  "parameter": "param",
  "old_value": "old",
  "new_value": "new",
  "restart_required": "t"
}
`

### DIFF
--- a/src/cli.c
+++ b/src/cli.c
@@ -143,11 +143,14 @@ static int annotate(SSL* ssl, int socket, char* server, char* backup, char* comm
 static int mode(SSL* ssl, int socket, char* server, char* action, uint8_t compression, uint8_t encryption, int32_t output_format);
 static int conf_ls(SSL* ssl, int socket, uint8_t compression, uint8_t encryption, int32_t output_format);
 static int conf_get(SSL* ssl, int socket, char* config_key, uint8_t compression, uint8_t encryption, int32_t output_format);
+static int conf_get_postgresql(SSL* ssl, int socket, char* server, char* guc_param, uint8_t compression, uint8_t encryption, int32_t output_format);
 static int conf_set(SSL* ssl, int socket, char* config_key, char* config_value, uint8_t compression, uint8_t encryption, int32_t output_format);
+static int conf_set_postgresql(SSL* ssl, int socket, char* server, char* guc_param, char* guc_value, bool reload, uint8_t compression, uint8_t encryption, int32_t output_format);
 
 static int process_result(SSL* ssl, int socket, int32_t output_format);
 static int process_get_result(SSL* ssl, int socket, char* param, int32_t output_format);
 static int process_set_result(SSL* ssl, int socket, char* config_key, int32_t output_format);
+static int process_get_postgresql_result(SSL* ssl, int socket, int32_t output_format, char* guc_param);
 
 static int get_config_key_result(char* config_key, struct json* j, uintptr_t* r, int32_t output_format);
 
@@ -211,12 +214,12 @@ usage(void)
    printf("                           - 'prometheus' to reset the Prometheus statistics\n");
    printf("  compress                 Compress a file using configured method\n");
    printf("  conf <action>            Manage the configuration, with one of subcommands:\n");
-   printf("                           - 'get' to obtain information about a runtime configuration value\n");
-   printf("                             conf get <parameter_name>\n");
-   printf("                           - 'ls' to print the configurations used\n");
    printf("                           - 'reload' to reload the configuration\n");
-   printf("                           - 'set' to modify a configuration value;\n");
-   printf("                             conf set <parameter_name> <parameter_value>;\n");
+   printf("                           - 'ls' to print the configurations used\n");
+   printf("                           - 'get <parameter_name>' to obtain information about a runtime configuration value\n");
+   printf("                           - 'get postgresql <server> <guc_parameter>' to obtain information about a postgresql guc parameter\n");
+   printf("                           - 'set <parameter_name> <parameter_value>' to modify a configuration value\n");
+   printf("                           - 'set postgresql <server> <guc_parameter> <guc_value>' to modify a PostgreSQL GUC parameter\n");
    printf("  decompress               Decompress a file using configured method\n");
    printf("  decrypt                  Decrypt a file using master-key\n");
    printf("  delete                   Delete a backup from a server\n");
@@ -357,13 +360,13 @@ struct pgmoneta_command command_table[] = {
     .log_message = "<conf ls>"},
    {.command = "conf",
     .subcommand = "get",
-    .accepted_argument_count = {0, 1},
+    .accepted_argument_count = {0, 1, 2, 3},
     .action = MANAGEMENT_CONF_GET,
     .deprecated = false,
     .log_message = "<conf get> [%s]"},
    {.command = "conf",
     .subcommand = "set",
-    .accepted_argument_count = {2},
+    .accepted_argument_count = {2, 4, 5},
     .action = MANAGEMENT_CONF_SET,
     .deprecated = false,
     .log_message = "<conf set> [%s]"},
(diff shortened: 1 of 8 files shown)


sommye-ctr:
@jesperpedersen I read through the `pgmoneta-ext` docs and I found 2 ways we can go about the set command

1. Implement server side function in the extension to call the `ALTER SYSTEM..` using SPI
2. Use `pgmoneta_ext_get_file()` to fetch the current config file and then using `pgmoneta_ext_receive_file_chunk()` to return the updated postgres conf file

The 2 is obviously not very recommended since it directly edits the conf file

jesperpedersen:
@sommye-ctr Yes, this is mostly about getting an insight into the PostgreSQL configuration so for example `SHOW shared_buffers;`.

If no GUC is entered then it is basically a `SHOW ALL;`

jesperpedersen:
@sommye-ctr If it is very simple you can keep it in pgmoneta, and just create a message for it

sommye-ctr:
@jesperpedersen I have actually already done the implementation of `get` command and as you said added the methods within `pgmoneta` itself. I was looking for the implementation to `set` command

sommye-ctr:
> @jesperpedersen I read through the `pgmoneta-ext` docs and I found 2 ways we can go about the set command
> 
> 1. Implement server side function in the extension to call the `ALTER SYSTEM..` using SPI
> 2. Use `pgmoneta_ext_get_file()` to fetch the current config file and then using `pgmoneta_ext_receive_file_chunk()` to return the updated postgres conf file
> 
> The 2 is obviously not very recommended since it directly edits the conf file

@jesperpedersen For the set command, how should I go about then? and my current implementation of get command is it alright or should I make some changes?

jesperpedersen:
```
get postgresql primary  -- receives all GUCs
get postgresql primary shared_buffers -- receives the shared_buffers GUC
```
```
set postgresql primary shared_buffers 512MB -- set the shared_buffers GUC
```

I'll give you a quick code review

jesperpedersen:
@sommye-ctr You need documentation that describes how the functionality work. And, how the access user should be setup - what are the pros/cons for the `SUPERUSER` bit ? If the user isn't a superuser  what can be accessed ?

sommye-ctr:
@jesperpedersen I have documented my plan for this feature

sommye-ctr:
@jesperpedersen any thoughts on the document? particularly the `set` part?

jesperpedersen:
@sommye-ctr Yes, follow the "same" rukes as the existing `get` / `set`. 

`get` is all or one.
`set` is one only. 

Both are both a request and response.

sommye-ctr:
@jesperpedersen I have updated the `get` command flow to meet the desired requirements. It now returns all the GUC values if no param is given otherwise it returns 1 param. Could you review it?

sommye-ctr:
@jesperpedersen I’ve updated the code and adjusted the PR description to align with the current design and existing get/set semantics. Review requested.

src/cli.c
```diff
@@ -217,6 +220,10 @@ usage(void)
    printf("                           - 'reload' to reload the configuration\n");
    printf("                           - 'set' to modify a configuration value;\n");
    printf("                             conf set <parameter_name> <parameter_value>;\n");
+   printf("                           - 'get-postgresql' to obtain information about a postgresql guc parameter\n");
```
jesperpedersen: You should the existing the `get` and `set` function, and redirect to the `postgresql` function if the next argument is `postgresql` after `get`/`set`

src/cli.c
```diff
@@ -1028,10 +1051,25 @@ main(int argc, char** argv)
          exit_code = conf_get(s_ssl, socket, NULL, compression, encryption, output_format);
       }
    }
+   else if (parsed.cmd->action == MANAGEMENT_CONF_GET_POSTGRESQL)
```
jesperpedersen: You can keep the explicit management functions

src/cli.c
```diff
@@ -1028,10 +1051,25 @@ main(int argc, char** argv)
...
+      else
+      {
+         exit_code = conf_get_postgresql(s_ssl, socket, parsed.args[0], NULL, compression, encryption, output_format);
+      }
+   }  
    else if (parsed.cmd->action == MANAGEMENT_CONF_SET)
    {
       exit_code = conf_set(s_ssl, socket, parsed.args[0], parsed.args[1], compression, encryption, output_format);
    }
+   else if (parsed.cmd->action == MANAGEMENT_CONF_SET_POSTGRESQL)
```
jesperpedersen: Same

src/cli.c
```diff
@@ -1182,9 +1220,11 @@ help_conf(void)
 {
    printf("Manage the configuration\n");
    printf("  pgmoneta-cli conf [reload]\n");
-   printf("  pgmoneta-cli conf [ls]\n");
+   printf("  pgmoneta-cli conf [ls]\n"); 
    printf("  pgmoneta-cli conf [get] <parameter_name>\n");
```
jesperpedersen, line 1192: `pgmoneta-cli conf [get] [postgresql] <parameter_name>`

src/libpgmoneta/extension.c
```diff
@@ -94,6 +94,14 @@ pgmoneta_ext_promote(SSL* ssl, int socket, struct query_response** qr)
    return query_execute(ssl, socket, "SELECT pgmoneta_ext_promote();", qr);
 }
 
+int
+pgmoneta_ext_set_guc(SSL* ssl, int socket, char* guc_param, char* guc_val, struct query_response** qr)
+{
+   char query[MAX_QUERY_LENGTH];
+   snprintf(query, MAX_QUERY_LENGTH, "SELECT pgmoneta_ext_set_guc('%s', '%s');", guc_param, guc_val);
```
jesperpedersen: Use `pgmoneta_snprintf` instead

jesperpedersen:
@sommye-ctr You are on right track

sommye-ctr:
@jesperpedersen Made the requested changes. I plan to start working on the pgmoneta_ext implementation for the extension function.

sommye-ctr:
@jesperpedersen As I was working on the pgmoneta-ext function, I came across limitation that ALTER SYSTEM cannot be called using SPI from a function (I was not aware of this limitation earlier).
So this leaves only 1 option for the extension function to edit then conf file directly - bypassing these limitations.

jesperpedersen:
@sommye-ctr What is wrong with a simple `SET` ?

1) Connect to `postgres` as a superuser, or another role
2) Issue the `SET` statement
3) Optional reload / restart

The `SET` will take effect with future sessions.

I don't think editing the postgresql.conf is a path we should take.

sommye-ctr:
@jesperpedersen I was initially thinking in terms of server-wide parameters, but yeah this makes more sense. 
Opened a PR in pgmoneta_ext

jesperpedersen:
I think there are too many issues with going directly to PostgreSQL's GUCs. So, lets just close this

Issue #820, opened 2025-12-18
Log test case start points in testing suite


Jubilee101:
Log each test case start point to pin point the failing testing cases log faster

sksingh2005:
Is this issue open sir ? Can I look into this ?

jesperpedersen:
@sksingh2005 Go for it !

PR #824, opened 2025-12-19
[#820] Log test case start to identify failing tests faster

sksingh2005:
## Problem
When a test failed in CI or local runs, it was difficult and time-consuming to identify which specific test case was executing at the time of failure, especially for long-running or flaky tests (e.g. container-based tests).

The test output often only showed a failure or timeout without clearly indicating the last executed test case.

## Solution
Added a lightweight log at the start of every test case to clearly indicate which test is being executed.

Each test now prints:

`TEST START: <test_function_name>`

### Implementation Details

Logged the test start using:

`fprintf(stderr, "TEST START: %s\n", __func__);`

Applied consistently across all test cases

-  Uses stderr to ensure output is visible in CI and ctest logs

- No test logic or behavior was changed

- No additional dependencies introduced

### Fixes-[#820]

### DIFF
--- a/test/testcases/test_art.c
+++ b/test/testcases/test_art.c
@@ -49,6 +49,7 @@ static void test_obj_destroy_cb(uintptr_t obj);
 
 START_TEST(test_art_create)
 {
+   fprintf(stderr, "TEST START: %s\n", __func__);
    struct art* t = NULL;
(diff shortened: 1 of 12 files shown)


jesperpedersen:
Merged.

Issue #603, opened 2025-04-20
Backup restored size calculation that supports advanced restore flags


Jubilee101:
Backup restored size in backup.info should only reflect the size without the wal segments (it's correct when you use pgmoneta_backup_size, not so much when you just calculate the size of the data directory). And the WAL segment size should be added at runtime during restore. This is to support the restore flags such as `current` which involves copying over all the WAL segments we have during restore, instead of only the ones under pg_wal/. The restore size check should change accordingly.

sksingh2005:
Is this issue open sir ? Can i work on this issue ?

jesperpedersen:
@sksingh2005 Go for it !

PR #828, opened 2025-12-21
[#603] Exclude WAL size from incremental backup restore_size

sksingh2005:
## Problem
The `restore_size` in backup.info for incremental backups incorrectly included the size of WAL segments `(pg_wal)`. This caused inconsistency because the restore process (`restore.c`) dynamically calculates and adds the WAL size at runtime, leading to double-counting and inaccurate disk space checks.

## Solution
Updated `incr_backup_execute_14_to_16` in `src/libpgmoneta/wf_backup_incremental.c` to use `pgmoneta_directory_size_excludes`
with {"pg_wal", NULL} when calculating the backup size. This ensures `restore_size` reflects only the data directory size, allowing the restore logic to correctly add the dynamic WAL size when required.

### Fixes- #603

### DIFF
--- a/src/include/walfile/rmgr.h
+++ b/src/include/walfile/rmgr.h
@@ -57,33 +57,39 @@ extern "C" {
 
 #include <stdint.h>
 
-#define RM_XLOG_ID                                        0
-#define RM_XACT_ID                                        1
-#define RM_SMGR_ID                                        2
-#define RM_CLOG_ID                                        3
-#define RM_DBASE_ID                                       4
-#define RM_TBLSPC_ID                                      5
-#define RM_MULTIXACT_ID                                   6
-#define RM_RELMAP_ID                                      7
-#define RM_STANDBY_ID                                     8
-#define RM_HEAP2_ID                                       9
-#define RM_HEAP_ID                                        10
-#define RM_BTREE_ID                                       11
-#define RM_HASH_ID                                        12
-#define RM_GIN_ID                                         13
-#define RM_GIST_ID                                        14
-#define RM_SEQ_ID                                         15
-#define RM_SPGIST_ID                                      16
-#define RM_BRIN_ID                                        17
-#define RM_COMMIT_TS_ID                                   18
-#define RM_REPLORIGIN_ID                                  19
-#define RM_GENERIC_ID                                     20
-#define RM_LOGICALMSG_ID                                  21
+#define RM_XLOG_ID       0
+#define RM_XACT_ID       1
+#define RM_SMGR_ID       2
+#define RM_CLOG_ID       3
+#define RM_DBASE_ID      4
+#define RM_TBLSPC_ID     5
+#define RM_MULTIXACT_ID  6
+#define RM_RELMAP_ID     7
+#define RM_STANDBY_ID    8
+#define RM_HEAP2_ID      9
+#define RM_HEAP_ID       10
+#define RM_BTREE_ID      11
+#define RM_HASH_ID       12
+#define RM_GIN_ID        13
+#define RM_GIST_ID       14
+#define RM_SEQ_ID        15
+#define RM_SPGIST_ID     16
+#define RM_BRIN_ID       17
+#define RM_COMMIT_TS_ID  18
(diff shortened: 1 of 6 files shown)


jesperpedersen:
@sksingh2005 Can you rebase ?

src/include/walfile/rmgr.h
Jubilee101, line 1: Why include this format change?

src/libpgmoneta/restore.c
```diff
@@ -2195,6 +2210,16 @@ restore_backup_full(struct art* nodes)
    required_space =
       backup->restore_size + (pgmoneta_get_number_of_workers(server) * backup->biggest_file_size);
 
+   if (pgmoneta_art_contains_key(nodes, NODE_COPY_WAL))
+   {
+      if ((bool)pgmoneta_art_search(nodes, NODE_COPY_WAL))
+      {
+         char* wal_root = pgmoneta_get_server_wal(server);
+         required_space += pgmoneta_calculate_wal_size(wal_root, backup->wal);
+         free(wal_root);
```
Jubilee101: We prefer cleaning up in the end. So perhaps move the free to the end as well as the error section.

src/libpgmoneta/restore.c
```diff
@@ -2303,6 +2328,16 @@ restore_backup_incremental(struct art* nodes)
...
```
Jubilee101: same comment as above

src/libpgmoneta/restore.c
```diff
@@ -2195,6 +2210,16 @@ restore_backup_full(struct art* nodes)
...
+      }
+   }
```
Jubilee101, line 2226: Note that even without COPY_WAL we still include a few segments (it's in fact included within the backup from the beginning), so you need to account for that part.

src/libpgmoneta/wf_backup_incremental.c
```diff
@@ -960,6 +961,7 @@ incr_backup_execute_17_plus(char* name __attribute__((unused)), struct art* node
    backup->encryption = config->encryption;
    snprintf(backup->wal, sizeof(backup->wal), "%s", wal);
    backup->restore_size = size;
+   backup->backup_size = pgmoneta_directory_size_excludes(backup_data, excludes);
```
Jubilee101: This ain't right, backup size is the size of the backup after compression and encryption. What you should be actully fixing is the restore size.

Jubilee101: I think for incremental backup, the restore size natuarally excludes the WAL directory because we are calculating using the manifest instead of going through directories. So you probably don't have to do anything special here.

If that's the case (please verify), perhaps `pgmoneta_directory_size_excludes` should just be a static function for wf_backup for the time being. Please remove it from utils.c.

cc @ashu3103

Jubilee101 (changes requested):
Overall it's going in the right direction. Thanks for working on this.

src/include/walfile/rmgr.h
sksingh2005: sir i guess it was done from clang format

jesperpedersen: @sksingh2005 You have to use clang-format 21+ -- https://github.com/pgmoneta/pgmoneta/blob/main/CONTRIBUTING.md#code-style

sksingh2005: sir i  have Ubuntu clang-format version 21.1.8 which caused these changes

src/libpgmoneta/wf_backup_incremental.c
sksingh2005: should i just name it `directory_size_excludes`  sir??

jesperpedersen: @sksingh2005 Yes, `static` function should not have a `pgmoneta_` prefix

sksingh2005: sure sir

jesperpedersen:
Rebased, and merged.

PR #831, opened 2025-12-24
refactor: S3 configuration to be overwritten by the server config

Amr-Shams:
- Added per-server S3 configuration overrides entirely.
       - Added the now-redundant `s3_configuration` struct.
       - Revised documentation (CONFIGURATION.md, S3.md) to reflect the addition of per-server S3 options.


Amr-Shams:
@jesperpedersen i have tested with multiple servers each with their s3 config, (cloudflare, garage(locally), aws s3), and seems working ( i have handled some additional cases along the way. thank you for your time

Issue #851, opened 2026-01-08
CI: Tests for PostgreSQL 14/15 fails


ashu3103:
The issue is recurring after #847. As per the logs the error is emerging from WAL Summarization logic of the incremental workflow.

PR #788, opened 2025-10-18
[#787] Make `wal_file_name`  a public util API

ashu3103:
Changed the linkage of `wal_file_name` in wal.c to make it a public utility function

@Jubilee101 @jesperpedersen PTAL

Issue #839, opened 2026-01-02
WAL process locks repository, causing backup failure in test suite


Jubilee101:
Pinpointed the process that was holding the lock when backup failure happens:

```
2026-01-01 12:06:00 DEBUG main.c:2266 WAL: Acquired repository lock
2026-01-01 12:06:00 INFO  tsclient.c:246 /tmp/pgmoneta-test/base/conf/pgmoneta.conf /tmp/ 5
2026-01-01 12:06:00 DEBUG main.c:1042 Unencrypted request from localhost
2026-01-01 12:06:00 DEBUG main.c:1051 Management 1: {
  "Header": {
    "ClientVersion": "0.20.0",
    "Command": 1,
    "Compression": 0,
    "Encryption": 0,
    "Output": 1,
    "Timestamp": "20260101120600"
  },
  "Request": {
    "Backup": "newest",
    "Server": "primary"
  }
}
2026-01-01 12:06:00 INFO  backup.c:98 Backup: Server primary is active
2026-01-01 12:06:00 INFO  tsclient.c:211 outcome string {
  "Header": {
    "ClientVersion": "0.20.0",
    "Command": 1,
    "Compression": 0,
    "Encryption": 0,
    "Output": 1,
    "Timestamp": "20260101120600"
  },
  "Outcome": {
    "Error": 102,
    "Status": false,
    "Workflow": "backup"
  },
  "Request": {
    "Backup": "newest",
    "Server": "primary"
  },
  "Response": {
    "Server": "primary",
    "ServerVersion": "0.20.0"
  }
}
```

PR #852, opened 2026-01-08
[#851] CI: Tests for PostgreSQL 14/15 fails

ashu3103:
Fix: #851 

`pgmoneta_is_file` checks if a wal file is present in the directory or not but the issue is `pgmoneta_get_wal_files` captures both full and partial segments, and t a partial segment may convert to a full segment by the time we check the file existence.

Not to worry this has already been taken care of in the file copy logic, refer [this](https://github.com/pgmoneta/pgmoneta/pull/788/files#diff-28572a36916d0a56312f651385efdd568b28dbdf872f3a6ee940f1a9599c7818R1972)

### DIFF
--- a/src/libpgmoneta/walfile.c
+++ b/src/libpgmoneta/walfile.c
@@ -692,12 +692,6 @@ pgmoneta_summarize_walfiles(char* dir_path, uint64_t start_lsn, uint64_t end_lsn
          snprintf(file_path, MAX_PATH, "%s%s", dir_path, file);
       }
 
-      if (!pgmoneta_is_file(file_path))
-      {
-         pgmoneta_log_error("WAL file at %s does not exist", file_path);
-         goto error;
-      }
-
       if (pgmoneta_summarize_walfile(file_path, start_lsn, end_lsn, brt))
       {
          goto error;
--- a/src/libpgmoneta/wf_backup_incremental.c
+++ b/src/libpgmoneta/wf_backup_incremental.c
@@ -1747,12 +1747,6 @@ copy_wal_from_archive(char* start_wal_file, char* wal_dir, char* backup_data)
          src_file = pgmoneta_append(src_file, wal_dir);
          src_file = pgmoneta_append(src_file, file_name);
 
-         if (!pgmoneta_is_file(src_file))
-         {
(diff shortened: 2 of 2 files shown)


ashu3103:
@jesperpedersen @Jubilee101, just ran into another problem. After executing `pgmoneta_get_wal_files` I got something like:

```
2026-01-08 19:34:24 INFO  walfile.c:684 Found WAL file: 00000001000000000000001C.zstd
2026-01-08 19:34:24 INFO  walfile.c:684 Found WAL file: 00000001000000000000001D.zstd
2026-01-08 19:34:24 INFO  walfile.c:684 Found WAL file: 00000001000000000000001E
2026-01-08 19:34:24 INFO  walfile.c:684 Found WAL file: 00000001000000000000001E.zstd
```

Here as you can see, there are two occurrences of `00000001000000000000001E` which later causes one of them to fail the `pgmoneta_is_file` check. I'm not able to determine the cause of this and why it didn't happened in previous implementation.

Any suggestions?

jesperpedersen:
@ashu3103 Yes, it has to do with `pgmoneta_get_wal_files` need a lock... it is related to #839... I'll ping you when the next patch is in

ashu3103:
> @ashu3103 Yes, it has to do with `pgmoneta_get_wal_files` need a lock... it is related to #839... I'll ping you when the next patch is in

Sure

jesperpedersen:
@ashu3103 Take a look at `pgmoneta_summarize_walfiles` -- we need a lock on the server to get the WAL file names -- I don't like the function signature atm... Feel free to submit against this issue or #839

jesperpedersen:
I mean the `pgmoneta_summarize_walfiles` signature

ashu3103:
[this](https://github.com/pgmoneta/pgmoneta/blob/e756daea3ad693b44a4a0fd490a1c3a659a1eebc/src/libpgmoneta/walfile/wal_summary.c#L265) is failing for me, @jesperpedersen help wanted

jesperpedersen:
@ashu3103 The new logging shows that the .partial is renamed before `get_wal_files`, but it doesn't show up in the FILETRACKER for copy. So, I think we need more logging to figure out where the wheel falls off...

Issue #856, opened 2026-01-09
Buffer overwrites in test utils


ashu3103:
The issue is all the buffers are allocated the same size `128` which may cause buffer overwrite while doing things such as:

`sprintf(check_file, "%s/000000010000000000000001", to_dir)`

the compiler is concerned about the cases when `to_dir` is 128 which cause `check_file` buffer overwrite and therefore throws a warning.

Logs while compiling

```
/home/ashu3103/Desktop/pgmoneta/test/testcases/test_utils.c: In function ‘test_utils_files_advanced_fn’:
/home/ashu3103/Desktop/pgmoneta/test/testcases/test_utils.c:1011:20: warning: ‘/sub’ directive writing 4 bytes into a region of size between 1 and 128 [-Wformat-overflow=]
 1011 |    sprintf(sub, "%s/sub", src);
      |                    ^~~~
/home/ashu3103/Desktop/pgmoneta/test/testcases/test_utils.c:1011:4: note: ‘sprintf’ output between 5 and 132 bytes into a destination of size 128
 1011 |    sprintf(sub, "%s/sub", src);
      |    ^~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/ashu3103/Desktop/pgmoneta/test/testcases/test_utils.c:1012:24: warning: ‘/file.txt’ directive writing 9 bytes into a region of size between 1 and 128 [-Wformat-overflow=]
 1012 |    sprintf(subfile, "%s/file.txt", sub);
      |                        ^~~~~~~~~
/home/ashu3103/Desktop/pgmoneta/test/testcases/test_utils.c:1012:4: note: ‘sprintf’ output between 10 and 137 bytes into a destination of size 128
 1012 |    sprintf(subfile, "%s/file.txt", sub);
      |    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/ashu3103/Desktop/pgmoneta/test/testcases/test_utils.c:1035:25: warning: ‘/plain.txt’ directive writing 10 bytes into a region of size between 1 and 128 [-Wformat-overflow=]
 1035 |    sprintf(file_src, "%s/plain.txt", src);
      |                         ^~~~~~~~~~
/home/ashu3103/Desktop/pgmoneta/test/testcases/test_utils.c:1035:4: note: ‘sprintf’ output between 11 and 138 bytes into a destination of size 128
 1035 |    sprintf(file_src, "%s/plain.txt", src);
[...]

PR #857, opened 2026-01-09
[#856] Fix buffer overflow warnings in test_utils.c using pgmoneta_append()

sksingh2005:
## Problem
Fixed-size 128-byte buffers used with `sprintf()` for path concatenation could overflow when appending long filenames like `WAL` segment names (25 characters), causing 8 compiler warnings:

## Solution
1. Replaced Fixed Buffers with Dynamic Allocation
In both `test_utils_files_advanced` and `test_utils_missing_wal`, I replaced static arrays with pointers initialized to NULL.

2. Used `pgmoneta_append()` for Concatenation Instead of `sprintf()`, which is unsafe for fixed-size buffers, I used `pgmoneta_append()`. This function automatically grows the buffer to fit the new content.

3. Added Memory Cleanup Added `free()` calls to prevent memory leaks in the test suite.

### DIFF
--- a/test/testcases/test_utils.c
+++ b/test/testcases/test_utils.c
@@ -1001,15 +1001,19 @@ END_TEST
 
 START_TEST(test_utils_files_advanced)
 {
-   char src[128];
-   char dst[128];
-   char sub[128];
-   char subfile[128];
+   char* src = NULL;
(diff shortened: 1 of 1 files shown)


jesperpedersen:
Merged.

Issue #860, opened 2026-01-11
Verify command crashes with null pointer when files parameter is omitted


sksingh2005:
**Describe the bug**

When files parameter is not provided (NULL), `pgmoneta_art_search` returns NULL. Passing NULL to `strcasecmp()` is undefined behavior → crash.

**Error Message**
```
wf_verify.c:124:20: runtime error: null pointer passed as argument 1, which is declared to never be null
/usr/include/strings.h:117:33: note: nonnull attribute specified here
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior wf_verify.c:124:20
verify.c:198:20: runtime error: null pointer passed as argument 1, which is declared to never be null
```

jesperpedersen:
@sksingh2005 Or are you taking it ?

sksingh2005:
Sir i have fixed it with minor changes in `verify.c ` and `wf_verify.c`

jesperpedersen:
@sksingh2005 Please, submit a separate PR for it

PR #861, opened 2026-01-11
[#860] fix:Prevent null pointer crash in verify command when files parameter is omitted

sksingh2005:
## Fixes -  [#860]

Add null checks before calling strcasecmp() in verify workflow:
- `wf_verify.c` - Check `USER_FILES` art node before comparison
- `verify.c` - Check files parameter before comparison

### DIFF
--- a/src/libpgmoneta/verify.c
+++ b/src/libpgmoneta/verify.c
@@ -100,6 +100,11 @@ pgmoneta_verify(SSL* ssl, int client_fd, int server, uint8_t compression, uint8_
       goto error;
    }
 
+   if (files == NULL || strlen(files) == 0)
+   {
+      files = MANAGEMENT_ARGUMENT_FAILED;
+   }
+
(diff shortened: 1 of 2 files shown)


jesperpedersen:
@sksingh2005 Yes, that is correct,. But, it isn't the root cause - `USER_FILES` must be defined a valid value

sksingh2005:
`if (files == NULL)
{
   files = "";
}`
From this `USER_FILES` will always have a valid value (empty string "" if not provided), not NULL.
Is it right sir ?

jesperpedersen:
@sksingh2005 I would check in verify.c and set to `MANAGEMENT_ARGUMENT_FAILED` if it is `NULL` or `""`

jesperpedersen:
@sksingh2005 Did you rebase against `main` ?

sksingh2005:
yes sir

jesperpedersen:
Rebased, and merged.

Issue #868, opened 2026-01-19
pgmoneta_get_wal_files: Filter files


jesperpedersen:
Make sure that `pgmoneta_get_wal_files` only returns files that follow valid WAL names - including compression and encryption extensions

PR #870, opened 2026-01-19
[#868] filter files

saraanbih:
Filter and return only valid WAL files (including compressed, encrypted, and partial files)

Fixes [#868]

### DIFF
--- a/src/libpgmoneta/utils.c
+++ b/src/libpgmoneta/utils.c
@@ -1997,6 +1997,83 @@ error:
    return 1;
 }
 
+/**
+ * Check if a file has a valid WAL file suffix pattern
+ * 
+ * Valid patterns are:
+ * - 24-char hex (uncompressed)
+ * - 24-char hex + .partial
+ * - 24-char hex + compression suffix (.gz, .lz4, .zst, .zstd, .bz2)
+ * - 24-char hex + compression suffix + .aes
+ * 
+ * @param f The filename to check
+ * @return true if the file has a valid WAL suffix pattern, false otherwise
+ */
+static bool
+is_valid_wal_file_suffix(char* f)
+{
+   char* valid_suffixes[] = {
+      "",
+      ".partial",
+      ".gz",
+      ".lz4",
+      ".zstd",
+      ".bz2",
+      ".gz.aes",
+      ".lz4.aes",
+      ".zstd.aes",
+      ".bz2.aes",
+      NULL};
+
+   char path[MAX_PATH];
+
+   memset(path, 0, sizeof(path));
+   size_t len = strlen(f);
+   if (len >= sizeof(path))
+   {
+      len = sizeof(path) - 1;
+   }
+   memcpy(path, f, len);
+
+   char* name = basename(path);
+
+   int name_len = strlen(name);
+
+   for (int i = 0; valid_suffixes[i] != NULL; i++)
+   {
+      int suffix_len = strlen(valid_suffixes[i]);
+
+      if (suffix_len == 0)
+      {
+         if (name_len == 24 && pgmoneta_is_wal_file(name))
+         {
+            return true;
+         }
+      }
+      else
+      {
+         if (pgmoneta_ends_with(name, valid_suffixes[i]))
+         {
+            int prefix_len = name_len - suffix_len;
+            if (prefix_len == 24)
+            {
+               char prefix[25];
+
+               memset(prefix, 0, sizeof(prefix));
+               memcpy(prefix, name, 24);
+
+               if (pgmoneta_is_wal_file(prefix))
+               {
+                  return true;
+               }
+            }
+         }
+      }
+   }
+
+   return false;
+}
+
 int
 pgmoneta_get_wal_files(char* base, struct deque** files)
 {
@@ -2020,16 +2097,21 @@ pgmoneta_get_wal_files(char* base, struct deque** files)
 
    while ((entry = readdir(dir)) != NULL)
    {
+      if (entry->d_type != DT_REG)
+      {
+         continue;
+      }
+
       if (pgmoneta_ends_with(entry->d_name, ".history"))
       {
          continue;
       }
 
-      if (entry->d_type == DT_REG)
+      if (is_valid_wal_file_suffix(entry->d_name))
       {
          if (pgmoneta_deque_add(array, entry->d_name, (uintptr_t)entry->d_name, ValueString))
          {
-            pgmoneta_log_error("WAL: Error %s", entry->d_name);
+            pgmoneta_log_error("WAL: Error adding %s", entry->d_name);
             goto error;
          }
       }
@@ -3291,14 +3373,18 @@ error:
 bool
 pgmoneta_is_wal_file(char* file)
 {
-   if (pgmoneta_ends_with(file, ".history"))
+   if (strlen(file) != 24)
    {
       return false;
    }
 
-   if (strlen(file) != 24)
+   for (int i = 0; i < 24; i++)
    {
-      return false;
+      char c = file[i];
+      if (!isxdigit(c))
+      {
+         return false;
+      }
    }
 
    return true;


jesperpedersen:
@saraanbih CI...

jesperpedersen:
@saraanbih No, first you need to create a suffix to compare against. Then a prefix. Then check. Use `starts_with` / `ends_with`

jesperpedersen:
@saraanbih See get_wal_file_name in wal_summary.c - it need to be promoted to `utils.h` and then start from there

src/libpgmoneta/utils.c
```diff
@@ -2020,36 +2033,87 @@ pgmoneta_get_wal_files(char* base, struct deque** files)
...
+         continue;
+      }
+
       if (pgmoneta_ends_with(entry->d_name, ".history"))
       {
          continue;
       }
 
-      if (entry->d_type == DT_REG)
+      if (pgmoneta_ends_with(entry->d_name, ".pdf"))
```
jesperpedersen: What is this ?

saraanbih: @jesperpedersen  to skip pdf files, when i use it at interactive it was show pdf files

jesperpedersen:
@saraanbih Remember, that the prefix is the important part - a WAL file could have been decrypted and decompressed

src/libpgmoneta/utils.c
jesperpedersen: @saraanbih Everything that isn't valid is filtered out -- we are not looking at all extensions like .pdf

saraanbih:
> @saraanbih Remember, that the prefix is the important part - a WAL file could have been decrypted and decompressed

the prefix or the suffix?

src/libpgmoneta/utils.c
jesperpedersen: @saraanbih Its

```
00000001000000000000000F
00000001000000000000000F.zstd
00000001000000000000000F.zstd.aes
00000001000000000000000F.partial
```

for example

saraanbih: @jesperpedersen OK, this means that the part of check pdf isn't important, right?

jesperpedersen: @saraanbih These 4 are valid for the filter

jesperpedersen:
@saraanbih You may have to do something about

```
00000001000000000000000F
00000001000000000000000F.gz
00000001000000000000000F.gz.aes
00000001000000000000000F.partial
```

as they could be from an older backup (before the switch to ZSTD)

jesperpedersen:
@saraanbih Start with a 

```
static bool
is_valid_wal_file_suffix(char* f)
```

saraanbih:
@jesperpedersen I don't know what is the problem at CI :(

jesperpedersen:
@saraanbih Ignore that one - it is being worked on

src/libpgmoneta/utils.c
```diff
@@ -1997,6 +1997,82 @@ pgmoneta_get_files(char* base, struct deque** files)
...
+ */
+static bool
+is_valid_wal_file_suffix(char* f)
+{
+   char* valid_suffixes[] = {
+      "",
+      ".partial",
+      ".gz",
+      ".lz4",
+      ".zst",
```
jesperpedersen: We don't use that

src/libpgmoneta/utils.c
```diff
@@ -1997,6 +1997,82 @@ pgmoneta_get_files(char* base, struct deque** files)
...
+      ".zstd",
+      ".bz2",
+      ".gz.aes",
+      ".lz4.aes",
+      ".zst.aes",
```
jesperpedersen: Same

src/libpgmoneta/utils.c
```diff
@@ -1997,6 +1997,82 @@ pgmoneta_get_files(char* base, struct deque** files)
...
+      ".zstd.aes",
+      ".bz2.aes",
+      ".partial.gz",
```
jesperpedersen: .partial are not compressed and encrypted

src/include/utils.h
```diff
@@ -1441,6 +1441,16 @@ pgmoneta_is_wal_file(char* file);
...
 pgmoneta_wal_file_name(uint32_t tli, size_t segno, int segsize);
 
+/**
+ * Get the WAL file name with appropriate compression and encryption suffixes
+ * based on the configuration
+ *
+ * @param file The input file name (may include .partial extension)
+ * @return The derived WAL file name with appropriate suffixes
+ */
+char*
```
jesperpedersen: You don't need this function in this PR

src/libpgmoneta/utils.c
```diff
@@ -2020,36 +2096,37 @@ pgmoneta_get_wal_files(char* base, struct deque** files)
...
          {
-            pgmoneta_log_error("WAL: Error %s", entry->d_name);
+            pgmoneta_log_error("WAL: Error adding %s", entry->d_name);
             goto error;
          }
       }
    }
 
    pgmoneta_deque_sort(array);
-
```
jesperpedersen, line 2039: Leave the linefeeds

jesperpedersen:
@saraanbih You should change `pgmoneta_is_wal_file` such that the 24 characters are checked against `0-F`

jesperpedersen:
@saraanbih in that function I don't like the `.history` check - we should move that up

src/libpgmoneta/utils.c
```diff
@@ -1997,6 +1997,82 @@ pgmoneta_get_files(char* base, struct deque** files)
...
+         if (name_len == 24 && pgmoneta_is_wal_file(f))
+         {
+            return true;
+         }
+      }
+      else
+      {
+         if (pgmoneta_ends_with(f, valid_suffixes[i]))
+         {
+            int prefix_len = name_len - suffix_len;
```
jesperpedersen, line 2057: `man 3 basename `

src/include/utils.h
saraanbih: > @saraanbih See get_wal_file_name in wal_summary.c - it need to be promoted to utils.h and then start from there

@jesperpedersen should i return it to wal_summary or what?

jesperpedersen: @saraanbih Yes, all pull requests should be as minimal as possible. Committers have to review them

src/libpgmoneta/utils.c
```diff
@@ -1997,6 +1997,76 @@ pgmoneta_get_files(char* base, struct deque** files)
...
+      ".lz4",
+      ".zstd",
+      ".bz2",
+      ".gz.aes",
+      ".lz4.aes",
+      ".zstd.aes",
+      ".bz2.aes",
+      NULL};
+
+   char path[PATH_MAX];
```
jesperpedersen: MAX_PATH

src/libpgmoneta/utils.c
```diff
@@ -1997,6 +1997,76 @@ pgmoneta_get_files(char* base, struct deque** files)
...
+   strncpy(path, f, sizeof(path));
+   path[sizeof(path) - 1] = '\0';
+
```
jesperpedersen, line 2037: Use memset and memcpy instead

src/libpgmoneta/utils.c
jesperpedersen: ?

src/libpgmoneta/utils.c
```diff
@@ -1997,6 +1997,76 @@ pgmoneta_get_files(char* base, struct deque** files)
...
+      }
+      else
+      {
+         if (pgmoneta_ends_with(name, valid_suffixes[i]))
+         {
+            int prefix_len = name_len - suffix_len;
+            if (prefix_len == 24)
+            {
+               char prefix[25];
+               strncpy(prefix, name, 24);
```
jesperpedersen: memset / memcpy

jesperpedersen:
Merged.

Issue #875, opened 2026-01-24
wal_summary test


jesperpedersen:
The test case is currently being skipped

jesperpedersen:
@sksingh2005 @shashidhar-bm @ashu3103 Can one of you take the lead on this one, and figure out how it should be solved ?

shashidhar-bm:
Yes, I'll look into it.

jesperpedersen:
@shashidhar-bm Thanks, coordinate with the others

sksingh2005:
Sure sir

shashidhar-bm:
```
/home/shashidhar/project/pgmoneta/test/testcases/test_wal_summary.c:138:F:test_wal_summary:test_pgmoneta_wal_summary:0: failed to summarize the wal
```
@jesperpedersen ,I just rolled back a few commits to see how it used to behave with libcheck, and it's failing there too. I wonder, have we always had this issue?

jesperpedersen:
@ashu3103 Can you help with this ?

ashu3103:
> [@ashu3103](https://github.com/ashu3103) Can you help with this ?

Sure, looking into it!

@shashidhar-bm can you provide the logs if possible

jesperpedersen:
@ashu3103 @shashidhar-bm did a big update today, so everything should be in `/tmp/pgmoneta-test/log`

sksingh2005:
I guess the real issue is `summarize_walfiles()` in `wal_summary.c` always converted filenames to their compressed form (e.g .zstd) before checking if they exist.
When a `WAL` file was still active (.partial), the compressed version didn't exist yet, causing the function to fail after exhausting retries.
@jesperpedersen @ashu3103 sir ?

sksingh2005:
I have made a minimal fix for this which is now causing the test to get passed instead of skipping.

jesperpedersen:
@sksingh2005 Once a WAL segment is complete it is scheduled to be compressed and encrypted

jesperpedersen:
@shashidhar-bm We need to think about `SKIP` - it is a flag to show there are issues with testing in that area. Otherwise the test cases should fail. Upon release there shouldn't be any `SKIP` ones

sksingh2005:
[...]

PR #850, opened 2026-01-08
[#825] Test: Used management protocol for functional testing

sommye-ctr:
Fixes #825 

### Changes

- Added `pgmoneta_tsclient_list_backup` in `tsclient` to send LIST_BACKUP to management protocol
- Added `tsclient_helpers` for helper functions related to the tests
- Replaced any methods using internal apis with the above ones

#### Notes

- In relation with #839, the creation of incremental backups is resulting in an error

PR #895, opened 2026-01-30
[#875] Fix test skips: WAL summary fallback + migrate to list-backup API

sksingh2005:
Fixes - [#875]

- Replace `info.h` usage with `tsclient_list_backup` helpers

### DIFF
--- a/src/libpgmoneta/wal.c
+++ b/src/libpgmoneta/wal.c
@@ -1740,6 +1740,9 @@ pgmoneta_wal_server_compress_encrypt(int srv, char** argv, char* wal_file)
       int retry_count = 0;
       char* d = NULL;
 
+      /* Boost priority for compression to complete faster */
+      pgmoneta_set_priority(PRIORITY_HIGH);
+
       if (argv != NULL)
       {
          pgmoneta_set_proc_title(1, argv, "wal/ce", config->common.servers[srv].name);
--- a/src/libpgmoneta/walfile/wal_summary.c
+++ b/src/libpgmoneta/walfile/wal_summary.c
@@ -33,6 +33,7 @@
 #include <utils.h>
 #include <wal.h>
 #include <walfile.h>
+#include <walfile/wal_reader.h>
 #include <walfile/wal_summary.h>
 
 #include <dirent.h>
@@ -41,7 +42,7 @@
 static char* summary_file_name(uint64_t s_lsn, uint64_t e_lsn);
 static int summarize_walfile(char* path, uint64_t start_lsn, uint64_t end_lsn, block_ref_table* brt);
 static int summarize_walfiles(int srv, char* dir_path, uint64_t start_lsn, uint64_t end_lsn, block_ref_table* brt);
-static char* get_wal_file_name(char* file);
+static char* get_wal_file_name(char* dir_path, char* file);
 
 /**
  * Handles summarization of arbitary range in a timeline only
@@ -301,8 +302,30 @@ retry1:
    {
       char* file = (char*)file_iterator->value->data;
       char* fn = NULL;
+      xlog_seg_no seg_no = 0;
+      uint64_t seg_start_lsn = 0;
+      int wal_size = config->common.servers[srv].wal_size;
 
-      fn = get_wal_file_name(file);
+      /* Check if file is beyond end_lsn - skip if so */
+      if (pgmoneta_validate_wal_filename(file, NULL, &seg_no, wal_size) == 0)
+      {
+         XLOG_SEG_NO_OFFEST_TO_REC_PTR(seg_no, 0, wal_size, seg_start_lsn);
+         if (seg_start_lsn > end_lsn)
+         {
+            pgmoneta_log_debug("WAL summary: skipping %s (segment starts at %" PRIX64 ", beyond end_lsn %" PRIX64 ")",
+                               file, seg_start_lsn, end_lsn);
+            continue;
+         }
+      }
+
+      fn = get_wal_file_name(dir_path, file);
+
+      /* File needed but not available - error out */
+      if (fn == NULL)
+      {
+         pgmoneta_log_error("WAL summary: %s not available after waiting", file);
+         goto error;
+      }
 
       memset(file_path, 0, MAX_PATH);
       if (!pgmoneta_ends_with(dir_path, "/"))
@@ -316,39 +339,6 @@ retry1:
 
       free(fn);
       fn = NULL;
-
-      retry_count = 0;
-      long sleep_ns = 500000000L;       /* Start with 500ms */
-      long max_sleep_ns = 30000000000L; /* Cap at 30 seconds */
-
-      /* Wait for file with exponential backoff (compression/encryption may be in progress) */
-      while (!pgmoneta_exists(file_path) && retry_count < 10)
-      {
-         pgmoneta_log_debug("WAL file at %s does not exist - retrying (%d/10, wait %ldms)",
-                            file_path, retry_count + 1, sleep_ns / 1000000);
-         retry_count++;
-         active = false;
-         SLEEP(sleep_ns);
-
-         /* Exponential backoff: double the sleep time, up to max */
-         if (sleep_ns < max_sleep_ns)
-         {
-            sleep_ns *= 2;
-            if (sleep_ns > max_sleep_ns)
-            {
-               sleep_ns = max_sleep_ns;
-            }
-         }
-      }
-
-      if (!pgmoneta_exists(file_path))
-      {
-         char* fs = pgmoneta_deque_to_string(files, FORMAT_TEXT, NULL, 0);
-         pgmoneta_log_debug("%s", fs);
-         pgmoneta_log_error("WAL file at %s does not exist after retries", file_path);
-         free(fs);
-         goto error;
-      }
 
       pgmoneta_log_debug("WAL file at %s", file_path);
 
@@ -377,32 +367,49 @@ error:
 }
 
 static char*
-get_wal_file_name(char* file)
+get_wal_file_name(char* dir_path, char* file)
 {
    char* fn = NULL;
+   char* base = NULL;
    char* suffix = NULL;
+   char compressed_path[MAX_PATH];
+   char uncompressed_path[MAX_PATH];
+   char partial_path[MAX_PATH];
+   int retry_count = 0;
+   long sleep_ms = 500;       /* 500ms */
+   long max_sleep_ms = 30000; /* 30s cap */
+   int max_retries = 10;
+   int orig_priority = 0;
+   char* sep = NULL;
    struct main_configuration* config;
 
    config = (struct main_configuration*)shmem;
 
+   sep = pgmoneta_ends_with(dir_path, "/") ? "" : "/";
+
+   /* Strip .partial if present to get base name */
    if (pgmoneta_ends_with(file, ".partial"))
    {
-      fn = pgmoneta_remove_suffix(file, ".partial");
+      base = pgmoneta_remove_suffix(file, ".partial");
    }
    else
    {
-      fn = pgmoneta_append(fn, file);
+      base = pgmoneta_append(NULL, file);
    }
 
-   if (config->compression_type == COMPRESSION_CLIENT_GZIP || config->compression_type == COMPRESSION_SERVER_GZIP)
+   /* Build expected compression/encryption suffix */
+   if (config->compression_type == COMPRESSION_CLIENT_GZIP ||
+       config->compression_type == COMPRESSION_SERVER_GZIP)
    {
       suffix = pgmoneta_append(suffix, ".gz");
    }
-   else if (config->compression_type == COMPRESSION_CLIENT_ZSTD || config->compression_type == COMPRESSION_SERVER_ZSTD)
+   else if (config->compression_type == COMPRESSION_CLIENT_ZSTD ||
+            config->compression_type == COMPRESSION_SERVER_ZSTD)
    {
       suffix = pgmoneta_append(suffix, ".zstd");
    }
-   else if (config->compression_type == COMPRESSION_CLIENT_LZ4 || config->compression_type == COMPRESSION_SERVER_LZ4)
+   else if (config->compression_type == COMPRESSION_CLIENT_LZ4 ||
+            config->compression_type == COMPRESSION_SERVER_LZ4)
    {
       suffix = pgmoneta_append(suffix, ".lz4");
    }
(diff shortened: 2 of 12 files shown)


src/libpgmoneta/backup.c
```diff
@@ -82,7 +82,7 @@ pgmoneta_backup(int client_fd, int server, uint8_t compression, uint8_t encrypti
    if (!config->common.servers[server].valid)
    {
       ec = MANAGEMENT_ERROR_BACKUP_INVALID;
-      pgmoneta_log_error("Backup: Server %s is not in a valid configuration", config->common.servers[server].name);
+      pgmoneta_log_error("Backup: Server %s is not in a valid configuration (valid=false)", config->common.servers[server].name);
```
jesperpedersen: Remove

src/libpgmoneta/walfile/wal_summary.c
```diff
@@ -341,13 +341,31 @@ summarize_walfiles(int srv, char* dir_path, uint64_t start_lsn, uint64_t end_lsn
...
 
+      /* If compressed/encrypted file doesn't exist, check for uncompressed version as fallback */
       if (!pgmoneta_exists(file_path))
       {
-         char* fs = pgmoneta_deque_to_string(files, FORMAT_TEXT, NULL, 0);
-         pgmoneta_log_debug("%s", fs);
-         pgmoneta_log_error("WAL file at %s does not exist after retries", file_path);
-         free(fs);
-         goto error;
+         memset(file_path, 0, MAX_PATH);
```
jesperpedersen: What is all of this ?

src/main.c
```diff
@@ -2563,6 +2563,20 @@ init_receivewal(int server)
 
    if (strlen(config->common.servers[server].follow) == 0)
    {
+      if (config->common.servers[server].wal_streaming > 0)
```
jesperpedersen: What is this ?

test/testcases/test_wal_summary.c
```diff
@@ -186,7 +186,7 @@ MCTF_TEST(test_pgmoneta_wal_summary)
       MCTF_SKIP("failed to get checkpoint LSN");
    }
 
-   sleep(10);
+   sleep(30);
```
jesperpedersen: We need something better

jesperpedersen (changes requested):

jesperpedersen:
@sksingh2005 The rest looks ok

test/testcases/test_wal_summary.c
sksingh2005: Sure sir

src/libpgmoneta/walfile/wal_summary.c
sksingh2005: This block is a fallback after exponential backoff completes. If the compressed/encrypted WAL file still doesn't exist (compression in progress), we check for the uncompressed version before failing. This can handle edge cases where summarization runs before compression finishes. The fallback allows summarization to proceed using the complete (but not yet compressed) segment. Is my assumption incorrect ??

jesperpedersen: But `get_wal_file_name` and friends... Summary need to wait for compression / encryption
Cc: @ashu3103

test/testcases/test_wal_summary.c
sksingh2005: Currently working  on this only..

src/libpgmoneta/walfile/wal_summary.c
sksingh2005: Ohh got it !!

jesperpedersen:
Maybe we can work with

```
#include <sys/resource.h>
int setpriority(int which, id_t who, int prio);

#include <sched.h>
int sched_yield(void);
```

... we would need an API around it I think ...

jesperpedersen:
```
--- wal_summary ---
  test_pgmoneta_wal_summary (00:00:31,221) [SKIP] (test_wal_summary.c:216)
```

sksingh2005:
Sir now while running the test locally 
```
--- wal_summary ---
test_pgmoneta_wal_summary (00:03:02,154) [PASS]
```
1. Checkpoint LSN Obtained Successfully
`SELECT checkpoint_lsn, time...`
2. WAL File Compression Completed
`2026-02-01 06:58:14 TRACE wal.c:1820 WAL: Compressed/encrypted 000000010000000000000001`

3. WAL Summarization 
```
2026-02-01 06:58:44 DEBUG wal_summary.c:327 WAL file at /tmp/pgmoneta-test/base/backup/primary/wal/000000010000000000000001.zstd
2026-02-01 07:01:16 DEBUG wal_summary.c:132 summary dir: /tmp/pgmoneta-test/base/backup/primary/summary/
```

jesperpedersen:
it hangs here w/ `./test/check.sh`

sksingh2005:
Sir running `./test/check.sh` locally `test_pgmoneta_wal_summary` gets passed locally  [this](https://github.com/pgmoneta/pgmoneta/pull/895#issuecomment-3830084925) .
Also I see in CI it is getting passed once not in all though. Can you check the changes and confirm is it passing for you sir ?
Why is it happening like this sir?

sksingh2005:
@jesperpedersen sir should I than consider the changes coming from [#850]  then ??

jesperpedersen:
@sksingh2005 Until we have a stable testing "API" we have to consider all PRs... For now, rebase

sksingh2005:
@jesperpedersen sir can we retrigger the actions as it was down yesterday?? Or I have to push again ??

jesperpedersen:
@sksingh2005 running

src/libpgmoneta/wal.c
```diff
@@ -1740,6 +1740,9 @@ pgmoneta_wal_server_compress_encrypt(int srv, char** argv, char* wal_file)
       int retry_count = 0;
       char* d = NULL;
 
+      /* Boost priority for compression to complete faster (requires root) */
```
jesperpedersen: What do you mean ?

src/libpgmoneta/walfile/wal_summary.c
```diff
@@ -302,7 +302,14 @@ summarize_walfiles(int srv, char* dir_path, uint64_t start_lsn, uint64_t end_lsn
       char* file = (char*)file_iterator->value->data;
       char* fn = NULL;
 
-      fn = get_wal_file_name(file);
+      fn = get_wal_file_name(dir_path, file);
+
+      /* Skip files that aren't ready (e.g., still being compressed) */
```
jesperpedersen: No, we have to wait

src/libpgmoneta/walfile/wal_summary.c
```diff
@@ -317,39 +324,6 @@ summarize_walfiles(int srv, char* dir_path, uint64_t start_lsn, uint64_t end_lsn
       free(fn);
       fn = NULL;
 
-      retry_count = 0;
```
jesperpedersen, line 320: Keep this

src/libpgmoneta/walfile/wal_summary.c
```diff
@@ -416,11 +402,70 @@ get_wal_file_name(char* file)
...
 
-   if (suffix != NULL && !pgmoneta_ends_with(file, suffix))
+   /* Build expected filename */
+   fn = pgmoneta_append(NULL, base);
+   if (suffix != NULL && !pgmoneta_ends_with(base, suffix))
    {
       fn = pgmoneta_append(fn, suffix);
    }
 
+   /* Build full path for existence check */
```
jesperpedersen: So, you move it to a per file basis, yes ?

test/testcases/test_wal_summary.c
jesperpedersen: And, where are we on `sleep(30)` ?

src/libpgmoneta/wal.c
sksingh2005: This sets the compression process to `PRIORITY_HIGH` so it completes faster, reducing the wait time for the summary process. However, I can remove this if you prefer not to change process priorities.

src/libpgmoneta/walfile/wal_summary.c
sksingh2005: Yes, the waiting logic is now inside `get_wal_file_name()` on a per-file basis. If a file isn't ready after retries, the function returns NULL and that file is skipped, allowing the summary to continue with available files instead of failing entirely.

jesperpedersen: Yes, but that is wrong - we need all or nothing. So, you have to error out
@ashu3103 Cc:

src/libpgmoneta/wal.c
jesperpedersen: But, what does "requires root" mean ? The code is ok

src/libpgmoneta/walfile/wal_summary.c
sksingh2005: The `retry_count = 0` was reset here because the original code had the wait loop inline in `summarize_walfiles()`. Since I moved the wait logic into  `get_wal_file_name() ` (per-file basis as you noted), that function now has its own local `retry_count`. The outer `retry_count` in  `summarize_walfiles() `is only used for the WAL repository lock retry, so resetting it after each file is no longer needed.
 What you think about this sir ?

jesperpedersen: If a single WAL file fails then the whole thing should fail up the chain. All files or nothing.

I think that the per file is ok, since - likely - other files would have completed. However, we could now wait X x WAL files, instead of just X

src/libpgmoneta/wal.c
sksingh2005: What I meant/thought was that on some systems, raising process priority to negative  values requires elevated privileges.

jesperpedersen: well, then it is a no-op. pgmoneta can't run as root

src/libpgmoneta/walfile/wal_summary.c
sksingh2005: Ohh yes - worst case is X × N files total wait. In practice, if the first file is ready, others are likely ready too since compression runs continuously. Should I add a global timeout limit sir?

jesperpedersen: Each WAL file has its own process, so setting the priority is the key thing. Just leave as is, we are likely only waiting for the .partial -> WAL -> Compress -> Encryption cycle

src/libpgmoneta/wal.c
sksingh2005: `pgmoneta_set_priority()` handles this gracefully - if it can't set the priority, it just continues.

jesperpedersen: PTAL me once you are done - see new policy

sksingh2005:
@jesperpedersen sir regarding `sleep(30)` Initially I implemented a polling approach that checked for WAL files every second before calling summarization. However, I realized this was redundant since `get_wal_file_name()` in `wal_summary.c` already has per-file waiting with exponential backoff.

Final approach: I removed the polling loop entirely and now rely on the existing per-file waiting in  `get_wal_file_name()`. Uses the same robust waiting logic that was already implemented
The test now just gets the WAL directory and proceeds directly to summarization. The per-file waiting handles any compression/encryption delays.  
How is this approach sir ??

sksingh2005:
@jesperpedersen sir PTAL !!

test/testcases/test_wal_summary.c
```diff
@@ -177,17 +176,6 @@ MCTF_TEST(test_pgmoneta_wal_summary)
    }
    pgmoneta_test_cleanup_query_response(&qr);
 
-   // Append some more tuples just to ensure that a new wal segment is streamed
-
-   if (pgmoneta_server_checkpoint(PRIMARY_SERVER, srv_ssl, srv_socket, &tmp_lsn, &tli))
-   {
```
jesperpedersen, line 183: Why is this removed ?

jesperpedersen:
@sksingh2005 Yes, we should only wait in one place.

test/testcases/test_wal_summary.c
jesperpedersen: ?

sksingh2005: Sir this checkpoint was creating records in a second WAL segment  that remained `.partial` because it's the current active segment. The summarization code tries to process ALL WAL files in the directory, including this partial one, and times out waiting for it to become `.zstd`.

CI log showed:

```
WAL: waiting for .../000000010000000000000002.zstd (retry 10/10...)  
WAL summary: 000000010000000000000002.partial not available after waiting
```

jesperpedersen: @ashu3103 Cc:

src/libpgmoneta/walfile/wal_summary.c
ashu3103: Yes, @sksingh2005 you must error out in such cases. We need every WAL record for summarization. However, you can skip any file which you are sure contains WAL records beyond end_lsn of the summarizer

Hint: Summarization tells incremental backups which blocks (unit of data in a relation file) have been modified since the previous base backup. So we can't afford to lose even one WAL record.

test/testcases/test_wal_summary.c
ashu3103: I think this should not affect our summary process, so can be removed.

@sksingh2005, Also I think we can totally skip any `.partial` file we encounter as the WAL summarizer comes from `_backup_start` which anyways performs a switch WAL segment after checkpoint (which becomes our `end_lsn`).

@jesperpedersen Thoughts?

sksingh2005: Hi @ashu3103  sir , I tested the `.partial` skip approach, but found an issue from analyzing logs. In the test, sometimes the ONLY WAL file in the directory is `.partial` (e.g., `000000010000000000000001.partial`). If we skip all `.partial` files, no files get summarized at all.

The LSN-based check (`seg_start_lsn > end_lsn`) should already handle the case correctly - files beyond end_lsn get skipped regardless of whether they're `.partial` or not.

Should we:

- Skip `.partial` only if there are other completed segments to process?
- Rely solely on the LSN-based check?

I've already implemented the LSN-based approach.

sksingh2005: cc @jesperpedersen sir

jesperpedersen:
@ashu3103 @sksingh2005 We should know the `start_lsn` and `end_lsn` - that may be inside a `.partial` file that get renamed with a `CHECKPOINT`. So, there need to be a design for it - currently we are waiting for the WAL segment

test/testcases/test_wal_summary.c
ashu3103: @sksingh2005, Ignore my previous comment, it's actually misleading, the actual sequence of events is switch WAL -> CHECKPOINT. Refer [this](https://www.interdb.jp/pg/pgsql10/01.html#1011-pg_backup_start-ver14-or-earlier-pg_start_backup)

Apologies for the inconvenience caused.

ashu3103:
Hi @jesperpedersen @sksingh2005,

### Workaround

In `wal.c`

Just like we have `wal_repository` lock around compression and encryption of WAL segments we can extend that to locking of the functionality of renaming of `.partial` files too (you can find the code for this inside `wal_close()`).

In `wal_summary.c`

Instead of fetching the file names from the archive directory and waiting for their completion and compression/encryption, we can find the required WAL segments range directly from `start_lsn` and `end_lsn`. Now, for each file:

- Acquire `wal_repository` lock.
- check filename.partial ---- (if not exist) ---> filename ----- (if not exist) -----> filename.zstd (fallback mechanism)
- Read/Parse the file path that exists in the archive.
- Release `wal_repository` lock.

This will ensure that between searching for the file path and reading it, the filename doesn't transitions (due to `wal_repository` lock) avoids waiting for WAL segments completion.

Let me know if this is feasible @jesperpedersen @Jubilee101

jesperpedersen:
@ashu3103 Can you create a work branch for this ? Because we have https://github.com/pgmoneta/pgmoneta/blob/main/src/libpgmoneta/walfile/wal_summary.c#L380 - so we are waiting for the correct file names

ashu3103:
> @ashu3103 Can you create a work branch for this ? Because we have https://github.com/pgmoneta/pgmoneta/blob/main/src/libpgmoneta/walfile/wal_summary.c#L380 - so we are waiting for the correct file names

Sorry I didn't get what has to be done with the function attached above.

jesperpedersen:
@ashu3103 https://github.com/pgmoneta/pgmoneta/blob/main/src/libpgmoneta/walfile/wal_summary.c#L380 is ok as is. I don't understand your comment above, so I think it is best with a work branch with your ideas applied

ashu3103:
> @ashu3103 https://github.com/pgmoneta/pgmoneta/blob/main/src/libpgmoneta/walfile/wal_summary.c#L380 is ok as is. I don't understand your comment above, so I think it is best with a work branch with your ideas applied

Ah, okay

test/testcases/test_wal_summary.c
sksingh2005: I now understand the sequence clearly sir:

`pg_backup_start` → (1) switch WAL → (2) CHECKPOINT → returns checkpoint LSN

Since the WAL switch happens first, the checkpoint record lands in the new segment (N+1), which is `.partial`. This means end_lsn points inside a `.partial` file.

Is I am right sir ?

sksingh2005:
@jesperpedersen sir PTAL !!

sksingh2005:
@jesperpedersen sir I have implemented a design for `.partial` file handling. 
`get_wal_file_name`  now resolves files in three tiers matching the WAL lifecycle: compressed → uncompressed → `.partial`. This way if `end_lsn` is inside a `.partial` segment, we read it directly instead of waiting for compression that won't happen during this backup. All-or-nothing error handling is preserved — if no form of the file is found after retries, we error out.
Is this good sir?

jesperpedersen:
Rebased, and merged.

ashu3103:
> @ashu3103 https://github.com/pgmoneta/pgmoneta/blob/main/src/libpgmoneta/walfile/wal_summary.c#L380 is ok as is. I don't understand your comment above, so I think it is best with a work branch with your ideas applied

@jesperpedersen is the `wal_summary` working as expected or shall I proceed with this?

jesperpedersen:
@ashu3103 Please, double-check. The logic seems to handle corner-cases better now

ashu3103:
> @ashu3103 Please, double-check. 

Sure

> The logic seems to handle corner-cases better now

Great!

Issue #900, opened 2026-02-01
Authentication problem with pgmoneta-cli


ahmed-kamal2004:
**Describe the bug**
Authentication issue in pgmoneta-cli

**To Reproduce**
to reproduce this issue, I have just followed the Developer Guide
![Guide](https://pgmoneta.github.io/developers.html)

and got the issue at pgmoneta-cli authentication
```

pgmoneta-cli -U repl -P ahmed -c pgmoneta.conf backup primary
pgmoneta-cli: pgmoneta-cli: Authentication failed for user 'repl'. Verify username and password.

```
**Version**
The upstream (main).

**PostgreSQL**
PostgreSQL 16

**OS**

Ubuntu 24.04.2 LTS

**Configuration**
`pgmoneta.conf`

```
[pgmoneta]
host = *
metrics = 5001

base_dir = /home/pgmoneta/backup

compression = zstd

retention = 7

log_type = file
log_level = info
log_path = /tmp/pgmoneta.log

unix_socket_dir = /tmp/
[primary]
host = localhost
port = 5432
user = repl
wal_slot = repl

```

**Debug logs**
postgres logs
```
2026-02-01 18:20:48.137 EET [112782] repl@admin FATAL:  no pg_hba.conf entry for host "127.0.0.1", user "repl", database "admin", no encryption
2026-02-01 18:52:58.820 EET [127350] repl@admin FATAL:  no pg_hba.conf entry for host "127.0.0.1", user "repl", database "admin", no encryption
2026-02-01 18:53:15.244 EET [127452] repl@admin FATAL:  no pg_hba.conf entry for host "127.0.0.1", user "repl", database "admin", no encryption
2026-02-01 18:53:40.927 EET [127831] repl@admin FATAL:  no pg_hba.conf entry for host "127.0.0.1", user "repl", database "admin", no encryption
2026-02-01 18:54:15.278 EET [128270] repl@admin FATAL:  no pg_hba.conf entry for host "127.0.0.1", user "repl", database "admin", no encryption
2026-02-01 18:54:22.952 EET [128319] repl@admin FATAL:  no pg_hba.conf entry for host "127.0.0.1", user "repl", database "admin", no encryption
2026-02-01 18:54:52.045 EET [128724] repl@admin FATAL:  no pg_hba.conf entry for host "127.0.0.1", user "repl", database "admin", no encryption
[...]

PR #908, opened 2026-02-03
[#900] docs: use separate configuration file for pgmoneta_cli

ahmed-kamal2004:
pgmoneta CLI gives connection over tcp `host` and `port` a higher priority than `unix socket` connection. and using tcp connection resulted in me facing this issue #900 

This PR proposes a solution/change in the  `DEVELOPERS.md` to separate the configuration between `pgmoneta` and `pgmoneta-cli` and makes `pgmoneta-cli` uses it's own config file relying only on connection over `unix_socket_dir`.

**Tests**
image

### DIFF
--- a/doc/DEVELOPERS.md
+++ b/doc/DEVELOPERS.md
@@ -357,6 +357,19 @@ port = 5432
 user = repl
 wal_slot = repl
 ```
+And create the `pgmoneta_cli.conf` configuration file to use when running **pgmoneta-cli**.
+
+``` ini
+cat > pgmoneta_cli.conf
+[pgmoneta-cli]
+compression = zstd
+
+log_type = file
+log_level = info
+log_path = /tmp/pgmoneta-cli.log
+
+unix_socket_dir = /tmp/
+```
 
 In our main section called `[pgmoneta]` we setup [**pgmoneta**](https://github.com/pgmoneta/pgmoneta) to listen on all network addresses. We will enable Prometheus metrics on port 5001 and have the backups live in the `/home/pgmoneta/backup` directory.
 All backups are being compressed with zstd and kept for 7 days. Logging will be performed at `info` level and put in a file called `/tmp/pgmoneta.log`. Last we specify the location of the `unix_socket_dir` used for management operations.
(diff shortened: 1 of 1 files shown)


doc/DEVELOPERS.md
```diff
@@ -357,6 +357,26 @@ port = 5432
 user = repl
 wal_slot = repl
 ```
+And create the `pgmoneta_cli.conf` configuration file to use when running **pgmoneta-cli**.
```
jesperpedersen, line 360: This doesn't look correct. `pgmoneta_cli.conf` isn't the same as `pgmoneta.conf`

ahmed-kamal2004: wdyt should be done ?
I am using `pgmoneta_cli.conf` with `pgmoneta-cli` only

ahmed-kamal2004: hmm, we can remove the unneeded config from `pgmoneta_cli.conf` ?

jesperpedersen: It should print a warning that the option isn't supported. You have to look at the configuration documentation and create a proper file

ahmed-kamal2004: great
so to check my understanding 
- I will make `pgmoneta-cli` print a warning if it uses `host` and `port` instead of `unix socket`
- I will update the `pgmoneta_cli.conf` file I added within this pr to align properly with the docs in the guide

jesperpedersen: `host` and `port` are used for a remote pgmoneta instance, so they have priority if specified.

Yes, you need to update your configuration such that it follows the supported options

doc/DEVELOPERS.md
```diff
@@ -357,6 +357,19 @@ port = 5432
...
+
+``` ini
+cat > pgmoneta_cli.conf
+[pgmoneta]
```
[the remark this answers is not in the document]
jesperpedersen: It should be `[pgmoneta-cli]`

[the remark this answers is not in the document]
ahmed-kamal2004: great done
I verified it locally

jesperpedersen:
Rebased, and merged.

Issue #884, opened 2026-01-27
walinfo: Support TAR archives


jesperpedersen:
Add support for TAR - optionally compressed - with WAL files inside

jesperpedersen:
Maybe we need a function that return a bitmask of a file type... like a general function of `pgmoneta_is_compressed`

sksingh2005:
Can I take this issue sir ??

jesperpedersen:
@sksingh2005 Go for it !

PR #914, opened 2026-02-07
[#884] feat:add TAR archive support and file type bitmask detection to walinfo

sksingh2005:
## Description
This PR adds support for TAR archives (including compressed variants) to the walinfo tool and introduces a general file type detection function that returns a bitmask.

## Add TAR archive support and file type bitmask detection to walinfo

Add support for reading WAL files from TAR archives (including compressed variants) and introduces a general file type bitmask detection function.

- [`pgmoneta_get_file_type()`](cci:1://file:///home/shashank-singh/pgmoneta/src/libpgmoneta/utils.c:4427:0-4502:1) - Returns bitmask for WAL/compressed/encrypted/TAR/partial detection
- [`pgmoneta_is_tar_archive()`](cci:1://file:///home/shashank-singh/pgmoneta/src/libpgmoneta/utils.c:4406:0-4425:1) - TAR archive detection
- [`pgmoneta_extract_wal_from_tar()`](cci:1://file:///home/shashank-singh/pgmoneta/src/libpgmoneta/utils.c:4504:0-4635:1) - Extract WAL files from TAR archives

### Supported TAR formats

`.tar`, `.tar.gz`, `.tgz`, `.tar.zstd`, `.tar.lz4`, `.tar.bz2`

### Usage

```bash
pgmoneta-walinfo /path/to/wal_backup.tar.gz

### DIFF
--- a/doc/man/pgmoneta-walinfo.1.rst
+++ b/doc/man/pgmoneta-walinfo.1.rst
@@ -2,21 +2,26 @@
 pgmoneta-walinfo
 =====================
 
-----------------------------------------------------------------------
 Command line utility to read and display Write-Ahead Log (WAL) files
-----------------------------------------------------------------------
 
 :Manual section: 1
 
 SYNOPSIS
 ========
 
-pgmoneta-walinfo <file>
+pgmoneta-walinfo <file|directory|tar_archive>
 
 DESCRIPTION
 ===========
 
-pgmoneta-walinfo is a command line utility to read and display information about PostgreSQL Write-Ahead Log (WAL) files. It provides details of the WAL file in either raw or JSON format.
+pgmoneta-walinfo is a command line utility to read and display information about PostgreSQL Write-Ahead Log (WAL) files. It supports individual WAL files, directories containing WAL files, TAR archives (optionally compressed with .gz, .lz4, .zstd, or .bz2), and encrypted files (.aes). It provides details of the WAL file in either raw or JSON format.
+
+In addition to standard WAL files, pgmoneta-walinfo also supports:
+
+- Encrypted WAL files (**.aes**)
+- Compressed WAL files: **.zstd**, **.gz**, **.lz4**, and **.bz2**
+- TAR archives containing WAL files (**.tar**)
+- Directories containing WAL files
 
 OPTIONS
 =======
@@ -75,8 +80,8 @@ OPTIONS
 ARGUMENTS
 =========
 
-<file>
-  The path to the WAL file to be analyzed.
+<file|directory|tar_archive>
+  The path to the WAL file, directory containing WAL files, or TAR archive to be analyzed.
 
 USAGE
 =====
@@ -88,6 +93,10 @@ To display information about a WAL file in raw format:
 To display information in JSON format:
 
     pgmoneta-walinfo -F json /path/to/walfile
+
+To analyze WAL files from a TAR archive:
+
+    pgmoneta-walinfo /path/to/wal_backup.tar.gz
 
 To display information and translate the OIDs to the corresponding object names:
 
--- a/doc/manual/en/17-wal-tools.md
+++ b/doc/manual/en/17-wal-tools.md
@@ -11,7 +11,12 @@ pgmoneta provides two powerful command-line utilities for working with PostgreSQ
 
 `pgmoneta-walinfo` is a command-line utility designed to read and display information about PostgreSQL Write-Ahead Log (WAL) files. The tool provides output in either raw or JSON format, making it easy to analyze WAL files for debugging, auditing, or general information purposes.
 
-In addition to standard WAL files, `pgmoneta-walinfo` also supports encrypted (**aes**) and compressed WAL files in the following formats: **zstd**, **gz**, **lz4**, and **bz2**.
+In addition to standard WAL files, `pgmoneta-walinfo` also supports:
+
+- Encrypted WAL files (**.aes**)
+- Compressed WAL files: **.zstd**, **.gz**, **.lz4**, and **.bz2**
+- TAR archives containing WAL files (**.tar**)
+- Directories containing WAL files
 
 ### Usage
 
@@ -20,7 +25,7 @@ pgmoneta-walinfo 0.20.0
   Command line utility to read and display Write-Ahead Log (WAL) files
 
 Usage:
-  pgmoneta-walinfo <file>
+  pgmoneta-walinfo <file|directory|tar_archive>
 
 Options:
   -c,  --config      Set the path to the pgmoneta_walinfo.conf file
@@ -103,6 +108,18 @@ pgmoneta-walinfo -S /path/to/walfile
 
 ```bash
 pgmoneta-walinfo -r Heap -l 10 /path/to/walfile
+```
+
+6. **Analyze a directory containing WAL files:**
+
+```bash
+pgmoneta-walinfo /path/to/wal_directory/
+```
+
+7. **Analyze WAL files from a TAR archive:**
+
+```bash
+pgmoneta-walinfo /path/to/wal_backup.tar.gz
 ```
 
 ### OID Translation
--- a/src/include/achv.h
+++ b/src/include/achv.h
@@ -53,15 +53,6 @@ void
 pgmoneta_archive(SSL* ssl, int client_fd, int server, uint8_t compression, uint8_t encryption, struct json* request);
 
 /**
- * Extract from a tar file to a given directory
- * @param file_path The tar file path
- * @param destination The destination to extract to
- * @return 0 upon success, otherwise 1
- */
-int
-pgmoneta_extract_tar_file(char* file_path, char* destination);
-
-/**
  * Create a tar archive of the given directory
  * @param src The source directory
  * @param dst The destination tar file path
--- a/src/include/utils.h
+++ b/src/include/utils.h
@@ -69,6 +69,19 @@ extern "C" {
 #define COLOR_WHITE        "\033[97m"
 #define COLOR_RESET        "\033[0m" /* Reset to default color */
 
+/* File type bitmask constants */
+#define PGMONETA_FILE_TYPE_UNKNOWN    0x0000 /* Unknown file type */
+#define PGMONETA_FILE_TYPE_WAL        0x0001 /* WAL file (24-char hex name) */
+#define PGMONETA_FILE_TYPE_COMPRESSED 0x0002 /* Compressed (any type) */
+#define PGMONETA_FILE_TYPE_GZIP       0x0004 /* Compressed with gzip (.gz) */
+#define PGMONETA_FILE_TYPE_LZ4        0x0008 /* Compressed with lz4 (.lz4) */
+#define PGMONETA_FILE_TYPE_ZSTD       0x0010 /* Compressed with zstd (.zstd) */
+#define PGMONETA_FILE_TYPE_BZ2        0x0020 /* Compressed with bzip2 (.bz2) */
+#define PGMONETA_FILE_TYPE_ENCRYPTED  0x0040 /* Encrypted (.aes) */
+#define PGMONETA_FILE_TYPE_TAR        0x0080 /* TAR archive (.tar) */
+#define PGMONETA_FILE_TYPE_PARTIAL    0x0100 /* Partial file (.partial) */
+#define PGMONETA_FILE_TYPE_ALL        0xFFFF /* Match all file types */
+
 /** @struct signal_info
  * Defines the signal structure
  */
@@ -700,13 +713,28 @@ int
 pgmoneta_delete_directory(char* path);
 
 /**
- * Get files
- * @param base The base directory
- * @param files The deque of files
- * @return The result
+ * Get files with type filtering and optional recursion
+ * @param file_type_mask Bitmask of file types to include (e.g., PGMONETA_FILE_TYPE_WAL | PGMONETA_FILE_TYPE_TAR)
+ *                       Use PGMONETA_FILE_TYPE_ALL to match all files
+ * @param base The base directory to scan
+ * @param recursive If true, scan subdirectories recursively
+ * @param files Output deque of matching file paths
+ *              Recursive mode returns full paths; non-recursive mode returns basenames
+ * @return 0 on success, 1 on error
  */
 int
-pgmoneta_get_files(char* base, struct deque** files);
+pgmoneta_get_files(uint32_t file_type_mask, char* base, bool recursive, struct deque** files);
+
+/**
+ * Extract an archive file to a given directory
+ * File type is detected internally via pgmoneta_get_file_type
+ * Handles layered formats (e.g., file.tar.zstd.aes)
+ * @param file_path The archive file path
+ * @param destination The destination to extract to
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_extract_file(char* file_path, char* destination);
 
 /**
  * Get WAL files
@@ -1227,6 +1255,24 @@ pgmoneta_is_encrypted(char* file_path);
  */
 bool
 pgmoneta_is_compressed(char* file_path);
+
+/**
+ * Get the file type bitmask for a given file path
+ * The bitmask can include combinations of:
+ * - PGMONETA_FILE_TYPE_WAL (24-char hex WAL file)
+ * - PGMONETA_FILE_TYPE_COMPRESSED (any compression)
+ * - PGMONETA_FILE_TYPE_GZIP (.gz)
+ * - PGMONETA_FILE_TYPE_LZ4 (.lz4)
+ * - PGMONETA_FILE_TYPE_ZSTD (.zstd)
+ * - PGMONETA_FILE_TYPE_BZ2 (.bz2)
+ * - PGMONETA_FILE_TYPE_ENCRYPTED (.aes)
+ * - PGMONETA_FILE_TYPE_TAR (.tar)
+ * - PGMONETA_FILE_TYPE_PARTIAL (.partial)
+ * @param file_path The file path to check
+ * @return Bitmask of file type flags
+ */
+uint32_t
+pgmoneta_get_file_type(char* file_path);
 
 /**
  * Init a token bucket
--- a/src/libpgmoneta/archive.c
+++ b/src/libpgmoneta/archive.c
@@ -29,9 +29,7 @@
 /* pgmoneta */
 #include <pgmoneta.h>
 #include <achv.h>
-#include <gzip_compression.h>
 #include <logging.h>
-#include <lz4_compression.h>
 #include <management.h>
 #include <manifest.h>
 #include <network.h>
@@ -39,7 +37,6 @@
 #include <security.h>
 #include <utils.h>
 #include <workflow.h>
-#include <zstandard_compression.h>
 
 #include <archive.h>
 #include <archive_entry.h>
@@ -51,6 +48,7 @@
 #define NAME "archive"
 
 static bool is_server_side_compression(void);
+static const char* basebackup_archive_extension(void);
 
 static void write_tar_file(struct archive* a, char* src, char* dst);
 
@@ -276,88 +274,6 @@ error:
 }
 
 int
-pgmoneta_extract_tar_file(char* file_path, char* destination)
-{
-   char* archive_name = NULL;
-   struct archive* a;
-   struct archive_entry* entry;
-   struct main_configuration* config;
-
-   config = (struct main_configuration*)shmem;
-
-   a = archive_read_new();
-   archive_read_support_format_tar(a);
-
-   if (config->compression_type == COMPRESSION_SERVER_GZIP)
-   {
-      archive_name = pgmoneta_append(archive_name, file_path);
-      archive_name = pgmoneta_append(archive_name, ".gz");
-      pgmoneta_move_file(file_path, archive_name);
-      pgmoneta_gunzip_file(archive_name, file_path);
-   }
-   else if (config->compression_type == COMPRESSION_SERVER_ZSTD)
-   {
-      archive_name = pgmoneta_append(archive_name, file_path);
-      archive_name = pgmoneta_append(archive_name, ".zstd");
-      pgmoneta_move_file(file_path, archive_name);
-      pgmoneta_zstandardd_file(archive_name, file_path);
-   }
-   else if (config->compression_type == COMPRESSION_SERVER_LZ4)
-   {
-      archive_name = pgmoneta_append(archive_name, file_path);
-      archive_name = pgmoneta_append(archive_name, ".lz4");
-      pgmoneta_move_file(file_path, archive_name);
-      pgmoneta_lz4d_file(archive_name, file_path);
-   }
-   else
-   {
-      archive_name = pgmoneta_append(archive_name, file_path);
-   }
-
-   // open tar file in a suitable buffer size, I'm using 10240 here
-   if (archive_read_open_filename(a, file_path, 10240) != ARCHIVE_OK)
-   {
-      pgmoneta_log_error("Failed to open the tar file for reading");
-      goto error;
-   }
-
-   while (archive_read_next_header(a, &entry) == ARCHIVE_OK)
-   {
-      char dst_file_path[MAX_PATH];
-      memset(dst_file_path, 0, sizeof(dst_file_path));
-      const char* entry_path = archive_entry_pathname(entry);
-      if (pgmoneta_ends_with(destination, "/"))
-      {
-         snprintf(dst_file_path, sizeof(dst_file_path), "%s%s", destination, entry_path);
-      }
-      else
-      {
-         snprintf(dst_file_path, sizeof(dst_file_path), "%s/%s", destination, entry_path);
-      }
-
-      archive_entry_set_pathname(entry, dst_file_path);
-      if (archive_read_extract(a, entry, 0) != ARCHIVE_OK)
-      {
-         pgmoneta_log_error("Failed to extract entry: %s", archive_error_string(a));
-         goto error;
-      }
-   }
-
-   free(archive_name);
-
-   archive_read_close(a);
-   archive_read_free(a);
-   return 0;
-
-error:
-   free(archive_name);
-
-   archive_read_close(a);
-   archive_read_free(a);
-   return 1;
-}
-
-int
 pgmoneta_tar_directory(char* src, char* dst, char* destination)
 {
    struct archive* a = NULL;
@@ -409,6 +325,7 @@ pgmoneta_receive_archive_files(int srv, SSL* ssl, int socket, struct stream_buff
    {
       char file_path[MAX_PATH];
       char directory[MAX_PATH];
+      const char* archive_ext = basebackup_archive_extension();
       memset(file_path, 0, sizeof(file_path));
       memset(directory, 0, sizeof(directory));
       if (tup->data[1] == NULL)
@@ -416,12 +333,12 @@ pgmoneta_receive_archive_files(int srv, SSL* ssl, int socket, struct stream_buff
          // main data directory
          if (pgmoneta_ends_with(basedir, "/"))
          {
-            snprintf(file_path, sizeof(file_path), "%sdata/%s", basedir, "base.tar");
+            snprintf(file_path, sizeof(file_path), "%sdata/base%s", basedir, archive_ext);
             snprintf(directory, sizeof(directory), "%sdata/", basedir);
          }
          else
          {
-            snprintf(file_path, sizeof(file_path), "%s/data/%s", basedir, "base.tar");
+            snprintf(file_path, sizeof(file_path), "%s/data/base%s", basedir, archive_ext);
             snprintf(directory, sizeof(directory), "%s/data/", basedir);
          }
       }
@@ -440,12 +357,12 @@ pgmoneta_receive_archive_files(int srv, SSL* ssl, int socket, struct stream_buff
          }
          if (pgmoneta_ends_with(basedir, "/"))
          {
-            snprintf(file_path, sizeof(file_path), "%stblspc_%s/%s.tar", basedir, tblspc->name, tblspc->name);
+            snprintf(file_path, sizeof(file_path), "%stblspc_%s/%s%s", basedir, tblspc->name, tblspc->name, archive_ext);
             snprintf(directory, sizeof(directory), "%stblspc_%s/", basedir, tblspc->name);
          }
          else
          {
-            snprintf(file_path, sizeof(file_path), "%s/tblspc_%s/%s.tar", basedir, tblspc->name, tblspc->name);
+            snprintf(file_path, sizeof(file_path), "%s/tblspc_%s/%s%s", basedir, tblspc->name, tblspc->name, archive_ext);
             snprintf(directory, sizeof(directory), "%s/tblspc_%s/", basedir, tblspc->name);
          }
       }
@@ -523,7 +440,7 @@ pgmoneta_receive_archive_files(int srv, SSL* ssl, int socket, struct stream_buff
       fclose(file);
 
       // extract the file
-      pgmoneta_extract_tar_file(file_path, directory);
+      pgmoneta_extract_file(file_path, directory);
       remove(file_path);
       pgmoneta_free_message(msg);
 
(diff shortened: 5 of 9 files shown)

### CODE REVIEW

File: doc/manual/en/17-wal-tools.md [hunk 1]
```diff
@@ -11,7 +11,12 @@ pgmoneta provides two powerful command-line utilities for working with PostgreSQ
 
 `pgmoneta-walinfo` is a command-line utility designed to read and display information about PostgreSQL Write-Ahead Log (WAL) files. The tool provides output in either raw or JSON format, making it easy to analyze WAL files for debugging, auditing, or general information purposes.
 
-In addition to standard WAL files, `pgmoneta-walinfo` also supports encrypted (**aes**) and compressed WAL files in the following formats: **zstd**, **gz**, **lz4**, and **bz2**.
+In addition to standard WAL files, `pgmoneta-walinfo` also supports:
+
+- Encrypted WAL files (**aes**)
```
jesperpedersen: You should copy this to the man-page too

doc/manual/en/17-wal-tools.md
```diff
@@ -105,6 +110,18 @@ pgmoneta-walinfo -S /path/to/walfile
 pgmoneta-walinfo -r Heap -l 10 /path/to/walfile
 ```
 
+6. **Analyze WAL files from a TAR archive:**
+
+```bash
+pgmoneta-walinfo /path/to/wal_backup.tar.gz
+```
+
+7. **Analyze a directory containing WAL files:**
```
jesperpedersen: Swap 6. and 7.

src/include/utils.h
```diff
@@ -69,6 +69,14 @@ extern "C" {
 #define COLOR_WHITE        "\033[97m"
 #define COLOR_RESET        "\033[0m" /* Reset to default color */
 
+/* File type bitmask constants */
+#define PGMONETA_FILE_TYPE_UNKNOWN    0x00 /* Unknown file type */
+#define PGMONETA_FILE_TYPE_WAL        0x01 /* WAL file (24-char hex name) */
+#define PGMONETA_FILE_TYPE_COMPRESSED 0x02 /* Compressed (.gz, .lz4, .zstd, .bz2) */
```
jesperpedersen: We should have a `#define`  per compression type

src/include/utils.h
```diff
@@ -1228,6 +1236,46 @@ pgmoneta_is_encrypted(char* file_path);
...
+ * @return True if TAR archive, otherwise false
+ */
+bool
+pgmoneta_is_tar_archive(char* file_path);
+
+/**
+ * Get the file type bitmask for a given file path
+ * The bitmask can include combinations of:
+ * - PGMONETA_FILE_TYPE_WAL (24-char hex WAL file)
+ * - PGMONETA_FILE_TYPE_COMPRESSED (.gz, .lz4, .zstd, .bz2)
```
jesperpedersen: As above

src/include/utils.h
```diff
@@ -1228,6 +1236,46 @@ pgmoneta_is_encrypted(char* file_path);
...
+
+/**
+ * Extract WAL files from a TAR archive to a temporary directory
+ * @param tar_path Path to the TAR archive
+ * @param temp_dir Output: path to temp directory containing extracted WAL files
+ * @param wal_files Output: deque of extracted WAL file paths
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_extract_wal_from_tar(char* tar_path, char** temp_dir, struct deque** wal_files);
```
jesperpedersen: arch.h - pgmoneta_extract_tar_file ?

src/include/utils.h
```diff
@@ -1228,6 +1236,46 @@ pgmoneta_is_encrypted(char* file_path);
...
+
+/**
+ * Cleanup temporary directory created by pgmoneta_extract_wal_from_tar
+ * @param temp_dir The temporary directory to remove
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_cleanup_tar_extraction(char* temp_dir);
```
jesperpedersen: pgmoneta_delete_directory ?

jesperpedersen:
@sksingh2005 Lets start with those - look through your patch

src/include/utils.h
sksingh2005: Sir  existing `pgmoneta_extract_tar_file` in `archive.c` is little different because:

- It relies on `config->compression_type` from shared memory (server config)
- It extracts ALL files from the TAR
- Our function auto-detects compression from file extension and only extracts WAL files

src/include/utils.h
sksingh2005: Missed that. Removing the wrapper and will use it directly now. Thanks sir !

src/include/utils.h
jesperpedersen: My point is that we need to combine code when we can, so maybe `config->compression_type` and `config->encryption_type` can be passed in as the `uint32_t`. We should have less code ideally

sksingh2005: Sir should I refactor `pgmoneta_extract_tar_file` to accept an optional compression type parameter (use auto-detection if not provided)?
Since one problem is that `walinfo` is a standalone CLI tool that runs without the pgmoneta server (no shared memory or config). So I think it needs to auto-detect compression from the file extension since 
`config->compression_type` isn't available.

jesperpedersen: We should use your new pgmoneta_get_file_type function - we don't need pgmoneta_is_tar_archive - the bitmask will tell us that.

You pass in the bitmask to pgmoneta_extract_tar_file, so you load that first, and the functions guard against an invalid bitmask - e.g. pgmoneta_extract_tar_file need the TAR bit

jesperpedersen:
@sksingh2005 Please, rebase and account for the new WAL checks

sksingh2005:
@jesperpedersen sir rebased and updated through the checks

saraanbih:
Hi @sksingh2005 
Good Work, what about if this tar contain sub folders? u check only the valid WAL files, can u add check if it is contain directory also scan all valid WAL files in it?

sksingh2005:
Hi @saraanbih !!
Added `scan_directory_for_wal_files()`  recursive helper that scans extracted TAR directories for subdirectories and processes all valid WAL files found at any nesting level. Can you check that if it is good please !!
Considered reusing existing functions like `describe_walfiles_in_directory`or `pgmoneta_get_wal_files`, but they are not recursive and modifying them could affect other.

saraanbih:
Great @sksingh2005 ,
Can u look at walinfo interaction, walfilter?
When u add a new file type you have to think about all the tools, and
how to interact with that type.

jesperpedersen:
We should not have 100s of functions that does the same thing - we should think along the lines of 

```
  /* Get WAL files and standalone files; with possible WAL files inside TAR archives */
  pgmoneta_get_files(FILE_TYPE_WAL | FILE_TYPE_TAR | FILE_TYPE_AES | ..., directory, true, &files);

  /* Extract the files that need it ... */
 ...
```

jesperpedersen:
Remember `FILE_TYPE_ALL`

sksingh2005:
Hi @saraanbih 
`pgmoneta_get_files_ext(PGMONETA_FILE_TYPE_WAL | PGMONETA_FILE_TYPE_TAR, directory, true, &files);
`
walinfo now uses this API. For walfilter, should TAR support be added:

- As a separate follow-up PR? As this was mainly for walinfo in this pr 
- In this PR (scope expansion)?

jesperpedersen:
@sksingh2005 I would do walfilter in this PR as well, so we don't have to circle back to it

doc/man/pgmoneta-walinfo.1.rst
```diff
@@ -2,21 +2,28 @@
 pgmoneta-walinfo
 =====================
 
-----------------------------------------------------------------------
-Command line utility to read and display Write-Ahead Log (WAL) files
-----------------------------------------------------------------------
+--------------------------------------------------------------------------------------
```
jesperpedersen: Remove this - it is confusing

doc/man/pgmoneta-walinfo.1.rst
```diff
@@ -2,21 +2,28 @@
...
 ========
 
-pgmoneta-walinfo <file>
+pgmoneta-walinfo <file|directory|tar_archive>
 
 DESCRIPTION
 ===========
 
-pgmoneta-walinfo is a command line utility to read and display information about PostgreSQL Write-Ahead Log (WAL) files. It provides details of the WAL file in either raw or JSON format.
+pgmoneta-walinfo is a command line utility to read and display information about PostgreSQL Write-Ahead Log (WAL) files. It supports individual WAL files, directories containing WAL files, and TAR archives (optionally compressed with gzip, lz4, zstd, or bz2). It provides details of the WAL file in either raw or JSON format.
```
jesperpedersen: And .aes ?

doc/man/pgmoneta-walinfo.1.rst
```diff
@@ -2,21 +2,28 @@
...
+
+In addition to standard WAL files, pgmoneta-walinfo also supports:
+
+- Encrypted WAL files (**aes**)
```
jesperpedersen: aes -> .aes

doc/man/pgmoneta-walinfo.1.rst
```diff
@@ -2,21 +2,28 @@
...
+- Compressed WAL files: **zstd**, **gz**, **lz4**, and **bz2**
```
jesperpedersen: Same

doc/man/pgmoneta-walinfo.1.rst
```diff
@@ -2,21 +2,28 @@
...
+- TAR archives containing WAL files: **.tar**, **.tar.gz**, **.tgz**, **.tar.zstd**, **.tar.lz4**, **.tar.bz2**
```
jesperpedersen: I would just write .tar

doc/man/pgmoneta-walinfo.1.rst
```diff
@@ -75,8 +82,9 @@ OPTIONS
 ARGUMENTS
 =========
 
-<file>
-  The path to the WAL file to be analyzed.
+<file|directory|tar_archive>
+  The path to the WAL file, directory containing WAL files, or TAR archive to be analyzed.
+  Supported TAR formats: .tar, .tar.gz, .tgz, .tar.lz4, .tar.zstd, .tar.bz2
```
jesperpedersen: Remove this line

doc/manual/en/17-wal-tools.md
```diff
@@ -11,7 +11,12 @@ pgmoneta provides two powerful command-line utilities for working with PostgreSQ
 
 `pgmoneta-walinfo` is a command-line utility designed to read and display information about PostgreSQL Write-Ahead Log (WAL) files. The tool provides output in either raw or JSON format, making it easy to analyze WAL files for debugging, auditing, or general information purposes.
 
-In addition to standard WAL files, `pgmoneta-walinfo` also supports encrypted (**aes**) and compressed WAL files in the following formats: **zstd**, **gz**, **lz4**, and **bz2**.
+In addition to standard WAL files, `pgmoneta-walinfo` also supports:
+
+- Encrypted WAL files (**aes**)
+- Compressed WAL files: **zstd**, **gz**, **lz4**, and **bz2**
+- TAR archives containing WAL files: **.tar**, **.tar.gz**, **.tgz**, **.tar.zstd**, **.tar.lz4**, **.tar.bz2**
+- Directories containing WAL files
```
jesperpedersen, line 19: Same comments as above

src/include/utils.h
```diff
@@ -69,6 +69,19 @@ extern "C" {
 #define COLOR_WHITE        "\033[97m"
 #define COLOR_RESET        "\033[0m" /* Reset to default color */
 
+/* File type bitmask constants */
+#define PGMONETA_FILE_TYPE_UNKNOWN    0x0000 /* Unknown file type */
+#define PGMONETA_FILE_TYPE_WAL        0x0001 /* WAL file (24-char hex name) */
+#define PGMONETA_FILE_TYPE_COMPRESSED 0x0002 /* Compressed (any type) */
+#define PGMONETA_FILE_TYPE_GZIP       0x0004 /* Compressed with gzip (.gz) */
+#define PGMONETA_FILE_TYPE_LZ4        0x0008 /* Compressed with lz4 (.lz4) */
+#define PGMONETA_FILE_TYPE_ZSTD       0x0010 /* Compressed with zstd (.zstd) */
```
jesperpedersen, line 78: what happen to 0A ?

src/include/achv.h
```diff
@@ -56,10 +56,12 @@ pgmoneta_archive(SSL* ssl, int client_fd, int server, uint8_t compression, uint8
  * Extract from a tar file to a given directory
  * @param file_path The tar file path
  * @param destination The destination to extract to
+ * @param file_type The file type bitmask from pgmoneta_get_file_type()
+ *                  Must have PGMONETA_FILE_TYPE_TAR bit set
  * @return 0 upon success, otherwise 1
  */
 int
-pgmoneta_extract_tar_file(char* file_path, char* destination);
+pgmoneta_extract_tar_file(char* file_path, char* destination, uint32_t file_type);
```
jesperpedersen: It is the same deal here - `pgmoneta_extract_file` that can extract any file based on the *internal* call to `pgmoneta_get_file_type` - e.g. not part of the signature

src/include/utils.h
```diff
@@ -708,6 +721,18 @@ pgmoneta_delete_directory(char* path);
...
+ * Get files with type filtering and optional recursion
+ * @param file_type_mask Bitmask of file types to include (e.g., PGMONETA_FILE_TYPE_WAL | PGMONETA_FILE_TYPE_TAR)
+ *                       Use PGMONETA_FILE_TYPE_ALL to match all files
+ * @param base The base directory to scan
+ * @param recursive If true, scan subdirectories recursively and extract TAR archives
+ * @param files Output deque of matching file paths (full paths)
+ * @return 0 on success, 1 on error
+ */
+int
+pgmoneta_get_files_ext(uint32_t file_type_mask, char* base, bool recursive, struct deque** files);
```
jesperpedersen: There should only be one function

jesperpedersen (changes requested):

jesperpedersen:
@sksingh2005 I'll stop here for now. You have to unify functions such that there is a main function to call which can use other functions to work

src/include/utils.h
sksingh2005: Sir could you please clarify what you mean by "what happen to 0A?". I did not understood actually 
The current file type constants use powers of 2 (0x0001, 0x0002, 0x0004 ... ) for bitmask operations so they can be combined with `|`. Since `0x0A` (decimal 10) is not a power of 2, it wouldn't work as a unique bitmask flag.
Are you suggesting:
- A different numbering scheme?

jesperpedersen: No, disregard. Keep the bitmask

doc/man/pgmoneta-walinfo.1.rst
```diff
@@ -2,21 +2,26 @@
 pgmoneta-walinfo
 =====================
 
-----------------------------------------------------------------------
-Command line utility to read and display Write-Ahead Log (WAL) files
-----------------------------------------------------------------------
+Command line utility to read and display Write-Ahead Log (WAL) files and TAR archives
```
jesperpedersen: Remove "and TAR archives"

src/include/achv.h
```diff
@@ -53,13 +53,14 @@ void
...
- * @param file_path The tar file path
+ * Extract an archive file to a given directory
+ * File type is detected internally via pgmoneta_get_file_type
+ * @param file_path The archive file path
  * @param destination The destination to extract to
  * @return 0 upon success, otherwise 1
  */
 int
-pgmoneta_extract_tar_file(char* file_path, char* destination);
+pgmoneta_extract_file(char* file_path, char* destination);
```
jesperpedersen: Its a `utils.h` function

src/libpgmoneta/archive.c
```diff
@@ -276,39 +277,51 @@ pgmoneta_archive(SSL* ssl, int client_fd, int server, uint8_t compression, uint8
...
+      pgmoneta_log_error("pgmoneta_extract_file: file is not a TAR archive");
+      return 1;
+   }
 
    a = archive_read_new();
    archive_read_support_format_tar(a);
 
-   if (config->compression_type == COMPRESSION_SERVER_GZIP)
+   /* Decompress based on file_type bitmask */
+   if (file_type & PGMONETA_FILE_TYPE_GZIP)
```
jesperpedersen: You have to check the extension too - and there could be multiple - test with `wal.tar.zstd.aes`

src/libpgmoneta/utils.c
```diff
@@ -34,8 +34,11 @@
 #include <utils.h>
 #include <info.h>
 #include <shmem.h>
+#include <deque.h>
```
jesperpedersen: Sorting

src/libpgmoneta/utils.c
```diff
@@ -34,8 +34,11 @@
...
 
 /* system */
+#include <archive.h>
```
jesperpedersen, line 41: Remove

src/libpgmoneta/utils.c
```diff
@@ -1997,23 +2002,59 @@ pgmoneta_get_files(char* base, struct deque** files)
...
    while ((entry = readdir(dir)) != NULL)
    {
-      if (entry->d_type == DT_REG)
+      if (strcmp(entry->d_name, ".") == 0 || strcmp(entry->d_name, "..") == 0)
       {
-         if (pgmoneta_deque_add(array, entry->d_name, (uintptr_t)entry->d_name, ValueString))
+         continue;
+      }
+
+      snprintf(full_path, sizeof(full_path), "%s/%s", base, entry->d_name);
```
jesperpedersen: Use pgmoneta_snprintf

src/libpgmoneta/utils.c
```diff
@@ -1997,23 +2002,59 @@ pgmoneta_get_files(char* base, struct deque** files)
...
+      else if (entry->d_type == DT_REG)
+      {
+         uint32_t file_type = pgmoneta_get_file_type(full_path);
+
+         if (file_type_mask == PGMONETA_FILE_TYPE_ALL || (file_type & file_type_mask))
+         {
+            if (recursive)
+            {
+               /* Recursive: return full path with strdup (caller must free) */
+               char* path_copy = strdup(full_path);
```
jesperpedersen: Use pgmoneta_append

src/libpgmoneta/utils.c
```diff
@@ -1997,23 +2002,59 @@ pgmoneta_get_files(char* base, struct deque** files)
...
 
-   // sort the deque
+   closedir(dir);
+   dir = NULL;
+
    pgmoneta_deque_sort(array);
 
    *files = array;
 
-   closedir(dir);
```
jesperpedersen, line 2014: Why the switch ?

src/libpgmoneta/utils.c
```diff
@@ -4400,6 +4441,105 @@ pgmoneta_is_compressed(char* file_path)
...
+   char* basename_copy = NULL;
+   char* current = NULL;
+   char* dot = NULL;
+
+   if (file_path == NULL)
+   {
+      return type;
+   }
+
+   basename_copy = strdup(basename(file_path));
```
jesperpedersen: Use pgmoneta_append

jesperpedersen (changes requested):

jesperpedersen:
@sksingh2005 I'll stop here... you need to review your own patch based on the API and where functions belong

sksingh2005:
@jesperpedersen sir I have replaced `pgmoneta_extract_tar_file` with `pgmoneta_extract_file` and moved the public API to `utils.h` (removed from `achv.h`).

- extraction is now shared across modules (`walinfo`, `walfilter`, archive receive)
- the replacement supports layered extensions and extraction in one path (`.tar`, compressed TAR variants, and optional `.aes`);
- this removes duplicate extractor paths and keeps one consistent implementation.

Is this approach good sir ?

doc/man/pgmoneta-walinfo.1.rst
```diff
@@ -2,21 +2,26 @@
...
 ===========
 
-pgmoneta-walinfo is a command line utility to read and display information about PostgreSQL Write-Ahead Log (WAL) files. It provides details of the WAL file in either raw or JSON format.
+pgmoneta-walinfo is a command line utility to read and display information about PostgreSQL Write-Ahead Log (WAL) files. It supports individual WAL files, directories containing WAL files, TAR archives (optionally compressed with gzip, lz4, zstd, or bz2), and encrypted files (.aes). It provides details of the WAL file in either raw or JSON format.
+
+In addition to standard WAL files, pgmoneta-walinfo also supports:
+
+- Encrypted WAL files (**.aes**)
+- Compressed WAL files: **zstd**, **gz**, **lz4**, and **bz2**
+- TAR archives containing WAL files (**.tar**)
```
jesperpedersen, line 23: Either all with a `.` or without - not a mix

doc/manual/en/17-wal-tools.md
jesperpedersen: Same

src/libpgmoneta/archive.c
```diff
@@ -983,3 +901,28 @@ is_server_side_compression(void)
           config->compression_type == COMPRESSION_SERVER_LZ4 ||
           config->compression_type == COMPRESSION_SERVER_ZSTD;
 }
+
+static const char*
+archive_file_extension(void)
```
jesperpedersen: And encryption ?

src/walfilter.c
```diff
@@ -684,8 +700,77 @@ main(int argc, char* argv[])
...
+
+      /* Handle TAR files - extract and process WAL files inside */
+      if (file_type & PGMONETA_FILE_TYPE_TAR)
+      {
+         char tar_temp_template[MAX_PATH];
+         char* tar_temp_dir = NULL;
+         struct deque* tar_wal_files = NULL;
+         struct deque_iterator* tar_iter = NULL;
+
+         snprintf(tar_temp_template, sizeof(tar_temp_template), "/tmp/pgmoneta_walfilter_XXXXXX");
```
jesperpedersen: pgmoneta_snprintf

src/walfilter.c
```diff
@@ -684,8 +700,77 @@ main(int argc, char* argv[])
...
+            {
+               char* tar_wal_path = (char*)tar_iter->value->data;
+
+               if (pgmoneta_read_walfile(-1, tar_wal_path, &wf))
+               {
+                  pgmoneta_log_error("Failed to read WAL file from TAR: %s", tar_wal_path);
+                  continue;
+               }
+
+               /* Reallocate walfiles array if needed */
```
jesperpedersen: What is this for ?

src/walinfo.c
```diff
@@ -2152,10 +2154,11 @@ usage(void)
 {
    printf("pgmoneta-walinfo %s\n", VERSION);
    printf("  Command line utility to read and display Write-Ahead Log (WAL) files\n");
+   printf("  Supports individual WAL files, directories, and TAR archives (optionally compressed)\n");
```
jesperpedersen: Remove

test/testcases/test_utils.c
```diff
@@ -1315,6 +1318,208 @@ MCTF_TEST(test_utils_enc_comp)
...
+   // Test basic WAL file
+   type = pgmoneta_get_file_type("000000010000000000000001");
+   MCTF_ASSERT((type & PGMONETA_FILE_TYPE_WAL), cleanup, "WAL file type detection failed");
+   MCTF_ASSERT(!(type & PGMONETA_FILE_TYPE_COMPRESSED), cleanup, "WAL file should not be compressed");
+   MCTF_ASSERT(!(type & PGMONETA_FILE_TYPE_ENCRYPTED), cleanup, "WAL file should not be encrypted");
+
+   // Test compressed WAL file .gz
+   type = pgmoneta_get_file_type("000000010000000000000001.gz");
+   MCTF_ASSERT((type & PGMONETA_FILE_TYPE_WAL), cleanup, "Compressed WAL detection failed .gz");
+   MCTF_ASSERT((type & PGMONETA_FILE_TYPE_COMPRESSED), cleanup, "Compression detection failed .gz");
```
jesperpedersen, line 1334: Add _GZIP assert as well ... same for the rest

jesperpedersen: And, remember !_ZSTD and so on

src/walfilter.c
sksingh2005: Sir It will handle capacity growth when TAR extraction yields more WAL files than the initial top-level file count.

src/libpgmoneta/archive.c
sksingh2005: Sir `archive_file_extension()` is intentionally compression-only.

It is used while receiving PostgreSQL `BASE_BACKUP` archive stream files, where the stream can be compressed (`.tar.gz/.tar.zstd/.tar.lz4`) but is not encrypted at that stage.  
Encryption is handled in the workflow/output path (we already append `.aes` for final archive filename when `config->encryption != ENCRYPTION_NONE`).

If we append `.aes` here, we would mislabel plaintext received files as encrypted and `pgmoneta_extract_file()` would try to decrypt non-encrypted data.

src/walfilter.c
jesperpedersen: Yes, but the iterator doesn't change - you need a better comment for sure

src/libpgmoneta/archive.c
jesperpedersen: So, it is `basebackup_archive_extension()` and depend on the `_SERVER_` flags. See, if that makes it more clear

sksingh2005: Yes sir. I’ll rename it to `basebackup_archive_extension()` to make scope explicit.

jesperpedersen:
@sksingh2005 Hmm, `pgmoneta-walinfo mywal.tar.gz` ate my file :-1:

jesperpedersen:
@sksingh2005 Ok, I could recreate it - you need to extract it, and only work on those files

sksingh2005:
Sir I  suggest we treat input archives as immutable in `walinfo`/`walfilter`:  
stage the archive into a temporary workspace first, then run extraction/decompression/decryption on that staged copy only.
- current extract/decompress/decrypt flow can remove its input as part of layer processing
- CLI users expect `mywal.tar.gz` to remain unchanged after inspection
- this keeps existing extraction logic intact while making the behavior safe and predictable

If this direction looks right, I can also add a small regression test that verifies the original archive still exists after execution.

jesperpedersen:
@sksingh2005 Correct, you may need to check that the temporary space have space for the file, and the extraction before you begin

jesperpedersen:
@sksingh2005 Maybe you can use the `size` field in the TAR header to calculate what space you need

jesperpedersen:
@sksingh2005 For `pgmoneta-walinfo -I mywal.tar.gz` I get

```
Error: mywal.tar.gz is not a valid WAL file
```

It should load and open the first WAL file

saraanbih:
Hi @sksingh2005,

I reviewed your code and found a critical issue in `pgmoneta_get_file_type`, u are checking for `.gz` before `.tar.gz`, this is incorrect because `.tar.gz` files also end with `.gz`.

Please adjust the order of the checks accordingly.

I wish u understand my point.

saraanbih:
Another note @sksingh2005 ;)

Don't forget to check interactive mode, it uses `pgmoneta_validate_wal_file` , and this method doesn't check `.tar` suffix

Does `pgmoneta_validate_wal_file` & `pgmoneta_get_file_type` have the same goal? to be concise `pgmoneta_validate_wal_file` is considered a part of `pgmoneta_get_file_type`, so we can combine those 2 functions and make one `pgmoneta_get_file_type`?

sksingh2005:
Hi @saraanbih !!

> Hi @sksingh2005,
> 
> I reviewed your code and found a critical issue in `pgmoneta_get_file_type`, u are checking for `.gz` before `.tar.gz`, this is incorrect because `.tar.gz` files also end with `.gz`.
> 
> Please adjust the order of the checks accordingly.
> 
> I wish u understand my point.

I re-checked `pgmoneta_get_file_type()`: the detection is layered intentionally.
For `.tar.gz`, `.gz` is stripped first, then `.tar` is detected on the remaining basename.
So `.tar.gz` still gets TAR + compression flags correctly (and this is covered by tests).

sksingh2005:
> Another note @sksingh2005 ;)
> 
> Don't forget to check interactive mode, it uses `pgmoneta_validate_wal_file` , and this method doesn't check `.tar` suffix
> 
> Does `pgmoneta_validate_wal_file` & `pgmoneta_get_file_type` have the same goal? to be concise `pgmoneta_validate_wal_file` is considered a part of `pgmoneta_get_file_type`, so we can combine those 2 functions and make one `pgmoneta_get_file_type`?

Thanks, good point.

`pgmoneta_get_file_type()` and `pgmoneta_validate_wal_filename()` serve different purposes:
- `pgmoneta_get_file_type()` is a generic classifier (WAL/compressed/encrypted/TAR/partial bitmask)
- `pgmoneta_validate_wal_filename()` is strict WAL-name validation

So I kept both APIs separate, but I updated the flow so TAR inputs are detected first and bypass WAL-only validation (including interactive mode), then we extract TAR and open the first WAL file from it.

If you prefer, I can do a follow-up cleanup where WAL validation internally reuses the file-type parsing path to reduce duplication.
@jesperpedersen cc

saraanbih:
@sksingh2005 
>I re-checked pgmoneta_get_file_type(): the detection is layered intentionally.
For .tar.gz, .gz is stripped first, then .tar is detected on the remaining basename.
So .tar.gz still gets TAR + compression flags correctly (and this is covered by tests).

Is it work with u?

The problem was with Mr. jesper, i have solved it with fix the order

saraanbih:
> Thanks, good point.
>
> pgmoneta_get_file_type() and pgmoneta_validate_wal_filename() serve different purposes:
>
>    pgmoneta_get_file_type() is a generic classifier (WAL/compressed/encrypted/TAR/partial bitmask)
>  pgmoneta_validate_wal_filename() is strict WAL-name validation
>
> So I kept both APIs separate, but I updated the flow so TAR inputs are detected first and bypass WAL-only validation 
> (including interactive mode), then we extract TAR and open the first WAL file from it.
>
> If you prefer, I can do a follow-up cleanup where WAL validation internally reuses the file-type parsing path to reduce duplication.

Yes, one is generic and the other is specific

as Mr. Jesper want, so do 

but for now interactive mode doesn't work correctly with .tar

sksingh2005:
> We should not have 100s of functions that does the same thing - we should think along the lines of
> 
> ```
>   /* Get WAL files and standalone files; with possible WAL files inside TAR archives */
>   pgmoneta_get_files(FILE_TYPE_WAL | FILE_TYPE_TAR | FILE_TYPE_AES | ..., directory, true, &files);
> 
>   /* Extract the files that need it ... */
>  ...
> ```

Sir I suggest we handle interactive TAR input the same way as the non-interactive flow:
- detect TAR input first
- extract/stage it in temp
- discover WAL files from extracted content
- open the first WAL file in interactive mode

This would avoid running WAL-only filename validation on `mywal.tar.gz` and should resolve the `not a valid WAL file` error for `-I`.

How about this sir ??

jesperpedersen:
@sksingh2005 Yes, `pgmoneta-walinfo -I wal.tar.gz` should open the first WAL file detected

sksingh2005:
> @sksingh2005
> 
> > I re-checked pgmoneta_get_file_type(): the detection is layered intentionally.
> > For .tar.gz, .gz is stripped first, then .tar is detected on the remaining basename.
> > So .tar.gz still gets TAR + compression flags correctly (and this is covered by tests).
> 
> Is it work with u?
> 
> The problem was with Mr. jesper, i have solved it with fix the order

Thanks for checking.

Yes, I understand your fix-order point.  
In the current implementation, `.tar.gz` is still detected correctly because parsing is layered: compression suffix is stripped first, then `.tar` is checked on the remaining basename.

I verified behavior with tests (`backup.tar.gz` asserts TAR + COMPRESSED + GZIP).  
So functionally it works in current code as well.

jesperpedersen:
@sksingh2005 Does it work for `.tar.zstd.aes` ?

jesperpedersen:
@sksingh2005 Looks like it

jesperpedersen:
@sksingh2005 @saraanbih PTAL me once you both agree

saraanbih:
@sksingh2005 Great !
LGTM

I think now we are both agree ;)

sksingh2005:
Yes !!

jesperpedersen:
Merged.

Issue #918, opened 2026-02-09
HTTP reader truncates large GET responses over TLS


Amr-Shams:
## Description

The HTTP response reader in `src/libpgmoneta/http.c` (line 557) terminates prematurely when reading large responses over TLS connections:

```c
if (strstr(buffer, "\r\n0\r\n\r\n") || bytes_read < (ssize_t)(sizeof(buffer) - 1))
{
   break;
}
```

The condition `bytes_read < sizeof(buffer) - 1` breaks the read loop whenever an `SSL_read` returns less than 8191 bytes. With TLS, data arrives in record-sized chunks that are typically smaller than the 8192-byte buffer, so the loop often exits after the first read — capturing only a partial response body.

## Impact

This affects any S3 GET operation that returns a large response (e.g., `ListObjectsV2` with many objects). The XML response is truncated, resulting in incomplete data. PUT operations are unaffected since their responses are small (just headers + status).

## Steps to Reproduce

1. Configure pgmoneta with S3 storage (e.g., Cloudflare R2 or AWS)
2. Have a bucket with many objects under a prefix
3. Issue a ListObjectsV2 request via the HTTP library
4. Only a small fraction of objects are returned (e.g., 4 out of hundreds)

## Suggested Fix

The read loop should use `Content-Length` or properly handle chunked `Transfer-Encoding` to determine when the full response has been received, rather than relying on buffer fill level.

jesperpedersen:
@bassamadnan Cc:

Amr-Shams:
> [@bassamadnan](https://github.com/bassamadnan) Cc:

after i review the code of the http, the current flow needs to be updated heavily, 

**Response termination is decided before headers are parsed**

The current flow reads the entire HTTP response into a buffer and only then parses headers. However, correct response termination depends on header values (Content-Length, Transfer-Encoding), which must be known before reading the body. This makes it impossible to reliably determine when the full response has been received.

**useless headers on the response**
as it is extracted only after the full response has already been read. As a result, it cannot be used to control or validate the read loop, and truncated or oversized responses cannot be detected at read time.

**Chunked transfer decoding is line-based and incomplete**

The current parsing logic attempts to identify chunk boundaries using line tokenization and numeric checks. Proper chunked decoding is stateful and byte-oriented; chunk sizes may include extensions, and chunk data itself may contain arbitrary bytes, including newline characters.

**HTTP body handling is not binary-safe**

[...]

PR #923, opened 2026-02-10
[#918]http: parse headers separately and handle content-length/chunke…

Amr-Shams:
…d responses

### DIFF
--- a/src/libpgmoneta/http.c
+++ b/src/libpgmoneta/http.c
@@ -31,6 +31,7 @@
 #include <http.h>
 #include <logging.h>
 #include <network.h>
+#include <deque.h>
 #include <security.h>
 #include <utils.h>
 
@@ -40,8 +41,9 @@
 #include <unistd.h>
 #include <openssl/err.h>
 
-static int http_read_response(SSL* ssl, int socket, char** response);
-static int http_parse_response(char* response, struct http_response* http_response);
+static int http_parse_header(char** header, struct http_response* http_response);
+static int http_read_response_body(SSL* ssl, int socket, struct http_response* http_response);
+static int http_read_response_header(SSL* ssl, int socket, char** header_text, struct http_response* http_response);
 static int http_build_request(struct http* connection, struct http_request* request, char** full_request, size_t* full_request_size);
 static char* http_method_to_string(int method);
 
@@ -362,7 +364,14 @@ pgmoneta_http_get_response_header(struct http_response* response, char* name)
       return NULL;
    }
 
-   return (char*)pgmoneta_deque_get(response->payload.headers, name);
+   char* key = strdup(name);
+   if (key == NULL)
+   {
+      return NULL;
+   }
+   char* value = (char*)pgmoneta_deque_get(response->payload.headers, key);
+   free(key);
+   return value;
 }
 
 int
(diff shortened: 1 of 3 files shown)


jesperpedersen:
@Amr-Shams CI

Amr-Shams:
there are many tests that could be added to test the different responses
```
Content-Length 
Truncated-Chunk 
```
but both would need a very different pr(needs to setup the server in the correct way with a correct chunks/content response) 
worth noting this as well this, the new fallback (read EOF) is not safe should be guarded with some timeout or any safeguards here, but as far as i know this won't happen in any of the requests of s3 protocol

jesperpedersen:
@shashidhar-bm How far are you from to porting the test framework here ?

shashidhar-bm:
@jesperpedersen , 
just need to fix the ci error, and we can get started, using mctf!

Amr-Shams:
> there are many tests that could be added to test the different responses
> 
> ```
> Content-Length 
> Truncated-Chunk 
> ```
> 
> but both would need a very different pr(needs to setup the server in the correct way with a correct chunks/content response) worth noting this as well this, the new fallback (read EOF) is not safe should be guarded with some timeout or any safeguards here, but as far as i know this won't happen in any of the requests of s3 protocol

@jesperpedersen i have added some unit test for them and updated the code accordingly to cover the test cases

jesperpedersen:
@Amr-Shams @bassamadnan need to sign-off on this

src/libpgmoneta/http.c
```diff
@@ -27,6 +27,7 @@
  */
 
 /* pgmoneta */
+#include "deque.h"
```
bassamadnan: this is in the wrong place, PTAL at the other files.

bassamadnan: pgmoneta comes first, rest alphabetically

src/libpgmoneta/http.c
```diff
@@ -509,164 +524,408 @@ pgmoneta_http_destroy(struct http* connection)
...
+      {
+         free(*header_text);
+         goto error;
+      }
+
+      end = strstr(*header_text, "\r\n\r\n");
+   }
+
+   // store the rest as body/data of the http request
+   size_t header_len = (end - *header_text) + 4;
```
bassamadnan, line 603: abit of comment on the +4 would be good

bassamadnan: becauseu it later goes on to the extra variable too

src/libpgmoneta/http.c
```diff
@@ -509,164 +524,408 @@ pgmoneta_http_destroy(struct http* connection)
...
+
+   // handle content length
+   if (cl_str)
+   {
+      size_t content_length = strtoul(cl_str, NULL, 10);
+      return http_read_content_length_body(ssl, socket, http_response, content_length);
+   }
+
+   // no content_length no chucnked error for now
+   // there is this case of streaming but not for now
```
bassamadnan: ?

jesperpedersen:
Rebased, and merged.

jesperpedersen:
@Amr-Shams Please, port to pgexporter before you continue

PR #940, opened 2026-02-20
[#939] Fix missing SIGCHLD handler causing zombie process leak

### LINKED ISSUE #939: Missing SIGCHLD handler causes zombie process leak on every management and WAL command
The daemon makes use of fork() but never reaps its terminated child processes, resulting in zombie processes accumulating over time

fork() is used heavily throughout src/main.c for commands like backup, restore, status, details, etc

fork() is also used in src/libpgmoneta/wal.c for compressing and encrypting WAL segments

as SIGCHLD is missing from the event loop, every single one of those fork() calls leaves behind a zombie process.

fix:

register SIGCHLD in the libev setup in main.c.

port the sigchld_cb logic from pgagroal to catch the signal.

use that to cleanly reap exited children from the kernel's process table.

trxvorr:
fixes #939

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -75,6 +75,7 @@
 #include <sys/stat.h>
 #include <sys/time.h>
 #include <sys/types.h>
+#include <sys/wait.h>
 
 #include <openssl/crypto.h>
 #ifdef HAVE_SYSTEMD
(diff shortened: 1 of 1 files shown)


jesperpedersen:
@trxvorr You need to fix the stop loop as well

jesperpedersen:
Rebased, and merged.

Issue #943, opened 2026-02-24
Extract: Create a better extraction layer


jesperpedersen:
We need to make a public API to extract a specific file type (.aes, .zstd, .tar, ...) and then abstractions to deal with a file that use multiple (`tar.zstd.aes`) methods.

Most of this is already in place, but we need to remove code duplication, and make the APIs more clear

sksingh2005:
@jesperpedersen sir can i pick this issue ?

jesperpedersen:
@sksingh2005 Go for it !

jesperpedersen:
@sksingh2005 Maybe the best way forward is a new `.h` file with the functions to deal with "both" ways... that use other `.h` files. Feel free to move code around.

Having restore.c call `pgmoneta_extract_backup_file` is def something that need to be looked at

sksingh2005:
Agreed. sir I’ll introduce a dedicated extraction module (`extraction.h/.c`) that provides a clean public API for:

1. Extracting a single layer by type (`.aes`, `.zstd`, `.lz4`, `.bz2`, `.tar`)
2. Extracting chained layers (`tar.zstd.aes`) through one orchestration function

The module will wrap existing AES/compression/tar implementations (from their current headers) so we avoid duplication and keep one extraction flow.

sksingh2005:
I’ll also review `restore.c -> pgmoneta_extract_backup_file` and :
- move that path to the new extraction API, or
- make `pgmoneta_extract_backup_file` a thin wrapper over the new API

so restore no longer owns extraction-specific logic directly.
Will this be right approach sir ?

jesperpedersen:
@sksingh2005 And, the other way... Start by looking at `pgmoneta_extract_file(char* file_path, char* destination);`.

The other way - for this function - we would need a bitmask (_TAR | _ZSTD | _AES) -- follow `#define PGMONETA_FILE_TYPE_XYZ`

jesperpedersen:
@sksingh2005 `backup.c` and `restore.c` should both make use of this new layer in an easy way

Issue #764, opened 2025-09-21
Storage layer refactor


Jubilee101:
Decouple storage layer with logics, preparing for S3 and SSH support.

Amr-Shams:
@Jubilee101 i have some idea regarding the gradual refactoring, specially after the delete force(which would be easy in case of local) but other engines will be much spoil,  as the current impl, doesn't have a solid interface for all backend (which will leave hundreds of repeated code)
-  uniform Ops: Backup works everywhere, but restore/delete/list only work locally(accordingly retention policy,cleanups are only on the local)
- No interface for all of them: Each backend (se_*.c) implements ops independently
- Retention Policy very hard to add to the current backend(add to it the repeating almost of everything everytime a we need to do one of the func (view the diff between all the servers for example)
- HTTP needs a true update, the http only support empty body reqs, and only one type hard coded body(needs to support  xml as well)

Jubilee101:
@Amr-Shams Sorry was distracted last month by other matters. These are all valid pain points, some are the reason we created this issue. I'll think about them, thanks.

Jubilee101:
The basic infra of streamer and vfile is in. We will open other issues to tackle the remaining refactoring work.

PR #953, opened 2026-02-25
[#943] refactor:add bitmask-based extraction API and unify layered extraction flow

sksingh2005:
Fixes: [#943]

### DIFF
--- a/src/include/extraction.h
+++ b/src/include/extraction.h
@@ -1,0 +1,107 @@
+/*
+ * Copyright (C) 2026 The pgmoneta community
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
+ * are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this list
+ * of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice, this
+ * list of conditions and the following disclaimer in the documentation and/or other
+ * materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its contributors may
+ * be used to endorse or promote products derived from this software without specific
+ * prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+ * OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+ * THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT
+ * OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR
+ * TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef PGMONETA_EXTRACTION_H
+#define PGMONETA_EXTRACTION_H
+
+#include <stdbool.h>
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * Remove extraction-related suffixes from a file path.
+ * For example, "001.tar.zstd.aes" becomes "001" when the bitmask includes TAR, ZSTD, and ENCRYPTED.
+ *
+ * When type is 0, the bitmask is detected automatically from the file path.
+ *
+ * @param file_path The source file path
+ * @param type The file type bitmask (PGMONETA_FILE_TYPE_*), or 0 for auto-detect
+ * @param base_name Resulting base path without extraction suffixes
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_extraction_strip_suffix(char* file_path, uint32_t type, char** base_name);
+
+/**
+ * Build the compound suffix string for a given compression and encryption configuration.
+ * For example, with ZSTD compression and AES encryption, produces ".zstd.aes".
+ *
+ * @param compression The compression type (COMPRESSION_* constant)
+ * @param encryption The encryption type (ENCRYPTION_* constant)
+ * @param suffix The resulting suffix string (caller must free)
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_extraction_get_suffix(int compression, int encryption, char** suffix);
+
+/**
+ * Extract a file using the streamer for in-memory decryption and decompression.
+ * Decryption and decompression are handled in one pass via streamer(RESTORE).
+ * Assumes single compression + single encryption.
+ *
+ * When type is 0, the bitmask is detected automatically from the file path.
+ * When copy is true, the source is streamed (decrypted+decompressed) to the destination path,
+ * and the destination pointer is updated to the final extracted file path.
+ * When copy is false, the source is streamed to a tar, then extracted to the destination directory.
+ *
+ * @param file_path The source file path
+ * @param type The file type bitmask (PGMONETA_FILE_TYPE_*), or 0 for auto-detect
+ * @param copy If true, extract to destination path. If false, extract archive to destination directory.
+ * @param destination When copy is true, points to the destination path (updated to final extracted path).
+ *                    When copy is false, points to the output directory for extraction (not modified).
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_extract_file(char* file_path, uint32_t type, bool copy, char** destination);
+
+/**
+ * Extract a file from a backup to a target location.
+ * Wrapper around pgmoneta_extract_file for backup-relative paths.
+ *
+ * Builds the source path from backup data directory + relative_file_path,
+ * and the destination from target_directory (or workspace if NULL) + relative_file_path.
+ * Then calls pgmoneta_extract_file(from, 0, true, &to).
+ *
+ * @param server The server index
+ * @param label The backup label
+ * @param relative_file_path The file path relative to the backup data directory
+ * @param target_directory The target directory (NULL to use workspace/label/)
+ * @param target_file [out] The final extracted file path (caller must free)
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_extract_backup_file(int server, char* label, char* relative_file_path, char* target_directory, char** target_file);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif
--- a/src/include/tar.h
+++ b/src/include/tar.h
@@ -1,0 +1,58 @@
+/*
+ * Copyright (C) 2026 The pgmoneta community
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
+ * are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this list
+ * of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice, this
+ * list of conditions and the following disclaimer in the documentation and/or other
+ * materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its contributors may
+ * be used to endorse or promote products derived from this software without specific
+ * prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+ * OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+ * THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT
+ * OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR
+ * TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef PGMONETA_TAR_H
+#define PGMONETA_TAR_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * Create a tar archive of the given directory
+ * @param src The source directory
+ * @param dst The destination tar file path
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_tar(char* src, char* dst);
+
+/**
+ * Extract a tar archive to the given directory
+ * @param src The path to the tar file
+ * @param dst The destination directory
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_untar(char* src, char* dst);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif
--- a/src/include/utils.h
+++ b/src/include/utils.h
@@ -86,6 +86,10 @@ extern "C" {
 #define PGMONETA_FILE_TYPE_TAR        0x0080 /* TAR archive (.tar) */
 #define PGMONETA_FILE_TYPE_PARTIAL    0x0100 /* Partial file (.partial) */
 #define PGMONETA_FILE_TYPE_ALL        0xFFFF /* Match all file types */
+#define PGMONETA_FILE_TYPE_COMPRESSION_MASK \
+   (PGMONETA_FILE_TYPE_COMPRESSED | PGMONETA_FILE_TYPE_GZIP | PGMONETA_FILE_TYPE_LZ4 | PGMONETA_FILE_TYPE_ZSTD | PGMONETA_FILE_TYPE_BZ2)
+#define PGMONETA_FILE_TYPE_EXTRACTION_MASK \
+   (PGMONETA_FILE_TYPE_ENCRYPTED | PGMONETA_FILE_TYPE_TAR | PGMONETA_FILE_TYPE_COMPRESSION_MASK)
 
 /** @struct signal_info
  * Defines the signal structure
@@ -739,17 +743,6 @@ int
 pgmoneta_get_files(uint32_t file_type_mask, char* base, bool recursive, struct deque** files);
 
 /**
- * Extract an archive file to a given directory
- * File type is detected internally via pgmoneta_get_file_type
- * Handles layered formats (e.g., file.tar.zstd.aes)
- * @param file_path The archive file path
- * @param destination The destination to extract to
- * @return 0 upon success, otherwise 1
- */
-int
-pgmoneta_extract_file(char* file_path, char* destination);
-
-/**
  * Get WAL files
  * @param base The base directory
  * @param files The deque of files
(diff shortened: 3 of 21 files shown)


jesperpedersen:
@sksingh2005 @Jubilee101 will lead the review on this one

Jubilee101:
@sksingh2005 Can you check if #764 helps with this? It's WIP, but about to be done. Can you take a look and let me know your thoughts? I can use your opnions here, and you may need to change the design here a little bit so that we may leverage the streamer to do the extraction. 

With streamer you can decompress and decrypt all in mem in a streaming style, which will generate end result directly in the chunks of data, without generating any middle state files. And by implementing different types of vfiles and adding them as destinations, You can determine what to do with these chunks of data, you can write it to disk (vfile_local), or send it to S3 (vfile_s3, not yet implemented), or just parse it in place (maybe vfile_walinfo). You will find a usage example in test_streamer.c.

I did it intially so that we can support S3 more easily. But I think it might help with extraction. It will save A LOT of disk IO. Currently what I'm about to do next is to use the streamer during base backup. libarchive also supports untar in a streaming style, we can feed the untarred data to streamer directly, which will compress and encrypt the chunks and write them to disk directly. This may even be wrapped as a vfile_tarball so that we can chain things together. 

One thing I'm currently finding incompatible is that the scope of pgmoneta_extract_file is a little too big, it's dealing with files that may have gone through multiple compression and encryption, which is rarely the case we care about. Streamer is built on the assumption that a file can be compressed and encrypted exactly once respectively. 

Let me know what you think. I think we can perhaps work together on this. Thanks.

sksingh2005:
Hi @Jubilee101 sir, 
I think extraction and streamer are complementary: extraction handles the reverse (extracting existing on-disk files), streamer handles the forward (streaming data through compress+encrypt as it arrives). They don't conflict.

I agree the multi-layer scope can be simplified to single-compress + single-encrypt, matching the streamer's assumption. Would you prefer I simplify that in this PR itself, or should we handle it as a follow-up once the streamer is closer to merge?
Happy to coordinate on this.

jesperpedersen:
@sksingh2005 https://github.com/Jubilee101/pgmoneta/tree/764

src/include/compression.h
```diff
@@ -49,4 +49,15 @@ typedef int (*compression_func)(char*, char*);
 int
 pgmoneta_decompress(char* from, char* to);
 
+/**
```
jesperpedersen: We should only have a single function

src/include/extraction.h
```diff
@@ -0,0 +1,112 @@
...
+
+#ifndef PGMONETA_EXTRACTION_H
+#define PGMONETA_EXTRACTION_H
+
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
```
jesperpedersen, line 38: This is too complex.

We need a way to obtain the bitmask for a file. Then we can use the bitmask for the rest of the functions

src/include/extraction.h
```diff
@@ -0,0 +1,112 @@
...
+ * Extract an archive file to a given directory.
+ * File type is detected internally via pgmoneta_get_file_type.
+ * Handles layered formats (for example file.tar.zstd.aes).
+ *
+ * @param file_path The archive file path
+ * @param destination The destination to extract to
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_extract_file(char* file_path, char* destination);
```
jesperpedersen: This is the main function - it need `uint32_t type` and `bool copy`

jesperpedersen: The problem is `char* destination` - maybe it is always `char**`

jesperpedersen:
@sksingh2005  I think the first step is to unify the naming / type of files - all the `suffix = pgappend(suffix, ".gz")` need to be in a central place. And, the same in the other direction - going from `001.tar.zstd.aes` to `001`

jesperpedersen:
@sksingh2005 Start with a `..._get_suffix(char** suffix)` function and see how much code you can remove

jesperpedersen:
@sksingh2005 The important part is the bitmask - so follow that around the code

jesperpedersen:
@sksingh2005 Maybe we need a `tar.h` with 

```
int
pgmoneta_tar_directory(char* src, char* dst, char* destination);
```

jesperpedersen:
@sksingh2005 We can detect the bitmask within `_extract_file`

sksingh2005:
@jesperpedersen sir  `pgmoneta_tar_directory()` currently lives in  achv.h. Should I move it to a separate tar.h ?

jesperpedersen:
@sksingh2005 We need it as simple as possible, so think of `tar.h` as

```
pgmoneta_tar(...
pgmoneta_untar(...
```

src/include/extraction.h
```diff
@@ -0,0 +1,87 @@
...
+ * For example, "001.tar.zstd.aes" becomes "001" when the bitmask includes TAR, ZSTD, and ENCRYPTED.
+ *
+ * When type is 0, the bitmask is detected automatically from the file path.
+ *
+ * @param file_path The source file path
+ * @param type The file type bitmask (PGMONETA_FILE_TYPE_*), or 0 for auto-detect
+ * @param base_name Resulting base path without extraction suffixes
+ * @return 0 upon success, otherwise 1
+ */
+int
```
jesperpedersen, line 50: utils.h function

src/include/extraction.h
```diff
@@ -0,0 +1,87 @@
...
+ * Build the compound suffix string for a given compression and encryption configuration.
+ * For example, with ZSTD compression and AES encryption, produces ".zstd.aes".
+ *
+ * @param compression The compression type (COMPRESSION_* constant)
+ * @param encryption The encryption type (ENCRYPTION_* constant)
+ * @param suffix The resulting suffix string (caller must free)
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_get_suffix(int compression, int encryption, char** suffix);
```
jesperpedersen: utils.h function

src/include/tar.h
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Create a tar archive of the given directory
+ * @param src The source directory
+ * @param dst The destination tar file path
+ * @param destination The destination name inside the archive
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_tar(char* src, char* dst, char* destination);
```
jesperpedersen: Don't we just need `src` and `dst` ?

src/include/tar.h
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Extract a tar archive to the given directory
+ * @param archive_path The path to the tar file
+ * @param destination The destination directory
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_untar(char* archive_path, char* destination);
```
jesperpedersen: `src` and `dst`

jesperpedersen:
@sksingh2005 Just know we will pull in 764 first, and then use your patch to refactor / remove / ... functions

src/include/tar.h
sksingh2005: `destination` was only used to override the archive root name. I will update so the root is derived from `basename(src)` to keep the interface minimal.

jesperpedersen: My point is the the layer above have to provide the correct `dst`

sksingh2005: Got it sir I’ll keep tar layer minimal and make the caller provide the final correct `dst` path/name.

src/include/utils.h
```diff
@@ -1279,6 +1259,32 @@ pgmoneta_is_compressed(char* file_path);
...
+ *
+ * When type is 0, the bitmask is detected automatically from the file path.
+ *
+ * @param file_path The source file path
+ * @param type The file type bitmask (PGMONETA_FILE_TYPE_*), or 0 for auto-detect
+ * @param base_name Resulting base path without extraction suffixes
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_strip_suffix(char* file_path, uint32_t type, char** base_name);
```
jesperpedersen: I would move this to extraction.h and call it `pgmoneta_extraction_strip_suffix`

src/include/utils.h
```diff
@@ -1279,6 +1259,32 @@ pgmoneta_is_compressed(char* file_path);
...
+ * Build the compound suffix string for a given compression and encryption configuration.
+ * For example, with ZSTD compression and AES encryption, produces ".zstd.aes".
+ *
+ * @param compression The compression type (COMPRESSION_* constant)
+ * @param encryption The encryption type (ENCRYPTION_* constant)
+ * @param suffix The resulting suffix string (caller must free)
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_get_suffix(int compression, int encryption, char** suffix);
```
jesperpedersen: I would move this to extraction.h and call it `pgmoneta_extraction_get_suffix`

src/include/extraction.h
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * When type is 0, the bitmask is detected automatically from the file path.
+ *
+ * @param file_path The source archive path
+ * @param destination The destination directory
+ * @param type The file type bitmask (PGMONETA_FILE_TYPE_*), or 0 for auto-detect
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_extract_archive(char* file_path, char* destination, uint32_t type);
```
jesperpedersen: Why is this back ?

src/include/extraction.h
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * @param server The server
+ * @param label The backup label
+ * @param relative_file_path Path relative to backup data directory
+ * @param target_directory Optional target root directory (NULL to use workspace)
+ * @param target_file The final extracted path (caller must free)
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_extract_backup_file(int server, char* label, char* relative_file_path, char* target_directory, char** target_file);
```
jesperpedersen, line 101: Likewise

src/include/extraction.h
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * @param file_path The source file path
+ * @param destination When copy is true, points to the destination path (updated to final extracted path).
+ *                    When copy is false, points to the output directory for extraction (not modified).
+ * @param type The file type bitmask (PGMONETA_FILE_TYPE_*), or 0 for auto-detect
+ * @param copy If true, copy the file before extracting
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_extract_file(char* file_path, char** destination, uint32_t type, bool copy);
```
jesperpedersen: `destination` is last

src/include/extraction.h
```diff
@@ -0,0 +1,129 @@
...
+ * When type is 0, the bitmask is detected automatically from the file path.
+ * The destination points to the target copy path and is updated to the final extracted path.
+ *
+ * @param file_path The source file path
+ * @param destination The target path (updated to final extracted path)
+ * @param type The file type bitmask (PGMONETA_FILE_TYPE_*), or 0 for auto-detect
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_extract_file_to_path(char* file_path, char** destination, uint32_t type);
```
jesperpedersen: Nuke this

src/include/extraction.h
sksingh2005: Ohh sorry  sir, it was from an earlier revision where I split `pgmoneta_extract_file` into separate functions.

Jubilee101:
Hey @sksingh2005 , sorry for the late reply.. As you may have noticed, #764 has completed and we merged the streamer. 

> extraction handles the reverse (extracting existing on-disk files), streamer handles the forward (streaming data through compress+encrypt as it arrives). They don't conflict.

Nah the streamer works in both directions, sorry for the confusion. 

> Would you prefer I simplify that in this PR itself, or should we handle it as a follow-up once the streamer is closer to merge?

I would prefer that we start using the streamer in the extraction APIs. An example you can refer to is `pgmoneta_extract_backup_tar_file` in archive.c. You can likely copy a lot of code from there. But I can't think of a good way to get rid of this function and just reuse your APIs -- for extracting backup tar file we do some other stuff such as checksum checking, and in the future the incremental file size calculation, and also files like backup_label or backup_manifest are exempt from compression/encryption, therefore a lot of special handling is happening for backup tar file there. I think for now we can tolerate this code duplication, but feel free to pitch in your ideas here. Cc @jesperpedersen .

> I agree the multi-layer scope can be simplified to single-compress + single-encrypt, matching the streamer's assumption. Would you prefer I simplify that in this PR itself, or should we handle it as a follow-up once the streamer is closer to merge?

I think by adopting the streamer the logic will naturally be reduce down to single compression + encryption. We just need to document the assumption.

I do realize that this requires some additional code changes and can put some of your work in vain, for that I'm deeply sorry. I do hope the streamer is easy enough to use, let me know if you think the API can use any improvements.

sksingh2005:
No problem sir I will take a look and get back to you !!

sksingh2005:
@Jubilee101 sir I'll adopt the streamer in the extraction APIs — the `pgmoneta_extract_backup_tar_file` reference is really helpful. I agree the code duplication with that function is tolerable given the backup-specific handling (checksums, exempt files). I'll add a note about the single-compress + single-encrypt assumption. Will push the updated version soon.

jesperpedersen:
We can look at `pgmoneta_extract_backup_tar_file` later

src/libpgmoneta/extraction.c
```diff
@@ -0,0 +1,606 @@
...
+      /* No encryption or compression — just copy */
+      if (pgmoneta_copy_file(file_path, *destination, NULL))
+      {
+         goto error;
+      }
+      return 0;
+   }
+
+   /* Step 1: Decrypt if needed (file-based, creates temp file) */
+   if (file_type & PGMONETA_FILE_TYPE_ENCRYPTED)
```
Jubilee101: Any reason why you can't decrypt and decompress using streamer in one go without creating a temp file?

src/libpgmoneta/restore.c
```diff
@@ -1694,7 +1698,7 @@ copy_backup_file(int server,
    // copy the full file from input dir to output dir
    // extract before copy
    snprintf(manifest_path, MAX_PATH_CONCAT, "%s%s", relative_dir, file_name);
-   if (pgmoneta_extract_backup_file(server, label, manifest_path, NULL, &extracted_file_path))
```
Jubilee101, line 1697: Maybe we should keep pgmoneta_extract_backup_file as a wrapper of pgmoneta_extract_file instead of having two identical static function defined in two modules. The goal here is to reuse code, it's fine that we have multiple extraction wrapper for different needs as long as internally they use the same APIs.

src/libpgmoneta/extraction.c
sksingh2005: Sir I initially tried to decrypt+decompress in one pass using the streamer, but hit `EVP_CipherFinal_ex` failures. 
I think to make one-pass work, the encryptor's decrypt path would need to read the salt from the first bytes of input. Should I make that change?

sksingh2005: I think sir the cause is `create_aes_encryptor`in `aes.c` derives key/iv with a zero salt, while `pgmoneta_encrypt_file` (`wf_encryption.c`) uses a random salt prepended to the ciphertext. Different salt = different key/iv = decryption fails.  Or I am missing something obvious sir ?

src/libpgmoneta/restore.c
jesperpedersen: @trxvorr Please, look at this

sksingh2005: Restored `pgmoneta_extract_backup_file` as a shared wrapper

src/libpgmoneta/extraction.c
Jubilee101: Inside aes_encryptor it uses derive_key_iv which is shared by pgmoneta_encrypt_file, and internally it takes in NULL salt. Can you point me as to where the salt is read and taken in? And how is your testing file encrypted? It shouldn't encounter the problem if it's also encrypted by encrypter. That said, something encrypted by pgmoneta_encrypt_file() should still work. 

If you are willing to try, replacing whatever encrypts your file with encryptor should resolve the problem. You don't even have to have the streamer/vfile overhead, simply write the encryptor output to file. But I agree that this is a separate issue and we shall look at it later.

sksingh2005: `create_aes_encryptor()` currently derives key/iv from a zero-filled salt:
- [`src/libpgmoneta/aes.c:1507`](https://github.com/pgmoneta/pgmoneta/blob/main/src/libpgmoneta/aes.c#L1507) (`unsigned char salt[PBKDF2_SALT_LENGTH];`)
- [`src/libpgmoneta/aes.c:1511`](https://github.com/pgmoneta/pgmoneta/blob/main/src/libpgmoneta/aes.c#L1511) (`memset(salt, 0, sizeof(salt));`)
- [`src/libpgmoneta/aes.c:1524`](https://github.com/pgmoneta/pgmoneta/blob/main/src/libpgmoneta/aes.c#L1524) (`derive_key_iv(master_key, salt, ae->key, ae->iv, mode)`)

`encrypt_file()` (used by `pgmoneta_encrypt_file`) uses per-file random salt and salt-prefixed file format:
- [`src/libpgmoneta/aes.c:1148`](https://github.com/pgmoneta/pgmoneta/blob/main/src/libpgmoneta/aes.c#L1148)-[`1156`](https://github.com/pgmoneta/pgmoneta/blob/main/src/libpgmoneta/aes.c#L1156) (random salt generation + derive with random salt)
- [`src/libpgmoneta/aes.c:1163`](https://github.com/pgmoneta/pgmoneta/blob/main/src/libpgmoneta/aes.c#L1163)-[`1177`](https://github.com/pgmoneta/pgmoneta/blob/main/src/libpgmoneta/aes.c#L1177) (decrypt path reads salt from file)
- [`src/libpgmoneta/aes.c:1207`](https://github.com/pgmoneta/pgmoneta/blob/main/src/libpgmoneta/aes.c#L1207)-[`1210`](https://github.com/pgmoneta/pgmoneta/blob/main/src/libpgmoneta/aes.c#L1210) (salt is prepended to output file)

So ig the issue is not `derive_key_iv` itself, but that streamer AES decrypt currently does not consume the salt-prefixed file framing used by `pgmoneta_encrypt_file`.

My test file is encrypted via `pgmoneta_encrypt_file` (same path used from `wf_encryption.c`), which explains the `EVP_CipherFinal_ex` failure in one-pass restore.

jesperpedersen: @trxvorr Cc:

trxvorr: @jesperpedersen @Jubilee101 
@sksingh2005 s analysis is correct. to properly enable streaming decrypt, aes_encryptor process would need to buffer the first PBKDF2_SALT_LENGTH bytes on the initial chunk, extract the salt, derive key/iv from it, and only then call EVP_CipherInit_ex  
the current code initializes the cipher immediately with a zero salt, which is why it fails on files encrypted by pgmoneta_encrypt_file

trxvorr: i agree with @Jubilee101 that the salt handling fix in the encryptor should be a separate issue from this pr

src/libpgmoneta/restore.c
trxvorr: looks good, it as a shared wrapper is the right approach

src/libpgmoneta/extraction.c
```diff
@@ -0,0 +1,669 @@
...
+   *decrypted_path = NULL;
+
+   if (file_type & PGMONETA_FILE_TYPE_ENCRYPTED)
+   {
+      if (pgmoneta_strip_extension(file_path, &result))
+      {
+         goto error;
+      }
+
+      if (pgmoneta_decrypt_file(file_path, result))
```
trxvorr: @sksingh2005 @jesperpedersen @Jubilee101 
the old pgmoneta_extract_backup_file used pgmoneta_copy_and_extract_file, which copied the source to the workspace before decrypting. 
this new path calls pgmoneta_decrypt_file directly on the backup file, which deletes the source after decryption (aes line 550)
```
 if (pgmoneta_exists(from))
   {
      pgmoneta_delete_file(from, NULL);
   }
   ```
 for the pgmoneta_extract_backup_file wrapper, the source file should be copied first to preserve the backup right??

File: src/libpgmoneta/info.c [hunk 17]
```diff
@@ -1624,65 +1625,6 @@ pgmoneta_incremental_rfile_initialize(int server, char* label, char* relative_di
...
-      to = pgmoneta_append(to, target_directory);
-   }
-
-   if (!pgmoneta_ends_with(to, "/"))
-   {
-      to = pgmoneta_append_char(to, '/');
-   }
-   to = pgmoneta_append(to, relative_file_path);
-
-   if (pgmoneta_copy_and_extract_file(from, &to))
```
trxvorr, line 1665: ..

src/libpgmoneta/extraction.c
Jubilee101: @sksingh2005 Replace the encrypt_file() with encryptor. It's ok if it breaks related UTs. We'll enable salt for encryptor later. Cc @jesperpedersen

src/libpgmoneta/extraction.c
sksingh2005: Good catch! That was indeed a problem in the previous version which used `pgmoneta_decrypt_file`. The latest push switched to one-pass streaming via the streamer's encryptor — `stream_restore_file` reads the source via vfile in read mode and writes to a separate destination, so the source backup file is never modified or deleted.
cc: @Jubilee101

src/libpgmoneta/info.c
sksingh2005: @trxvorr This call is now replaced by
```
pgmoneta_extract_backup_file(server, label, path, NULL, &extracted_file_path) 
```
public wrapper in `extraction.h` that internally calls `pgmoneta_extract_file(from, 0, true, &to)`. The streamer reads the source in read-only mode and writes the decrypted+decompressed output directly to the workspace, so the backup source is never modified or deleted.

src/libpgmoneta/extraction.c
sksingh2005: @Jubilee101 @jesperpedersen sir replaced the file-based `pgmoneta_decrypt_file` with the streamer's encryptor for one-pass decrypt+decompress. Added a code comment documenting the temporary limitation.

jesperpedersen:
@sksingh2005 CI

sksingh2005:
Sir the extraction now uses the streamer's encryptor for one-pass decrypt+decompress instead of pgmoneta_decrypt_file. The test encrypts with `pgmoneta_encrypt_file` (random salt), but the encryptor currently uses a zero salt — so decryption is getting failed 
cc: @Jubilee101

Jubilee101:
@sksingh2005 Merged, thanks for your contribution! We'll fix the encryptor soon.

PR #959, opened 2026-02-27
[#956] refactor: Use bitmask for compression constants

### LINKED ISSUE #956: Refactor: Use bitmask for COMPRESSION_XYZ constants
Replace compression constants with bitmask flags

shashidhar-bm:
- closes #956 

**changes**: 
- Redefine compression constants as bitmask (bit 4 = server, bits 0-3 = algorithm)
- Add `COMPRESSION_IS_SERVER()`, `COMPRESSION_ALGORITHM()` helper macros
- Replace all `if`/`else if` chains with switch (`COMPRESSION_ALGORITHM()`)
- Remove `is_server_side_compression()` function, use macro instead

### DIFF
--- a/src/include/pgmoneta.h
+++ b/src/include/pgmoneta.h
@@ -45,6 +45,7 @@ extern "C" {
 #include <sys/types.h>
 #include <openssl/ssl.h>
 
+// clang-format off
 #define VERSION                      "0.21.0"
 
 #define PGMONETA_HOMEPAGE            "https://pgmoneta.github.io/"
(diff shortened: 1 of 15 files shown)


Jubilee101:
LGTM @shashidhar-bm , but we should support compression on both sides (server & client). Current logic compresses on both sides when server side compression is configured, which is actually not correct behavior. There should be a set of separate constants that suggest both-side compression. We don't have to fix this in this PR, but we should save constants for that purpose.

Maybe it's better to decouple compression sides and compression methods. I think you are already doing that, but instead of having a constant for each and every combination, maybe we can just leave it as two sets of constants which can be combined together. 

Cc @jesperpedersen

src/include/pgmoneta.h
```diff
@@ -106,14 +106,26 @@ extern "C" {
...
+#define COMPRESSION_ALG_BZIP2        0x04
+
+#define COMPRESSION_NONE             0x00
+#define COMPRESSION_CLIENT_GZIP      0x01
+#define COMPRESSION_CLIENT_ZSTD      0x02
+#define COMPRESSION_CLIENT_LZ4       0x03
+#define COMPRESSION_CLIENT_BZIP2     0x04
+#define COMPRESSION_SERVER_GZIP      0x11
+#define COMPRESSION_SERVER_ZSTD      0x12
+#define COMPRESSION_SERVER_LZ4       0x13
```
jesperpedersen: I don't really like this ... if we want explicit `COMPRESSION_CLIENT_XYZ`, `COMPRESSION_SERVER_XYZ` then we should use `#define COMPRESSION_CLIENT_GZIP (COMPRESSION_TYPE_CLIENT | COMPRESSION_ALG_GZIP)`

I think we can use another bit for this... such that we can have a `COMPRESSION_TYPE_BOTH`

src/libpgmoneta/walfile/wal_summary.c
```diff
@@ -398,24 +398,20 @@ get_wal_file_name(char* dir_path, char* file)
    }
 
    /* Build expected compression/encryption suffix */
-   if (config->compression_type == COMPRESSION_CLIENT_GZIP ||
-       config->compression_type == COMPRESSION_SERVER_GZIP)
+   switch (COMPRESSION_ALGORITHM(config->compression_type))
```
jesperpedersen: Stuff like this we should have in `compression.h`

```
int
pgmoneta_compression_get_suffix(int type, char** suffix);
```

There are a lot of places where we can use a central implementation

src/libpgmoneta/configuration.c
```diff
@@ -1868,49 +1868,48 @@ pgmoneta_validate_main_configuration(void* shm)
...
-   {
-      if (config->compression_level < 1)
-      {
-         config->compression_level = 1;
-      }
-      else if (config->compression_level > 9)
-      {
-         config->compression_level = 9;
-      }
+      case COMPRESSION_ALG_GZIP:
```
jesperpedersen: in `compression.h`

```
int
pgmoneta_compression_get_level(int type, int* level);
```

src/libpgmoneta/link.c
```diff
@@ -495,19 +495,16 @@ trim_suffix(char* str)
    if (!pgmoneta_compare_string(str, "backup_label") &&
        !pgmoneta_compare_string(str, "backup_manifest"))
    {
-      switch (config->compression_type)
+      switch (COMPRESSION_ALGORITHM(config->compression_type))
```
jesperpedersen: We should call a function in `compression.h` which use `pgmoneta_remove_suffix` internally instead

jesperpedersen:
@shashidhar-bm I have added some general comments - there are likely other places that need fixing, but you get the general idea

jesperpedersen:
@shashidhar-bm Can you rebase, and update for the new layout ?

Jubilee101:
@shashidhar-bm Now that we have #967 done, I think the changes need to be ported to a few additional spots (streamer.c and compression.c for example).

src/include/pgmoneta.h
```diff
@@ -45,6 +45,7 @@ extern "C" {
 #include <sys/types.h>
 #include <openssl/ssl.h>
 
+// clang-format off
```
shashidhar-bm, line 48: @jesperpedersen , without this off line, `clang-format` reformats the surrounding macro definitions as well, what can be done on this?
(At present, formatting is re-enabled with `// clang-format on` after the macro block.)

Jubilee101:
LGTM @shashidhar-bm , can you rebase and resolve conflicts? Don't worry about UT failures. We have some issues in encryption which will be fixed later. Cc @jesperpedersen

jesperpedersen:
@shashidhar-bm CI

src/libpgmoneta/configuration.c
```diff
@@ -29,10 +29,10 @@
 /* pgmoneta */
 #include <pgmoneta.h>
 #include <aes.h>
+#include <compression.h>
 #include <configuration.h>
 #include <logging.h>
 #include <management.h>
-#include <memory.h>
```
jesperpedersen, line 35: Why ?

src/libpgmoneta/configuration.c
```diff
@@ -3551,7 +3508,6 @@ pgmoneta_conf_get(SSL* ssl, int client_fd, uint8_t compression, uint8_t encrypti
    double total_seconds;
 
    pgmoneta_start_logging();
-   pgmoneta_memory_init();
```
jesperpedersen, line 3518: Why ?

jesperpedersen (changes requested):

jesperpedersen:
@shashidhar-bm You need to revert the memory changes

jesperpedersen:
@shashidhar-bm Please, rebase

src/libpgmoneta/aes.c
```diff
@@ -248,29 +249,18 @@ pgmoneta_encrypt_wal(char* d)
...
-         break;
-      case COMPRESSION_CLIENT_LZ4:
-      case COMPRESSION_SERVER_LZ4:
-         compress_suffix = ".lz4";
-         break;
-      case COMPRESSION_CLIENT_BZIP2:
-         compress_suffix = ".bz2";
-         break;
-      case COMPRESSION_NONE:
+   {
```
jesperpedersen, line 256: Why ?

jesperpedersen: There are multiple of these

jesperpedersen:
@shashidhar-bm We don't need code blocks

src/libpgmoneta/aes.c
```diff
@@ -330,29 +317,18 @@ pgmoneta_encrypt_wal_file(char* d, char* f)
...
```
jesperpedersen: Here is another one

shashidhar-bm: sorry for that, i've updated it now

jesperpedersen:
Rebased, and merged.

Issue #958, opened 2026-02-26
fix conf get/set crashes by initializing memory in child processes


trxvorr:
port https://github.com/pgagroal/pgagroal/pull/758

sksingh2005:
I noticed other forked child process handlers (e.g. in security.c, prometheus.c, remote.c) already call memory_init() / memory_destroy(). Was  conf_get/conf_set simply overlooked when those were added, or were they added later without the same pattern?

jesperpedersen:
@sksingh2005 It was overlooked

PR #968, opened 2026-03-01
[#958] fix conf get/set crashes by initializing memory in child processes

sksingh2005:
Closes - [#958]

cc: @trxvorr


jesperpedersen:
Rebased, and merged.

Issue #947, opened 2026-02-25
Introduce restart_time() helper to restore reload logging


trxvorr:
check and port https://github.com/pgagroal/pgagroal/pull/737 if needed

PR #969, opened 2026-03-01
[#947] Introduce restart_time() helper to restore reload logging

sksingh2005:
Fixes - [#947]

cc: @trxvorr

### DIFF
--- a/src/libpgmoneta/configuration.c
+++ b/src/libpgmoneta/configuration.c
@@ -91,6 +91,7 @@ static void copy_user(struct user* dst, struct user* src);
 static int restart_bool(char* name, bool e, bool n);
 static int restart_int(char* name, int e, int n);
 static int restart_string(char* name, char* e, char* n);
+static int restart_time(char* name, pgmoneta_time_t e, pgmoneta_time_t n, bool requires_restart);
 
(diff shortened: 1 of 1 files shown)


src/libpgmoneta/configuration.c
```diff
@@ -5706,11 +5713,13 @@ transfer_configuration(struct main_configuration* config, struct main_configurat
    }
    config->common.log_level = reload->common.log_level;
 
-   if (pgmoneta_time_convert(config->verification, FORMAT_TIME_S) != pgmoneta_time_convert(reload->verification, FORMAT_TIME_S))
+   if (restart_time("verification", config->verification, reload->verification, true))
    {
       changed = true;
    }
 
+   memcpy(&config->verification, &reload->verification, sizeof(config->verification));
```
Amr-Shams: why would u need to populate the verification of the config?

sksingh2005: Ohh thansk sir !! 
The `memcpy` for `verification` shouldn't be there since it requires a restart. The original code only flagged changed = true without copying the value. I'll remove it.

jesperpedersen:
Merged.

PR #973, opened 2026-03-02
[972]fix recursive S3 upload symlink handling and error propagation

Amr-Shams:


Amr-Shams:
i might be wrong here, but as far as i know i had this issue, where the symlink would cause lots of trouble, so i safely ignore these files

jesperpedersen:
@Amr-Shams But, the symlinks are there to create a full backup... You need to replicate the symlinks on S3

Amr-Shams:
> @Amr-Shams But, the symlinks are there to create a full backup... You need to replicate the symlinks on S3

yep, totally fair point. if symlinks are part of making a full backup, I shouldn’t just skip them.
Quick check: on S3 do we want to upload the actual target file content, or save the symlink itself so we can recreate it on restore? Also for broken/cyclic links, should backup fail, or just warn and continue?

jesperpedersen:
@Amr-Shams Well, first question is: Does S3 support symlinks ? We need to be able to create the full restore from a S3 storage.

Create 2 backup locally - the 2nd one should use a lot of symlinks since the files are the same. Then try to replicate that with S3.

Just expand the S3 chapter in the manual with your findings - if S3 doesn't support symlinks then people will have to pay (if not using Garage) for the full file

jesperpedersen:
@Amr-Shams And, play around with it -- there is a test case that creates 3 backups, and then deletes the one in the middle. The symlinks are updated and works

Amr-Shams:
> @Amr-Shams Well, first question is: Does S3 support symlinks ? We need to be able to create the full restore from a S3 storage.
> 
> Create 2 backup locally - the 2nd one should use a lot of symlinks since the files are the same. Then try to replicate that with S3.
> 
> Just expand the S3 chapter in the manual with your findings - if S3 doesn't support symlinks then people will have to pay (if not using Garage) for the full file

Great point. i dug into this and the short version is: S3 itself doesn’t support symlinks as symlinks, so we can’t preserve link semantics there the same way we do on local storage.
What happens in practice is AWS-style sync behavior follows links and uploads file content as normal objects. That means restores from S3 are content-correct, but link relationships are not preserved, and storage can grow compared to local/Garage cases where links are cheap. i also tested the symlink-cycle case (A -> B, B -> A). AWS-style traversal does terminate and doesn’t recurse forever, so cycle handling is safe if traversal is implemented properly.
On your test scenario (3 backups, delete the middle, relink continues to work): that behavior is valid locally where symlink relationships exist, but it won’t map 1:1 to S3 because S3 only sees objects, not links.
a middle ground would to skip the recursion if already visited so we need to adjust the dfs here to use a cycle detection aka saving the visited links and check against them to detect the cycle

jesperpedersen:
@Amr-Shams Well, symlinks doesn't work then you have to copy the file instead. Every backup need to be able to be restored

Amr-Shams:
> @Amr-Shams Well, symlinks doesn't work then you have to copy the file instead. Every backup need to be able to be restored

Hey @jesperpedersen, you're totally right. After thinking it over, I'm just going to close this PR.

Using lstat was definitely the wrong move since S3 actually needs the real file contents to make the backups restorable. We definitely need it to follow those links.

Also, since a normal Postgres setup isn't going to have circular symlinks anyway or even the extra folders (very rare case anyways it is handled by the os not the nicest error but it won't break anything) , building out cycle detection or depth limits just feels like over-engineering for a problem that won't really happen. Looks like the original code had it right all along!

Thanks for talking this through with me and for the sanity check, I really appreciate the guidance. I'll go ahead and close this PR and the issue.

Issue #950, opened 2026-02-25
Fix daemon crash and CLI disconnect on configuration reload


trxvorr:
check and port if needd https://github.com/pgagroal/pgagroal/pull/730

PR #978, opened 2026-03-02
[#950] Fix daemon crash and CLI disconnect on configuration reload

sksingh2005:
Fixes : [#950]

cc: @trxvorr


src/main.c
```diff
@@ -2471,9 +2477,8 @@ accept_fatal(int error)
 }
 
 static bool
```
Amr-Shams: you can change the sign to be void here no need for a return

sksingh2005: Ohh yes since no caller checks the return value of this

jesperpedersen:
Rebased, and merged.

PR #979, opened 2026-03-03
robust memory cleanup in error paths

trxvorr:
Refactored `pgmoneta_destroy_walfile` to independently check and destroy `wf->records` and `wf->page_headers`. The previous combined check would short-circuit if one deque failed to initialize, leaking the rest of the `struct walfile`.
Added comprehensive cleanup for the globally accessible `partial_record` structure in the `error:` path of `pgmoneta_summarize_wal`.
 Ensured the dynamically allocated destination path (`to`) is properly freed in `pgmoneta_encrypt_file`
 Fixed multiple pgmoneta_log_fatal calls in decode_xlog_record that had format specifiers (e.g., %u, %s) but were missing the corresponding variables.
 In pgmoneta_wal_parse_wal_file, explicitly transferred ownership of long_header to wal_file->long_phd and immediately nullified the local pointer (long_header = NULL). This prevents the local error-handling path from freeing the pointer while the caller simultaneously attempts to free it via pgmoneta_destroy_walfile. Updated subsequent math calculations to safely reference the struct.
 Added a missing free(page_header) in the early error-return path of read_all_page_headers
 Explicitly used pgmoneta_deque_peek_last in the finish: block to ensure the next_lsn termination logic is applied to the guaranteed last record, rather than relying on a potentially stale loop pointer.
 
 CI:
 Replaced ck_assert_msg with explicit conditional checks (if (!ret)) when summarizing and saving WALs. Because libcheck instantly aborts the thread on assertion failure, the previous code would skip downstream free() statements during slow I/O runs, triggering LeakSanitizer. Memory is now cleanly freed before ck_abort_msg is called.
 
 Edit: The C code is stable and free of ASan/LSan errors. The final CI failure is an llvm-cov zlib decompression error caused by an incomplete .profraw flush during test failures. The actual memory leaks and SEGVs are resolved.

### DIFF
--- a/src/libpgmoneta/aes.c
+++ b/src/libpgmoneta/aes.c
@@ -470,6 +470,10 @@ pgmoneta_encrypt_file(char* from, char* to)
 
    if (encrypt_file(from, to, 1) != 0)
    {
+      if (flag)
+      {
+         free(to);
+      }
       return 1;
    }
 
--- a/src/libpgmoneta/walfile.c
+++ b/src/libpgmoneta/walfile.c
@@ -321,46 +321,55 @@ pgmoneta_destroy_walfile(struct walfile* wf)
       return;
    }
 
-   if (pgmoneta_deque_iterator_create(wf->records, &record_iterator) || pgmoneta_deque_iterator_create(wf->page_headers, &page_header_iterator))
+   /* Cleanup records - only if deque was initialized */
+   if (wf->records != NULL)
    {
-      return;
+      if (pgmoneta_deque_iterator_create(wf->records, &record_iterator) == 0)
+      {
+         while (pgmoneta_deque_iterator_next(record_iterator))
+         {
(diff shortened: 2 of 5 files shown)


trxvorr:
@shashidhar-bm could you check it out now

jesperpedersen:
@trxvorr CI

trxvorr:
hey @jesperpedersen I genuinely can't figure out what's wrong with the CI

it's 1-4 fails everytime i force push with no changes

trxvorr:
@shashidhar-bm could you figure out what's wrong with it?

jesperpedersen:
@trxvorr Try and rebase, there are cores being dumped

trxvorr:
no fails the previous run
force pushed again for a 2nd run

trxvorr:
2nd run failed

jesperpedersen:
Merged.

Issue #902, opened 2026-02-02
Delete Support for S3


Amr-Shams:
**Is your feature request related to a problem? Please describe**

To align with full S3 compatibility, we need proper XML support and missing S3 APIs to keep S3 state clean and consistent with local storage. This requires adding lightweight XML build/parse support and integrating it with the existing S3 engine. The change does not affect the storage layer structure and can be implemented without heavy dependencies.

**Describe the solution you’d like**

1. add HTTP support for XML content types(aka removing the hard coded content-type header field) #893
2. add XML build/parse utilities(might use libxml2 if this needs to be heavy 
3. implement S3 `ListObjectsV2` #916 
4. implement S3 `DeleteObjects` (bulk)

Amr-Shams:
here is a rough header file for the xml(s3 only util) 
```c
/**
 * extract all values of a given tag from XML string.
 * @param xml The XML string to parse
 * @param tag The tag name (e.g. "Key")
 * @param values Output: deque of char* values (in case of errors as well)
 * @return 0 on success
 */
int pgmoneta_xml_extract_tag(char* xml, char* tag, struct deque** values);

/**
 * parse S3 ListBucketResult XML to extract object keys.
 * @param xml The XML response body
 * @param keys Output: deque of char* keys(files)
 * @return 0 on success
 */
int pgmoneta_xml_parse_s3_list(char* xml, struct deque** keys);

/**
 * parse S3 ListBucketResult to check for truncation.
 * @param xml The XML response body
 * @param is_truncated Output: true if more results available
 * @param continuation_token Output: token for next request (NULL if not truncated)
 * @return 0 on success
 */
int pgmoneta_xml_parse_s3_list_truncated(char* xml, bool* is_truncated, char** continuation_token);
```

Amr-Shams:
> here is a rough header file for the xml(s3 only util)
> 
> /**
>  * extract all values of a given tag from XML string.
>  * @param xml The XML string to parse
>  * @param tag The tag name (e.g. "Key")
>  * @param values Output: deque of char* values (in case of errors as well)
>  * @return 0 on success
>  */
> int pgmoneta_xml_extract_tag(char* xml, char* tag, struct deque** values);
> 
> /**
>  * parse S3 ListBucketResult XML to extract object keys.
>  * @param xml The XML response body
>  * @param keys Output: deque of char* keys(files)
>  * @return 0 on success
>  */
[...]

PR #987, opened 2026-03-05
[#902]   implement s3 delete command

Amr-Shams:
- pgmoneta-cli s3 delete <server> <prefix> now exists, backed by a matching management command, workflow type, and S3 handler that paginates list/delete calls, signs the XML payload, parses delete responses, and reports per-key issues while avoiding metadata lookups when only remote files exist.
 - previously the delete command expected local backup metadata, so you couldn’t clean up orphaned S3 objects; the new command lets you trim remote prefixes without touching the normal delete workflow.

### DIFF
--- a/src/include/management.h
+++ b/src/include/management.h
@@ -95,6 +95,7 @@ extern "C" {
 #define MANAGEMENT_LIST_USERS     104
 
 #define MANAGEMENT_S3_LS          200
+#define MANAGEMENT_S3_DELETE      201
 
 /**
  * Management categories
@@ -171,6 +172,7 @@ extern "C" {
 #define MANAGEMENT_ARGUMENT_RETENTION_YEARS       "RetentionYears"
 #define MANAGEMENT_ARGUMENT_S3_KEY                "S3Key"
 #define MANAGEMENT_ARGUMENT_S3_OBJECTS            "S3Objects"
+#define MANAGEMENT_ARGUMENT_S3_PREFIX             "S3Prefix"
 #define MANAGEMENT_ARGUMENT_SERVER                "Server"
 #define MANAGEMENT_ARGUMENT_SERVERS               "Servers"
 #define MANAGEMENT_ARGUMENT_SERVER_SIZE           "ServerSize"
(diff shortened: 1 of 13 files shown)


src/include/management.h
```diff
@@ -383,6 +385,13 @@ extern "C" {
 #define MANAGEMENT_ERROR_LIST_S3_NETWORK                    2904
 #define MANAGEMENT_ERROR_LIST_S3_ERROR                      2905
 
+#define MANAGEMENT_ERROR_DELETE_S3_NOSERVER                 2910
```
jesperpedersen: We bump by 100 per command, so these are `300x`

src/libpgmoneta/s3.c
```diff
@@ -42,6 +42,33 @@
 #include <workflow.h>
 
 #define NAME "s3"
+
+static bool
+s3_is_safe_prefix(char* prefix)
```
jesperpedersen, line 47: This need some work - `pgmoneta_contains`

src/libpgmoneta/se_s3.c
```diff
@@ -32,6 +32,7 @@
 #include <http.h>
 #include <logging.h>
 #include <security.h>
+#include <stddef.h>
```
jesperpedersen: This is in the system section

src/libpgmoneta/se_s3.c
```diff
@@ -42,20 +43,28 @@
 #include <stdlib.h>
 #include <string.h>
 
+#define S3_MAX_DELETE_KEYS 1000
```
jesperpedersen: We need a list

Amr-Shams: i am not sure i am follwoing up here, you mean store the entire objects to a list and then batch them?, i have been doing this list page → delete → next page

Amr-Shams: The `ListObjectsV2` API already limits the number of returned objects to **1000**, which is the same limit enforced by the `DeleteObjects` API. Because of this, the number of objects returned by the listing step will always be ≤ the maximum number of keys allowed in a delete request.

Given that both operations share the same bound, this guard might be unnecessary here.

jesperpedersen: Ok

jesperpedersen:
@Amr-Shams Please, rebase

src/include/management.h
```diff
@@ -407,8 +416,7 @@ extern "C" {
  * @param json The target json
  * @return 0 upon success, otherwise 1
  */
-int
-pgmoneta_management_create_header(int32_t command, uint8_t compression, uint8_t encryption, int32_t output_format, struct json** json);
+int pgmoneta_management_create_header(int32_t command, uint8_t compression, uint8_t encryption, int32_t output_format, struct json** json);
```
jesperpedersen: Formatting

jesperpedersen:
@Amr-Shams CI

jesperpedersen:
Rebased, and merged.

PR #988, opened 2026-03-05
Skip in-progress .partial WAL segments in summarize_walfiles

shashidhar-bm:
this issue is only on `0.20.x` , because of the outdated `get_wal_file_name() `

how it is handled in `main`:
```
read 1C.partial → check each record LSN → all > e_lsn → discard all → BRT unchanged
```

current pr fix: 
```
see 1C.partial → continue → never read → BRT unchanged
```
adding continue just before `get_wal_file_name()`
this is safe because `.partial` only exists after `pg_switch_wal()`, which is always called after `e_lsn` is captured, so its records are always > `e_lsn`.

### DIFF
--- a/src/libpgmoneta/walfile/wal_summary.c
+++ b/src/libpgmoneta/walfile/wal_summary.c
@@ -301,6 +301,12 @@ retry1:
       char* file = (char*)file_iterator->value->data;
       char* fn = NULL;
 
+      if (pgmoneta_ends_with(file, ".partial"))
+      {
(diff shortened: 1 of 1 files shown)


shashidhar-bm:
@jesperpedersen, I felt this could be a straightforward fix, but are there any edge cases where a `.partial` file could contain records within [ `s_lsn`, `e_lsn` )?

jesperpedersen:
@shashidhar-bm Closing this - we are focus on `main` now

PR #989, opened 2026-03-06
Fix unsafe file overwrite in encrypt_file

trxvorr:


jesperpedersen:
Merged.

Issue #971, opened 2026-03-02
AES encryptor: Reuse buffer


Jubilee101:
Currently AES encryptor mallocs memory for every chunk of outputs. We should malloc once and reuse that buffer. The tricky part is that we don't know how much memory we need beforehand.

sksingh2005:
Sir I am thinking of letting the encryptor own a single reusable output buffer internally.

- Allocate on first use, grow only if a larger chunk arrives
- Caller receives a borrowed pointer, valid until the next call
- No callback signature changes needed
- Apply the same model to the noop (no-encryption) path for uniform behavior

This keeps cipher-specific sizing details internal to the encryptor and avoids pushing buffer management onto callers. How is this approach sir ? If it is good I would like to work upon this sir.

Jubilee101:
Exactly what’s on my mind. Go for it, and good luck!

sksingh2005:
Sure sir !!

PR #990, opened 2026-03-06
[#971] Reuse output buffer in AES encryptor

sksingh2005:
Fixes - [#971]

Description
This PR removes per-chunk output buffer allocation in the AES streaming encryptor and reuses a single internal buffer for the lifetime of each encryptor instance.

Previously, aes_encryptor_process() allocated a fresh output buffer for every chunk, and the caller freed it immediately afterward. For large backups this caused thousands of repeated malloc/free cycles for similarly sized buffers. The same pattern existed in the noop encryptor path.

This change makes the encryptor own its output buffer internally:

- allocate on first use
- reuse across subsequent chunks
- grow only if a later chunk requires more capacity
- return a borrowed out_buf pointer valid until the next encrypt/decrypt call or close()

The same model is applied to the noop encryptor so buffer ownership is uniform across encryption modes. No callback signatures were changed.

### DIFF
--- a/src/libpgmoneta/aes.c
+++ b/src/libpgmoneta/aes.c
@@ -38,6 +38,7 @@
 /* System */
 #include <dirent.h>
 #include <errno.h>
+#include <limits.h>
 #include <openssl/evp.h>
 #include <openssl/rand.h>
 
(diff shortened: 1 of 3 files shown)


Amr-Shams (commented):
Nice idea with the reusable buffer. Out of curiosity, did you measure whether the malloc/free here actually shows up in profiling?
cause i happen to do the same thing, but it was not a thread safe, and the buffer expands and never shrinks, but if the profiling shows something big then i think we can expand it to a some sort of safe memory pool (like the dequeue in the project)

src/libpgmoneta/aes.c
```diff
@@ -79,8 +80,57 @@ struct aes_encryptor
...
+   unsigned char* tmp = NULL;
+
+   if (required == 0 || *capacity >= required)
+   {
+      return 0;
+   }
+
+   if (*capacity > SIZE_MAX / 2)
+   {
+      new_cap = required;
```
Amr-Shams, line 113: i would go for a safe minimal init size,smthing like this ` new_cap = MAX(required, MIN_INIT_SIZE)`

src/libpgmoneta/aes.c
```diff
@@ -1635,82 +1725,129 @@ aes_encryptor_process(struct encryptor* encryptor, void* in_buf, size_t in_size,
...
-   memcpy(out, in_buf, in_size);
-   *out_buf = out;
+
+   if (ensure_capacity(&this->out_buf, &this->out_capacity, in_size))
+   {
+      pgmoneta_log_error("noop_encryptor_decrypt: failed to ensure buffer capacity");
+      goto error;
+   }
+
+   memcpy(this->out_buf, in_buf, in_size);
```
Amr-Shams, line 1960: since this is a noop, why are we copying anything? just do `*out_buf = in_buf` and skip the allocation entirely, as long as i see it

sksingh2005:
> Nice idea with the reusable buffer. Out of curiosity, did you measure whether the malloc/free here actually shows up in profiling? cause i happen to do the same thing, but it was not a thread safe, and the buffer expands and never shrinks, but if the profiling shows something big then i think we can expand it to a some sort of safe memory pool (like the dequeue in the project)

@Amr-Shams Profiling shows AES computation dominates — malloc/free doesn't show up as a bottleneck.

src/libpgmoneta/aes.c
sksingh2005: For this path I kept first allocation equal to the actual required size, since chunk sizes are usually stable and close to the streamer buffer size. I can add a small minimum initial capacity if you prefer, but I did not see a correctness or obvious performance need for it here ??

Amr-Shams: This means the first allocation has zero headroom, the next slightly larger chunk will trigger another realloc immediately(which is high props to happen)

src/libpgmoneta/aes.c
sksingh2005: The copy keeps the `out_buf` lifetime consistent with the AES path  always encryptor-owned, valid until the next call. Skipping it would tie `out_buf` validity to the caller's input lifetime, which is a different contract.

Amr-Shams: Man u are cooking

src/libpgmoneta/aes.c
sksingh2005: I was thinking about the `MIN_INIT_SIZE` suggestion, but in this path `required` is already roughly the chunk size, so a smaller fixed minimum would not actually add any headroom. Would a small dynamic headroom on the first allocation make more sense here? I tried `required + required / 4` for the initial allocation, with an overflow guard, and kept the existing doubling behavior for later growth. That seemed to address the immediate-realloc concern without introducing an arbitrary fixed size. What are your thoughts in this ?

Amr-Shams: it is better to have a headroom, but yes this is a little over complicating from my side, it will do and with very little effect

jesperpedersen:
@trxvorr Cc:

jesperpedersen:
Merged.

PR #992, opened 2026-03-07
fix wal summary test flakiness

trxvorr:
rewrite get_wal_file_name() to try multiple file forms (compressed, uncompressed, .partial) with exponential backoff retry loop instead of assuming the WAL file is already in its final compressed state.

under AddressSanitizer overhead in clang Debug builds, the WAL file may not yet be compressed when the check runs, causing intermittent test failures.

The new approach:
- Check for compressed file (fully processed)
- Check for uncompressed file (not yet compressed)
- Check for .partial file (still-active segment)
- Retries with exponential backoff (500ms to 30s, 10 retries)


trxvorr:
```

TEST START: test_pgmoneta_wal_summary_fn
Running suite pgmoneta_test_wal_summary
/home/runner/work/pgmoneta/pgmoneta/test/testcases/test_wal_summary.c:140:P:test_wal_summary:test_pgmoneta_wal_summary:0: Passed
```

jesperpedersen:
Merged.

jesperpedersen:
... the correct solution is to have a call-back for `CHECKPOINT` in the WAL stream

jesperpedersen:
... well, together with the compression and encryption pipeline

PR #993, opened 2026-03-07
Implement proactive WAL summarization and CHECKPOINT scanner

trxvorr:
reverted `wal_summary.c` changes from #992

### DIFF
--- a/src/include/walfile/wal_summary.h
+++ b/src/include/walfile/wal_summary.h
@@ -47,7 +47,7 @@
  * @return 0 is success, otherwise failure
  */
 int
-pgmoneta_summarize_wal(int srv, char* wal_dir, uint64_t start_lsn, uint64_t end_lsn, block_ref_table** brt);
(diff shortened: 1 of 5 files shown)


saraanbih:
Hi @trxvorr 
Thanks for your work

There are only a few changes, based on my review, we should make `pgmoneta_wal_scan_checkpoints` return int (1 on failure & 0 on success), not void, also see if we make it int, when we return with failure, the error message, e.g., `pgmoneta_log_error`. 

Go on...

trxvorr:
hey @saraanbih 

thanks for taking a look at this

i designed `pgmoneta_wal_scan_checkpoints` to return void deliberately because it operates as a zero allocation state machine that strictly parses an existing in memory buffer `(msg->data)`

since the function doesnt allocate any memory, open any files, perform network I/O, or have any failure modes, it actually cannot encounter runtime errors after it has been called. Because it physically cant fail, there are no error states to log or return

saraanbih:
@trxvorr 
Ah, I got it. I was looking at it from another angle, expecting a possible error path, but what you said makes sense

saraanbih:
@jesperpedersen LGTM

jesperpedersen:
Rebased, and merged.

PR #995, opened 2026-03-07
Fix side-effects inside assert() in tscommon.c

trxvorr:


jesperpedersen:
@shashidhar-bm Cc:

jesperpedersen:
@trxvorr Please, rebase

shashidhar-bm:
> @shashidhar-bm Cc:

@jesperpedersen LGTM. In NDEBUG the assert is compiled out, so the call has to be on its own line so it always runs.
```
#ifdef NDEBUG
   #define assert(expression)  ((void)0) 
```

jesperpedersen:
Rebased, and merged.

PR #996, opened 2026-03-07
Harden file handling

trxvorr:


Jubilee101:
Thanks for this @trxvorr . Can you also look at vfile_local? It can likely use the same protection. If the reader and writer use different protection we can perhaps separate them to vfile_create_local_write() and vfile_create_local_reader(). 

And you need to rebase, I fixed the encryptor.

jesperpedersen:
`pgmoneta_permission` ?

jesperpedersen:
@trxvorr Please, rebase

src/libpgmoneta/vfile_local.c
Jubilee101: Shouldn't you follow the same pattern here just as the rest of what you did for other fopen()? If you don't need special handling for writer then there's no point of separating the API to create_reader and create_writer.

src/include/vfile.h
```diff
@@ -74,14 +74,23 @@ struct vfile
...
+/**
+ * Create a local vfile writer
  * @param file_path The file path
  * @param mode The open mode
  * @param vfile [out] The vfile
  * @return 0 if success, 1 if otherwise
  */
 int
-pgmoneta_vfile_create_local(char* file_path, char* mode, struct vfile** vfile);
+pgmoneta_vfile_create_local_writer(char* file_path, char* mode, struct vfile** vfile);
```
Jubilee101: Hey @trxvorr , if `pgmoneta_fopen_secure` handles the mode internally then we don't need to distinguish reader and writer creation. The old way of `pgmoneta_vfile_create_local(char* file_path, char* mode, struct vfile** vfile);` will do.

Jubilee101 (changes requested):

trxvorr:
its not ready yet there are a few issues i need to fix

jesperpedersen:
@trxvorr PTAL me when ready

jesperpedersen:
@trxvorr WDYT ?

jesperpedersen:
@trxvorr Rebase, please

trxvorr:
@jesperpedersen will do, my schedule just freed up will work on the final bits left

trxvorr:
@jesperpedersen i think its ready PTAL

jesperpedersen:
@trxvorr Please, rebase and fix

jesperpedersen:
Merged.

jesperpedersen:
Please, cross-port to the other projects

Issue #948, opened 2026-02-25
Fix uninitialized client_addr struct in accept()


trxvorr:
check and port https://github.com/pgagroal/pgagroal/pull/744 if needed

PR #999, opened 2026-03-08
[#948] Fix uninitialized client_addr in accept()

sksingh2005:
Fixes - [#948]

Zero-initialize `client_addr` before each `accept()` call in main.c. This avoids leaving trailing bytes uninitialized when `accept()` returns a shorter sockaddr than sockaddr_in6.


jesperpedersen:
Rebased, and merged.

Issue #1019, opened 2026-03-13
Test: AES


jesperpedersen:
Create a test case that uses AES for backup

sksingh2005:
Beyond running a standard backup with AES enabled and verifying it completes without errors, are there specific edge cases or validation steps ? For instance, verifying the encrypted files on disk, attempting a restore, or testing backup chains involving mixed encryption states.

jesperpedersen:
@sksingh2005 Yeah, I think it need to be a full backup/verify/restore cycle. We can do a manual test if the files ends with .aes

sksingh2005:
Can i take this issue sir ?

jesperpedersen:
@sksingh2005 Go for it !

PR #1026, opened 2026-03-14
[#1019] Add AES backup verify/restore CLI test

sksingh2005:
## Summary

Add end-to-end AES coverage for the backup workflow.

This adds a CLI integration test that:
- enables `aes-256-cbc`
- takes a full backup
- checks that backup metadata reports AES
- checks that backup artifacts on disk include `.aes` files
- runs `verify` and asserts there are no failed files
- runs `restore` and asserts expected plaintext files are restored

### DIFF
--- a/src/libpgmoneta/restore.c
+++ b/src/libpgmoneta/restore.c
@@ -310,17 +310,52 @@ error:
 bool
 pgmoneta_is_restore_last_name(char* file_name)
 {
+   char* normalized = NULL;
+   char* tmp = NULL;
+   bool result = false;
    int number_of_elements = sizeof(restore_last_files_names) / sizeof(restore_last_files_names[0]);
 
-   for (int i = 0; i < number_of_elements; i++)
+   if (file_name == NULL)
    {
-      if (strstr(restore_last_files_names[i], file_name) != NULL)
+      return false;
+   }
+
+   normalized = file_name;
+
+   while (*normalized == '/')
+   {
+      normalized++;
+   }
+
+   if (pgmoneta_starts_with(normalized, "data/"))
+   {
+      tmp = pgmoneta_remove_prefix(normalized, "data/");
+      if (tmp != NULL)
       {
-         return true;
+         normalized = tmp;
+
+         while (*normalized == '/')
+         {
+            normalized++;
+         }
       }
    }
 
-   return false;
+   for (int i = 0; i < number_of_elements; i++)
+   {
+      if (pgmoneta_compare_string(normalized, restore_last_files_names[i] + 1))
+      {
+         result = true;
+         break;
+      }
+   }
+
+   if (tmp != NULL)
+   {
+      free(tmp);
+   }
+
+   return result;
 }
 
 void
--- a/test/include/tsclient.h
+++ b/test/include/tsclient.h
@@ -77,11 +77,12 @@ pgmoneta_tsclient_restore(char* server, char* backup_id, char* position, int exp
  * @param backup_id the backup
  * @param directory the directory
  * @param files the files
+ * @param response optional output for JSON response (NULL to ignore)
  * @param expected_error expected error code
  * @return 0 upon success, otherwise 1
  */
 int
-pgmoneta_tsclient_verify(char* server, char* backup_id, char* directory, char* files, int expected_error);
+pgmoneta_tsclient_verify(char* server, char* backup_id, char* directory, char* files, struct json** response, int expected_error);
 
 /**
  * Execute archive command on the server
--- a/src/libpgmoneta/compression.c
+++ b/src/libpgmoneta/compression.c
@@ -367,6 +367,14 @@ pgmoneta_compression_trim_suffix(char* str, int compression_type, int encryption
       return 1;
    }
 
+   /* Strip encryption suffix first so ".zstd.aes" becomes ".zstd" before compression trimming. */
+   if (encryption != ENCRYPTION_NONE && pgmoneta_ends_with(res, ".aes"))
+   {
+      tmp = pgmoneta_remove_suffix(res, ".aes");
+      free(res);
+      res = tmp;
+   }
+
    /* Strip the compression algorithm suffix, if any */
    if (!pgmoneta_compression_get_suffix(compression_type, &alg_suffix) && alg_suffix != NULL)
    {
@@ -376,14 +384,6 @@ pgmoneta_compression_trim_suffix(char* str, int compression_type, int encryption
          free(res);
          res = tmp;
       }
-   }
-
-   /* Strip encryption suffix if present */
-   if (encryption != ENCRYPTION_NONE && pgmoneta_ends_with(res, ".aes"))
-   {
-      tmp = pgmoneta_remove_suffix(res, ".aes");
-      free(res);
-      res = tmp;
    }
 
    *result = res;
--- a/src/libpgmoneta/link.c
+++ b/src/libpgmoneta/link.c
@@ -31,6 +31,7 @@
 #include <compression.h>
 #include <link.h>
 #include <logging.h>
+#include <restore.h>
 #include <utils.h>
 
 /* system */
@@ -90,6 +91,7 @@ pgmoneta_link_manifest(char* base_from, char* base_to, char* from, struct art* c
             // file in newer dir is not added nor changed, nor is an incremental file
             if (!pgmoneta_art_contains_key(added, from_file_trimmed) &&
                 !pgmoneta_art_contains_key(changed, from_file_trimmed) &&
+                !pgmoneta_is_restore_last_name(from_file_trimmed) &&
                 !pgmoneta_is_incremental_path(from_file_trimmed))
             {
                to_entry = pgmoneta_append(to_entry, base_to);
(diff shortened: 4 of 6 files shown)


sksingh2005:
Not able to convert to draft dont know why. Its not ready yet

sksingh2005:
@jesperpedersen sir I checked the PG14 and PG15 CI failures. In both logs, the AES CLI test passes, while the failing tests are all incremental-chain related and share the same error: `\ERROR message.c:1286 Message length not available`. From the logs and code path, this looks like a  bug in the PG14-16 incremental query-response handling around `pgmoneta_ext_get_files('.')`, where partial response data is scanned before the full message has arrived. 
Should I open a issue for this ?

jesperpedersen:
@sksingh2005 Yeah, it sounds like we need to fix that first

shashidhar-bm:
@jesperpedersen, this seems a bit unusual,  I’m encountering this log error locally on the latest commit (main) itself., I’ll also look into it from my side.

jesperpedersen:
@sksingh2005 Please, rebase

src/libpgmoneta/configuration.c
```diff
@@ -4173,6 +4173,18 @@ apply_main_configuration(struct main_configuration* config, struct server* srv,
             config->progress = false;
          }
       }
+      else if (!strcmp(key, "encryption"))
```
jesperpedersen: What is this ?

jesperpedersen (changes requested):

src/libpgmoneta/configuration.c
sksingh2005: That was from an earlier version of the AES test where I used `conf set encryption ..` plus reload. The test no longer works that way, so this `configuration.c` change is stale. I was removing that.

test/include/tsclient.h
```diff
@@ -77,11 +77,12 @@ pgmoneta_tsclient_restore(char* server, char* backup_id, char* position, int exp
  * @param backup_id the backup
  * @param directory the directory
  * @param files the files
+ * @param response optional output for JSON response (NULL to ignore)
  * @param expected_error expected error code
  * @return 0 upon success, otherwise 1
  */
 int
-pgmoneta_tsclient_verify(char* server, char* backup_id, char* directory, char* files, int expected_error);
+pgmoneta_tsclient_verify(char* server, char* backup_id, char* directory, char* files, struct json** response, int expected_error);
```
sksingh2005, line 85: @jesperpedersen sir I added an optional response output to `pgmoneta_tsclient_verify(...)` so the AES test can assert that `verify` reports zero failed files, not just that the command returns success. I thought that gave slightly stronger coverage for the backup/verify/restore cycle. Is that okay to keep, or would you prefer I simplify it and only assert verify success?

jesperpedersen: @shashidhar-bm WDYT ?

shashidhar-bm: LGTM, as the status check alone doesn't cover the failed-files case.

jesperpedersen:
@sksingh2005 Is this still a draft ?

sksingh2005:
@jesperpedersen sir Was thinking of extending AES integration coverage to include full + incremental backup chains, with verify and restore assertions across the chain. Would you prefer that as a follow-up PR?

jesperpedersen:
@sksingh2005 We can take that as part of this PR

jesperpedersen:
@sksingh2005 This fails locally - please, help @Sakshii-27 on #1039 and #1040 - that might help

jesperpedersen:
@sksingh2005 It still fails locally... so we need the verify PRs in first

sksingh2005:
> @sksingh2005 It still fails locally... so we need the verify PRs in first

Yes sir !!

jesperpedersen:
@sksingh2005 `test_cli_backup_verify_restore_aes_chain` still fails locally after a rebase

jesperpedersen:
`2026-03-19 09:45:14 ERROR backup.c:92 Backup: Server primary is not WAL streaming`

sksingh2005:
@jesperpedersen sir I did not consistently reproduce the same local failure line, but it points to the same pre-backup readiness path. I think `backup.c` means backup started before WAL streaming was active ? 
I updated the test helper so `mode online` waits for `online + valid + wal_streaming > 0` before returning. `test_cli_backup_verify_restore_aes_chain` passes locally for me in both normal and `ci` runs.

sksingh2005:
> @jesperpedersen sir I did not consistently reproduce the same local failure line, but it points to the same pre-backup readiness path. I think `backup.c` means backup started before WAL streaming was active ? I updated the test helper so `mode online` waits for `online + valid + wal_streaming > 0` before returning. `test_cli_backup_verify_restore_aes_chain` passes locally for me in both normal and `ci` runs.

@jesperpedersen sir  I’ve removed that approach because the test process was polling its own sh mem rather than the pgmoneta daemon’s live state.
I updated the test helper around the exact failing path.
After `mode online`, the first backup can race with WAL streaming startup, so the test now retries only that first backup and only when the daemon returns `MANAGEMENT_ERROR_BACKUP_WAL` (`"Server primary is not WAL streaming"`). Other errors still fail immediately. Also updated  `pgmoneta_tsclient_mode()` so `online` / `offline` checks the actual `Response.Online` state instead of only trusting `Outcome.Status`.

test/include/tsclient.h
```diff
@@ -49,6 +49,16 @@ extern "C" {
...
 
+/**
+ * Execute backup command and return the daemon error code
+ * @param server the server to perform backup on
+ * @param incremental execute backup in incremental mode
+ * @return 0 on success, positive management error code on daemon error,
+ *         -1 on connection/protocol failure
+ */
+int
+pgmoneta_tsclient_backup_error(char* server, char* incremental);
```
jesperpedersen: Why can't we use the function above ?

jesperpedersen (changes requested):

test/include/tsclient.h
sksingh2005: `pgmoneta_tsclient_backup()` only gives success/failure, but this retry needs the actual daemon error so it can retry only on `MANAGEMENT_ERROR_BACKUP_WAL` and still fail immediately on other errors.
So I added `pgmoneta_tsclient_backup_error()` for that distinction. If you’d prefer, I can keep the same behavior with a local helper instead of exposing it in the public test header.

jesperpedersen: @sksingh2005 Lets start with a local helper, and I'll look in detail once CI is green

jesperpedersen:
@sksingh2005 It still fails locally

sksingh2005:
@jesperpedersen sir can i have logs i am not able to reproduce the failure.

jesperpedersen:
@sksingh2005 I'll send them to you. In the meantime look at the new AES integration

jesperpedersen:
@sksingh2005 Please, rebase - as verify should be better now

sksingh2005:
> @sksingh2005 Please, rebase - as verify should be better now

Rebased. `test_cli_backup_verify_restore_aes_chain` passing on my side sir.

sksingh2005:
> @sksingh2005 I'll send them to you. In the meantime look at the new AES integration

@jesperpedersen sir just to confirm by “new AES integration”, do you want the CLI coverage added in this PR, or the test to create encrypted backups through the real config-driven AES backup path instead of post-processing the fixture after backup creation? I did not clearly understood the point, sorry.

jesperpedersen:
@sksingh2005 Use

```
[pgmoneta]
...
encryption = aes
compression = zstd
```

sksingh2005:
Ohh understood. I’ll rework the tests to use the config driven AES + zstd backup path via the `[pgmoneta]` settings, instead of locally rewriting/encrypting the fixture after backup creation.

sksingh2005:
@jesperpedersen sir I updated the AES CLI tests to use the config-driven `compression = zstd` / `encryption = aes` path.
What I observed:
- both new AES CLI tests now pass on the intended path
- when I tried removing the PG15 gate, that exposed a  PG14 incremental restore/verify issue where verify uses invalid flattened paths during the restore/combine flow

So should I keep the PG15 gate in this PR and the PG14 behavior as a separate follow-up issue ?

jesperpedersen:
@sksingh2005  For compression and encryption then PG17 is enough

sksingh2005:
@jesperpedersen sir PTAL !!

src/libpgmoneta/restore.c
```diff
@@ -310,11 +310,29 @@ pgmoneta_get_restore_last_files_names(char*** output)
...
    int number_of_elements = sizeof(restore_last_files_names) / sizeof(restore_last_files_names[0]);
 
+   if (file_name == NULL)
+   {
+      return false;
+   }
+
+   normalized = file_name;
+
+   if (pgmoneta_starts_with(normalized, "data/"))
```
jesperpedersen: pgmoneta_remove_suffix

jesperpedersen:
@sksingh2005 There are no `!` for people below Committer

jesperpedersen:
Merged.

PR #1034, opened 2026-03-15
Implement s3 restore to download backups from S3

### LINKED ISSUE #1010: S3 Restore Support (GET Object)
**Is your feature request related to a problem? Please describe.**
Currently pgmoneta has no way to restore a backup stored in S3. Once a backup is uploaded, there is no mechanism to retrieve it back to a local path, making S3 storage effectively write-only from pgmoneta's perspective.

**Describe the solution you'd like**
Add `s3_restore_files()` in `src/libpgmoneta/s3.c`: enumerate all objects under a backup prefix and restore each to its mirrored local path, recreating the directory structure as needed. GET requests should be signed with Signature V4. Both empty files and nested directory trees should be handled correctly.

**Additional information**
this is now possible cause we have #916 integrated

Amr-Shams:
## Summary

  Implements `pgmoneta-cli s3 restore <server> <timestamp>` which downloads a backup
  from S3 to the local backup directory, making it available for `pgmoneta-cli restore`.

  Closes #1010

  ## Design

  The restore is a two-phase download:

  1. **Bootstrap** (`s3_bootstrap`): Downloads root-level files (`backup.sha512`,
     `backup.info`, `backup.manifest`) and verifies `backup.info` integrity via SHA512
     checksum before proceeding with the bulk download.

  2. **Bulk download** (`s3_download_files`): Parses the CSV `backup.manifest` to get
     the file listing, loads `backup.info` to determine compression/encryption type,
     and downloads each data file with the correct extension (e.g. `.zstd`, `.aes`).
     `backup_label` and `backup_manifest` are excluded from extension appending as
     they are not compressed/encrypted by PostgreSQL.

  Downloads go to a temporary directory (`.pgmoneta_temp_<label>/`) which is atomically
  renamed on success, or cleaned up on error.

  ## Known limitations / future work

  - **No per-file checksum verification**: The CSV manifest checksums are for
    uncompressed files, but files on S3 are stored compressed. Verification during
    download is not possible with current data.
  - **Sequential single-file GETs**: Each file is downloaded one at a time. Performance
    can be improved with range-based downloads and parallel GETs.

### DIFF
--- a/doc/CLI.md
+++ b/doc/CLI.md
@@ -153,6 +153,31 @@ pgmoneta-cli restore primary newest time=2025-01-15\ 10:30:00 /tmp
 pgmoneta-cli restore primary newest timeline=2 /tmp
 ```
 
+## s3 restore
+
+Download a backup from S3 to the local backup directory
+
+This subcommand downloads a backup from S3 storage and makes it available locally. The flow is:
+
+* Download `backup.sha512`, `backup.info`, and `backup.manifest` from S3
+* Verify `backup.info` integrity using the SHA512 checksum
+* Download all data files listed in the manifest, applying the correct compression and encryption extensions
+* Atomically move the downloaded backup into the local backup directory
+
+After `s3 restore` completes, the backup appears in `status details` and can be restored using `pgmoneta-cli restore`.
+
+Command
+
+```sh
+pgmoneta-cli s3 restore <server> <timestamp>
+```
+
+Example
+
+```sh
+pgmoneta-cli s3 restore primary 20260316000957
+```
+
 ## verify
 
 Verify a backup from a server
@@ -487,12 +512,14 @@ Subcommand
 
 - `ls` : List all the files in s3
 - `delete` : Delete all files under a remote prefix in s3
+- `restore` : Download a backup from s3 to the local backup directory
 
 Example
 
 ```sh
 pgmoneta-cli s3 ls primary
 pgmoneta-cli s3 delete primary 20260302163357
+pgmoneta-cli s3 restore primary 20260316000957
 ```
 ### s3 ls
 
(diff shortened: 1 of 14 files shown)


jesperpedersen:
@Amr-Shams Did you check if you can use `struct workers` for the parallel download ?

Amr-Shams:
> @Amr-Shams Did you check if you can use `struct workers` for the parallel download ?

Yes, I looked into struct workers and it fits naturally here since each file download is independent, same pattern as gzip compression and copy directory. I plan to introduce a small struct s3_download_input with struct worker_common as the first member to carry the S3 path, server index, and range offsets through the task queue. Each worker would own its own HTTP connection and write its chunk independently. I'll add this as a follow up on top of this PR to keep the core download logic and verification separate from the parallelism layer.

jesperpedersen:
@Amr-Shams Sounds good !

jesperpedersen:
@Amr-Shams Then we likely can do the same for the backup upload

Amr-Shams:
> @Amr-Shams Then we likely can do the same for the backup upload

 100%, the upload/delete can benefit from the same worker pool approach. one additional random thought, i was looking to wire the progress bar to the s3 ops(highly needed somehow)

jesperpedersen:
@Amr-Shams Good idea with the progress bar - probably a separate issue that covers all the operations

Amr-Shams:
> @Amr-Shams Good idea with the progress bar - probably a separate issue that covers all the operations

Good idea, I'll open a separate issue for parallelism. It needs some careful thought around how to pass the S3 context through the worker input struct(even though i have this idea of extending the struct)

Amr-Shams:
@jesperpedersen , kept the original structure mostly intact but made better use of the workers' input setup. now it fully supports parallel downloads as an atomic operation, and it’s noticeably faster

jesperpedersen:
Rebased, and merged.

PR #1037, opened 2026-03-16
Fix #1035: Improve error logging in streamer pipeline

### LINKED ISSUE #1035: Improve error logging in the streamer pipeline
Currently, if an error occurs within the streamer pipeline (for example during compression, encryption, or while writing chunks to disk) in `src/libpgmoneta/stream.c`, the callbacks return `1`. However, there is no specific error logged to the end-user or the system logs explaining **why the failure occurred**.

There is also a `TODO` comment highlighting this gap in `pgmoneta_streamer_write()` on line 143:
```c
         if (streamer->stream_cb(streamer, last_chk))
         {
            //TODO: need to log what has failed at least
            goto error;
         }
```

**Describe the solution you'd like**
1. Add descriptive `pgmoneta_log_error()` calls inside `backup_stream_cb` and `restore_stream_cb` whenever `compress()`, `decompress()`, `encrypt()`, or `decrypt()` fails. 
2. Replace or update the `TODO` in `pgmoneta_streamer_write()` (line 143) to log a generic pipeline failure if `stream_cb` returns an error, ensuring at least *some* context is captured.

**Expected Outcome**

When a backup or restore operation fails during streaming, the logs should clearly indicate **which subsystem failed (compression, encryption, or I/O)**.
This will reduce the time required to diagnose and resolve streaming failures.

Sakshii-27:
Fixes #1035

### DIFF
--- a/src/include/vfile.h
+++ b/src/include/vfile.h
@@ -33,11 +33,25 @@
 extern "C" {
 #endif
 
+#include <pgmoneta.h>
+
+/** @enum vfile_type
+ * Defines the type of a virtual file
+ */
+enum vfile_type {
+   VFILE_TYPE_UNKNOWN = 0, /**< Unknown vfile type */
+   VFILE_TYPE_LOCAL,       /**< Local file system */
+   VFILE_TYPE_S3,          /**< S3 storage */
+};
+
 /** @struct vfile
  * Defines a virtual file
  */
 struct vfile
 {
+   enum vfile_type type; /**< The vfile type */
+   char name[MAX_PATH];  /**< The vfile name/path */
+
    /**
     * The read callback
     * @param vfile The vfile
@@ -82,6 +96,14 @@ struct vfile
  */
 int
 pgmoneta_vfile_create_local(char* file_path, char* mode, struct vfile** vfile);
+
+/**
+ * Get the string representation of a vfile type
+ * @param type The vfile type
+ * @return The string representation
+ */
+const char*
+pgmoneta_vfile_type_to_string(enum vfile_type type);
 
 /**
  * Close and destroy current vfile
--- a/src/libpgmoneta/stream.c
+++ b/src/libpgmoneta/stream.c
@@ -37,6 +37,7 @@
 #include <value.h>
 
 #include <stdlib.h>
+#include <string.h>
 
 static int noop_stream_cb(struct streamer* this, bool last_chunk);
 static int backup_stream_cb(struct streamer* this, bool last_chunk);
@@ -44,6 +45,7 @@ static int restore_stream_cb(struct streamer* this, bool last_chunk);
 static int get_backup_file_name_cb(struct streamer* this, char* file_name, char** dest_file_name);
 static int get_restore_file_name_cb(struct streamer* this, char* file_name, char** dest_file_name);
 static void vfile_destroy_cb(uintptr_t val);
+static void add_failed_destination(struct streamer* streamer, struct vfile* f);
 
 int
 pgmoneta_streamer_create(int mode, int encryption, int compression, struct streamer** streamer)
@@ -113,6 +115,7 @@ pgmoneta_streamer_destroy(struct streamer* streamer)
    pgmoneta_compressor_destroy(streamer->compressor);
    pgmoneta_encryptor_destroy(streamer->encryptor);
    pgmoneta_deque_destroy(streamer->destinations);
+   pgmoneta_deque_destroy(streamer->failed_destinations);
    free(streamer);
 }
 
@@ -140,7 +143,7 @@ pgmoneta_streamer_write(struct streamer* streamer, void* buffer, size_t size, bo
          /* process and send data when the buffer is full or we have no more data to send */
          if (streamer->stream_cb(streamer, last_chk))
          {
-            /* TODO: need to log what has failed at least */
+            pgmoneta_log_error("Streamer callback failed during execution");
             goto error;
          }
          /* update bytes written on success */
@@ -149,6 +152,18 @@ pgmoneta_streamer_write(struct streamer* streamer, void* buffer, size_t size, bo
       }
    }
    while (size > 0);
+
+   if (last_chunk && streamer->failed_destinations != NULL && !pgmoneta_deque_empty(streamer->failed_destinations))
+   {
+      struct deque_iterator* fail_iter = NULL;
+      pgmoneta_deque_iterator_create(streamer->failed_destinations, &fail_iter);
+      while (pgmoneta_deque_iterator_next(fail_iter))
+      {
+         char* desc = (char*)pgmoneta_value_data(fail_iter->value);
+         pgmoneta_log_warn("Streamer: destination failed during streaming: %s", desc);
+      }
+      pgmoneta_deque_iterator_destroy(fail_iter);
+   }
 
    return 0;
 error:
@@ -197,6 +212,7 @@ pgmoneta_streamer_reset(struct streamer* streamer)
    }
 
    pgmoneta_deque_clear(streamer->destinations);
+   pgmoneta_deque_clear(streamer->failed_destinations);
    streamer->size = 0;
    streamer->written = 0;
 }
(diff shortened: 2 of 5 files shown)


jesperpedersen:
@Jubilee101 PTAL

src/libpgmoneta/stream.c
```diff
@@ -140,7 +140,7 @@ pgmoneta_streamer_write(struct streamer* streamer, void* buffer, size_t size, bo
          // process and send data when the buffer is full or we have no more data to send
          if (streamer->stream_cb(streamer, last_chk))
          {
-            //TODO: need to log what has failed at least
+            pgmoneta_log_error("Streamer callback failed during execution");
```
Jubilee101, line 146: It's not like this :) In streamer we can have multiple destinations - local, S3, ssh... But when it fails we don't know which destination has failed. Currently we don't even return 1 when there's write failure (https://github.com/pgmoneta/pgmoneta/blob/e73b8887a22f3c5275d642a159b3daee82c7b024/src/libpgmoneta/stream.c#L219). The least we can do is to have clearer info as for which has failed exactly.

Sakshii-27: Hi @Jubilee101, thank you for the feedback!
I've updated the PR to add type and name to the vfile abstraction. Now, write failures will log the specific destination.

src/libpgmoneta/stream.c
```diff
@@ -218,7 +218,8 @@ noop_stream_cb(struct streamer* this, bool last_chunk)
       f = (struct vfile*)pgmoneta_value_data(vfile_iter->value);
       if (f->write(f, this->buffer, this->size, last_chunk))
       {
-         pgmoneta_log_error("Failed to write buffer");
+         pgmoneta_log_error("Streamer: Failed to write buffer to %s destination: %s", f->type != NULL ? f->type : "unknown", f->name != NULL ? f->name : "unknown");
+         goto error;
```
Jubilee101: It's not wise to go to error just because one destination failed. We should still continue with the rest, unless it's the sole destination. For now, I would remove this destination and continue. See `pgmoneta_deque_iterator_remove`

Jubilee101: Streamer should keep a list of failed destinations, in case in the future we can return that to user somehow.

src/libpgmoneta/vfile_local.c
```diff
@@ -68,6 +68,8 @@ pgmoneta_vfile_create_local(char* file_path, char* mode, struct vfile** vfile)
       goto error;
    }
    memcpy(file->file_path, file_path, strlen(file_path));
+   file->super.type = "local";
```
Jubilee101: This should be a constant in pgmoneta.h

src/libpgmoneta/vfile_local.c
```diff
@@ -68,6 +68,8 @@ pgmoneta_vfile_create_local(char* file_path, char* mode, struct vfile** vfile)
...
+   file->super.name = file->file_path;
```
Jubilee101: Don't share pointers, make a copy -- I would use inline buffer than malloc.

Jubilee101: You may want to have pgmoneta_vfile_set_metadata to wrap this.

Jubilee101 (changes requested):

Sakshii-27:
@Jubilee101 I have addressed the comments PTAL

jesperpedersen:
@Sakshii-27 Rebase

Sakshii-27:
done.

jesperpedersen:
@Sakshii-27 Can you rebase, please

src/libpgmoneta/stream.c
```diff
@@ -218,10 +221,23 @@ noop_stream_cb(struct streamer* this, bool last_chunk)
...
       if (f->write(f, this->buffer, this->size, last_chunk))
       {
-         pgmoneta_log_error("Failed to write buffer");
+         pgmoneta_log_error("Streamer: Failed to write buffer to %s destination: %s", strlen(f->type) > 0 ? f->type : "unknown", strlen(f->name) > 0 ? f->name : "unknown");
+         if (this->failed_destinations == NULL)
+         {
+            pgmoneta_deque_create(false, &this->failed_destinations);
+         }
+         pgmoneta_deque_add(this->failed_destinations, NULL, (uintptr_t)f, ValueRef);
+         pgmoneta_deque_iterator_remove(vfile_iter);
```
Jubilee101, line 238: This won't work. Because removing f from destination deque destroys f (it invokes the callback).

src/include/vfile.h
```diff
@@ -33,11 +33,16 @@
...
 #endif
 
+#include <pgmoneta.h>
+
 /** @struct vfile
  * Defines a virtual file
  */
 struct vfile
 {
+   char type[MISC_LENGTH]; /**< The vfile type (e.g. "local", "s3") */
```
Jubilee101: Type should be enum and gets translated to string by a util func when we need it.

src/include/vfile.h
```diff
@@ -83,6 +88,15 @@ struct vfile
...
 pgmoneta_vfile_create_local(char* file_path, char* mode, struct vfile** vfile);
 
+/**
+ * Set vfile metadata (type and name)
+ * @param vfile The vfile
+ * @param type The vfile type
+ * @param name The vfile name
+ */
+void
+pgmoneta_vfile_set_metadata(struct vfile* vfile, const char* type, const char* name);
```
Jubilee101: There's no need for this to be public. This should be done inside `vfile_create_local`. I don't even think there's a need to have this helper function.

src/libpgmoneta/stream.c
Jubilee101: I think for failed_destinations, storing a string of vfile type + destination is enough.

src/libpgmoneta/stream.c
```diff
@@ -269,12 +287,24 @@ backup_stream_cb(struct streamer* this, bool last_chunk)
...
+               pgmoneta_deque_create(false, &this->failed_destinations);
+            }
+            pgmoneta_deque_add(this->failed_destinations, NULL, (uintptr_t)f, ValueRef);
+            pgmoneta_deque_iterator_remove(vfile_iter);
          }
       }
       pgmoneta_deque_iterator_destroy(vfile_iter);
       vfile_iter = NULL;
 
+      if (pgmoneta_deque_empty(this->destinations))
```
Jubilee101, line 305: There should also be a warning if failed_destinations is not empty.

Jubilee101 (changes requested):

Jubilee101:
Hi @Sakshii-27 , are you still on this? Let's get this merged.

Sakshii-27:
@Jubilee101 Updated! All comments addressed.

src/libpgmoneta/stream.c
```diff
@@ -218,10 +233,29 @@ noop_stream_cb(struct streamer* this, bool last_chunk)
...
+         pgmoneta_log_error("Streamer: Failed to write buffer to %s destination: %s", type_str, name_str);
+         if (this->failed_destinations == NULL)
+         {
+            pgmoneta_deque_create(false, &this->failed_destinations);
+         }
+         char* desc = NULL;
+         desc = pgmoneta_append(desc, type_str);
+         desc = pgmoneta_append(desc, ": ");
+         desc = pgmoneta_append(desc, name_str);
+         pgmoneta_deque_add(this->failed_destinations, NULL, (uintptr_t)desc, ValueString);
```
Jubilee101: You need to free desc. deque_add copies the string, leaving the original one undestroyed. I'd suggest using a add_failed_dest helper function here.

Jubilee101: If you are interested, take a look at how the current core data structures (deque, ART) and type system work, that'll help you correctly work with them.

src/libpgmoneta/stream.c
```diff
@@ -269,12 +305,30 @@ backup_stream_cb(struct streamer* this, bool last_chunk)
          f = (struct vfile*)pgmoneta_value_data(vfile_iter->value);
          if (f->write(f, ebuf, ebuf_size, last_chunk))
          {
-            pgmoneta_log_error("Failed to write buffer");
+            const char* type_str = pgmoneta_vfile_type_to_string(f->type);
```
Jubilee101: Duplicated code, use a helper func.

src/libpgmoneta/stream.c
```diff
@@ -329,11 +385,29 @@ restore_stream_cb(struct streamer* this, bool last_chunk)
          f = (struct vfile*)pgmoneta_value_data(vfile_iter->value);
          if (f->write(f, cbuf, cbuf_size, last_chunk))
          {
-            pgmoneta_log_error("Failed to write buffer");
+            const char* type_str = pgmoneta_vfile_type_to_string(f->type);
```
Jubilee101: Same comment as above.

Jubilee101 (changes requested):
We are close.

Jubilee101:

PR #1038, opened 2026-03-16
[#1018] Refactored utils aes & compression

### LINKED ISSUE #1018: Compression: Workers
Compression / decompression functions in compression.h should optional support `struct workers`

sommye-ctr:
Resolves #1018 

### Changes
- Standardised compression/decompression flow. Removed duplicated codec specific directory/tablespace wrappers and updated callers to use the unified path in `compression.c`
- Moved functions related to file type detections from `utils.h` to `extraction.h`
- Moved compression and encryption related functions to their respective files
- Moved compression/extraction/AES tests to dedicated files


sommye-ctr:
@jesperpedersen I am unable to reproduce the CI error on my local machine...

jesperpedersen:
@sommye-ctr Are you fully rebased ?

jesperpedersen:
@sommye-ctr Rebase

src/include/extraction.h
```diff
@@ -36,6 +36,85 @@
...
+
+/**
+ * Normalize a file type bitmask.
+ * This ensures specific compression bits imply PGMONETA_FILE_TYPE_COMPRESSED.
+ *
+ * @param type The file type bitmask
+ * @return Normalized file type bitmask
+ */
+uint32_t
+pgmoneta_normalize_file_type(uint32_t type);
```
jesperpedersen: What do we need this for ?

src/include/extraction.h
```diff
@@ -36,6 +36,85 @@
...
+ *
+ * This helper maps COMPRESSION_* and ENCRYPTION_* settings to the matching
+ * PGMONETA_FILE_TYPE_* flags used by extraction helpers.
+ *
+ * @param compression_type The configured compression type
+ * @param encryption_type The configured encryption type
+ * @return File-type bitmask for configured suffix layers
+ */
+uint32_t
+pgmoneta_extraction_configured_file_type_mask(int compression_type, int encryption_type);
```
jesperpedersen: What do we need this for ?

src/include/extraction.h
```diff
@@ -36,6 +36,85 @@
...
+
+/**
+ * Strip extension suffix from a file name.
+ *
+ * @param s The source string
+ * @param name Resulting string without the last extension (caller must free)
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgmoneta_strip_extension(char* s, char** name);
```
jesperpedersen: This is a `utils.h` function

src/include/pgmoneta.h
```diff
@@ -131,6 +131,7 @@ extern "C" {
 #define COMPRESSION_SERVER_GZIP      (COMPRESSION_TYPE_SERVER | COMPRESSION_ALG_GZIP)
 #define COMPRESSION_SERVER_ZSTD      (COMPRESSION_TYPE_SERVER | COMPRESSION_ALG_ZSTD)
 #define COMPRESSION_SERVER_LZ4       (COMPRESSION_TYPE_SERVER | COMPRESSION_ALG_LZ4)
+#define COMPRESSION_SERVER_BZIP2     (COMPRESSION_TYPE_SERVER | COMPRESSION_ALG_BZIP2)
```
jesperpedersen: PostgreSQL doesn't support this

src/include/compression.h
```diff
@@ -31,6 +31,8 @@
 
 #include <pgmoneta.h>
 
+struct workers;
```
jesperpedersen, line 35: #include <workers.h>

src/include/workflow.h
```diff
@@ -40,6 +40,8 @@ extern "C" {
 #include <stdlib.h>
 #include <stdbool.h>
 
+struct workers;
```
jesperpedersen: #include <workers.h>

src/include/workflow.h
```diff
@@ -171,6 +173,20 @@ pgmoneta_common_setup(char* name, struct art* nodes);
...
+ * skips that directory and compresses the remaining immediate child directories.
+ *
+ * @param root The backup base root
+ * @param compression_type The compression type to use
+ * @param workers Optional worker pool
+ * @return 0 on success, otherwise 1
+ */
+int
+pgmoneta_workflow_compress_tablespaces(char* root, int compression_type, struct workers* workers);
+
```
jesperpedersen: Do we need one for encryption as well ?

src/libpgmoneta/walfile/wal_reader.c
```diff
@@ -33,6 +33,8 @@
 #include <compression.h>
 #include <json.h>
 #include <logging.h>
+#include <aes.h>
+#include <compression.h>
```
jesperpedersen: Sorting

src/libpgmoneta/walfile/wal_reader.c
```diff
@@ -209,7 +211,7 @@ pgmoneta_validate_wal_filename(char* path, char** base_filename, xlog_seg_no* se
       wal_filename = temp;
    }
 
-   if (pgmoneta_is_encrypted(wal_filename) || pgmoneta_compression_is_compressed(wal_filename))
+   if (pgmoneta_is_encrypted(wal_filename) || pgmoneta_is_compressed(wal_filename))
```
jesperpedersen: Use `pgmoneta_get_file_type` first

src/libpgmoneta/aes.c
```diff
@@ -312,22 +312,26 @@ pgmoneta_encrypt_wal(char* d)
...
-   const char* alg_suffix = NULL;
 
    config = (struct main_configuration*)shmem;
 
-   pgmoneta_compression_get_suffix(config->compression_type, &alg_suffix);
-   if (alg_suffix != NULL)
+   if (pgmoneta_extraction_get_suffix(config->compression_type, ENCRYPTION_NONE, &compress_suffix))
    {
-      compress_suffix = (char*)alg_suffix;
+      return 1;
```
jesperpedersen: `goto error;`

jesperpedersen:
@sommye-ctr Try and take these comments through the entire patch. PTAL me once it is sorted

src/include/workflow.h
sommye-ctr: `pgmoneta_encrypt_tablespaces` exists in aes.c

src/include/extraction.h
sommye-ctr: It was already being used, it is shifted from utils to extraction and now made into static (since it is only being used in extraction.c)

sommye-ctr:
@jesperpedersen PTAL

jesperpedersen:
@sommye-ctr We are moving in right direction.

The important part is that

* Compress
* Decompress
* Encrypt
* Decrypt

all take a `struct workers*` as their last parameter.

You also have to look at - f.ex. what is the difference between `pgmoneta_bzip2_wal_file` and `pgmoneta_bzip2_file`. We need as few public functions as possible.

So, `compression.h`, `aes.h` and `extraction.h` provides a simple API that other components can use.

```
   struct workers* workers = NULL;
   if (config->workers > 0)
   {
      /* Setup workers */
   }

   /* Do compression */
   if (workers != NULL)
   {
       /* Workers.wait */
   }

   /* Do encryption */
   if (workers != NULL)
   {
       /* Workers.wait */
   }
```

Yes, in order for this to work in certain cases we have to pass in file extensions that should be skipped - could be a `struct deque*` or a string array...

jesperpedersen:
@sommye-ctr We should expand the DEBUG logging around main.c:926 with all vault sizes, and the number of workers

jesperpedersen:
@sommye-ctr During development set workers to > 0, and make sure that all code paths are using the workers

jesperpedersen:
@sommye-ctr Are you ok on this ? Its a high-priority PR

sommye-ctr:
> @sommye-ctr Are you ok on this ? Its a high-priority PR

Yes, I'll push changes till tomorrow

sommye-ctr:
I have removed the codec specific wal functions and rather used the generic functions from aes & compression. Also added workers in all functions & cleaned up the public functions.

src/include/workflow.h
```diff
@@ -172,6 +175,20 @@ pgmoneta_common_setup(char* name, struct art* nodes);
...
+ * The backup base root may contain a "data" directory that is handled separately; this helper
+ * skips that directory and compresses the remaining immediate child directories.
+ *
+ * @param root The backup base root
+ * @param compression_type The compression type to use
+ * @param workers Optional worker pool
+ * @return 0 on success, otherwise 1
+ */
+int
+pgmoneta_workflow_compress_tablespaces(char* root, int compression_type, struct workers* workers);
```
jesperpedersen: This should be in `compression.h`

jesperpedersen:
@sommye-ctr Look into if we actually need tablespace specific functions now - is _directory enough ? Maybe we need an iterator on top...

sommye-ctr:
> @sommye-ctr Look into if we actually need tablespace specific functions now - is _directory enough ? Maybe we need an iterator on top...

We could do that, currently we have `pgmoneta_encrypt_data` which is called by `pgmoneta_encrypt_directory` and `pgmoneta_encrypt_tablespaces`. So the callers of tablespace function would need to pass in excludes list.

jesperpedersen:
@sommye-ctr We should see if we can get to

* _encrypt_file
* _encrypt_directory
* _decrypt_file
* _decrypt_directory
* _compress_file
* _compress_directory
* _decompress_file
* _decompress_directory

as the core API

sommye-ctr:
Removed both the encrypt and compress tablespaces function.

jesperpedersen:
@sommye-ctr CI

sommye-ctr:
@jesperpedersen I think the issue is caused by that I had replaced `pgmoneta_compress_tablespace(backup_base...)` and `pgmoneta_compress_directory(backup_data....)` with a single call of `pgmoneta_compress_directory(backup_base...)`.

Now there are some exclusions of metadata files by default in directory function but some are missed which cause them to be compressed and then later on the workflow tries to read one of the files (which are compressed) and is unable to read.

So to solve this we can pass in the exclude list with a set of meta files in the backup_base, here is the list of files I think that need to be skipped:
```
backup.manifest
backup.info
backup.sha*
backup_label (already does)
backup_manifest (already does)
```
Are there any other files I am missing that needs to be skipped for compressing/encryption?

jesperpedersen:
@sommye-ctr That should be it

src/include/aes.h
```diff
@@ -155,30 +156,24 @@ void
...
- * Encrypt the files under the tablespace directories in place recursively, also remove unencrypted files.
+ * Decrypt the tablespaces
  * @param root The root directory
  * @param workers The optional workers
+ * @param excludes The optional excludes
  * @return 0 upon success, otherwise 1
  */
 int
-pgmoneta_encrypt_tablespaces(char* root, struct workers* workers);
+pgmoneta_decrypt_tablespaces(char* root, struct workers* workers, struct deque* excludes);
```
jesperpedersen: We can remove _tablespaces from aes.h

src/include/workflow.h
```diff
@@ -36,10 +36,13 @@ extern "C" {
 #include <pgmoneta.h>
 #include <art.h>
 #include <info.h>
+#include <workers.h>
 
 #include <stdlib.h>
 #include <stdbool.h>
 
+struct workers;
```
jesperpedersen: You don't need this

jesperpedersen:
@sommye-ctr We are getting really close now - then I'll do the .c files but they should be ok, if the .h's are

src/include/aes.h
sommye-ctr: Must have missed it, fixed

jesperpedersen:
@sommye-ctr So, we need to move some call-sites, like

```
   if (workers != NULL)
   {
      pgmoneta_workers_wait(workers);
   }
```

from `aes.c` and `compression.c`. 

The wait / outcome checking has to be done in the clients of those APIs - see `restore.c` as a workers example

jesperpedersen:
Merged.

PR #1039, opened 2026-03-16
Fix memory leak in verify command error path

Sakshii-27:
The pgmoneta_verify() function creates three local JSON objects (failed, all, filesj) which are eventually transferred into the response payload tree via pgmoneta_json_put(). However, if an error occurs before the ownership transfer completes, these objects are leaked - the error path only called pgmoneta_json_destroy(payload), which does not cover the local objects that haven't been linked into the payload tree yet.

Fix:
- Null out local pointers (failed, all, filesj) immediately after pgmoneta_json_put() transfers their ownership, preventing double-frees.
- Add pgmoneta_json_destroy() calls for these three pointers in the error cleanup block. Since pgmoneta_json_destroy(NULL) is a no-op, this is safe regardless of which error path was taken.

Related to: [pgmoneta-mcp/pgmoneta-mcp#49](https://github.com/pgmoneta/pgmoneta_mcp/issues/49)

### DIFF
--- a/src/libpgmoneta/verify.c
+++ b/src/libpgmoneta/verify.c
@@ -238,11 +238,14 @@ pgmoneta_verify(SSL* ssl, int client_fd, int server, uint8_t compression, uint8_
    }
 
    pgmoneta_json_put(filesj, MANAGEMENT_ARGUMENT_FAILED, (uintptr_t)failed, ValueJSON);
+   failed = NULL;
    pgmoneta_json_put(filesj, MANAGEMENT_ARGUMENT_ALL, (uintptr_t)all, ValueJSON);
+   all = NULL;
 
    pgmoneta_json_put(response, MANAGEMENT_ARGUMENT_BACKUP, (uintptr_t)label, ValueString);
    pgmoneta_json_put(response, MANAGEMENT_ARGUMENT_SERVER, (uintptr_t)config->common.servers[server].name, ValueString);
    pgmoneta_json_put(response, MANAGEMENT_ARGUMENT_FILES, (uintptr_t)filesj, ValueJSON);
+   filesj = NULL;
 
    pgmoneta_delete_directory((char*)pgmoneta_art_search(nodes, NODE_TARGET_BASE));
 
@@ -290,6 +293,10 @@ error:
    pgmoneta_deque_iterator_destroy(aiter);
 
    pgmoneta_art_destroy(nodes);
+
+   pgmoneta_json_destroy(filesj);
+   pgmoneta_json_destroy(failed);
+   pgmoneta_json_destroy(all);
 
    pgmoneta_json_destroy(payload);
 
--- a/src/libpgmoneta/wf_restore.c
+++ b/src/libpgmoneta/wf_restore.c
@@ -810,8 +810,6 @@ restore_excluded_files_execute(char* name __attribute__((unused)), struct art* n
    char* to = NULL;
    char* suffix = NULL;
    struct backup* backup = NULL;
-   struct workers* workers = NULL;
-   int number_of_workers = 0;
    char** restore_last_files_names = NULL;
    struct main_configuration* config = (struct main_configuration*)shmem;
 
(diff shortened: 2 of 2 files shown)


jesperpedersen:
@Sakshii-27 You can remove all the comments

jesperpedersen:
@Sakshii-27 While you are on the right track you have to investigate the root cause of the failure. A backup just taken is VERY unlikely to have a verity error. The hash had to change, or the disk broke, or something... Fixing the memory leak is ok, but that isn't the real issue

jesperpedersen:
@Sakshii-27 Related #1040

sksingh2005:
@Sakshii-27 The leak fix is fine, but it does not explain the real failure.
A freshly taken backup is very unlikely to fail verify because of a genuine hash mismatch. The more likely issue is in the verify workflow itself i think:`restore -> decrypt -> decompress -> restore_excluded_files -> verify`

`restore_excluded_files()` copies excluded files like `global/pg_control`, `postgresql.conf`, and `pg_hba.conf` back into the target tree with their backup suffixes. For compressed/encrypted backups, this can reintroduce encoded files *after* decrypt/decompress has already finished.
That means verify may be hashing the wrong final bytes, by which a fresh backup can fail verification.
So i think the workflow ordering likely needs to be fixed, not just the memory leak. 
cc: @jesperpedersen

jesperpedersen:
@sksingh2005 We need to make sure that those 3 files are the last 3 files worked on

jesperpedersen:
We need a way that they are always processed last in each phase - for restore and verify

Sakshii-27:
Hi @jesperpedersen,
I’ve been looking into the `wf_verify` workflow, just wanted to confirm with you.
Tracing through the verify pipeline  `restore → decrypt/decompress → restore_excluded_files → verify` 

The issue occurs during the `restore_excluded_files` step. At this stage, the three excluded files:

* `pg_control`
* `postgresql.conf`
* `pg_hba.conf`
are copied directly from the backup source into the target directory, but they retain their compression/encryption suffixes (e.g., `.gz`).

Since this happens **after** the decompress step, these files remain in the target directory in their compressed/encrypted form.When the `verify` step runs, it reads from `backup.manifest`, which expects the plain filenames (e.g., `postgresql.conf`). Because those files are still present as `.gz`, the verify step cannot find the expected files and immediately marks them as failed.

You mentioned earlier the need to process these files last in each phase, which makes sense. Would the right approach be to add a step that specifically decrypts/decompresses these three files immediately after `restore_excluded_files` runs?

sksingh2005:
@Sakshii-27 I think those 3 files need special handling after restore/decode, and they need to be the last files processed in both restore and verify.Rather than adding a separate extra decrypt/decompress step after `restore_excluded_files()`, I think the cleaner fix ican be to make `restore_excluded_files()` itself restore `global/pg_control`, `postgresql.conf`, and `pg_hba.conf` as their final decoded files, and then ensure verify also processes those same 3 entries last.
cc: @jesperpedersen

Sakshii-27:
Thanks @sksingh2005! Yes, that's exactly the approach I'm taking, modifying [restore_excluded_files()](cci:1://file:///Users/sakshii/pgmoneta/src/libpgmoneta/wf_restore.c:802:0-806:1) itself to decrypt and decompress the 3 files right after copying them, rather than adding a separate workflow step. Since [restore_excluded_files](cci:1://file:///Users/sakshii/pgmoneta/src/libpgmoneta/wf_restore.c:802:0-806:1) already runs before [verify](cci:1://file:///Users/sakshii/pgmoneta/src/libpgmoneta/wf_verify.c:224:0-304:1) in the pipeline, this ensures those files are always fully decoded by the time verification runs.

jesperpedersen:
@sksingh2005 @Sakshii-27 Correct

Sakshii-27:
@jesperpedersen I’ve updated the PR to address both the memory leak and the root cause of the verification failure 
Let me know if everything looks good now!

jesperpedersen:
@Sakshii-27 You need to use `compression.h` and/or `extraction.h`

Sakshii-27:
@jesperpedersen Done

jesperpedersen:
Rebased, and merged.

jesperpedersen:
@Sakshii-27 Remember to reference the issue number in the commit message

PR #1052, opened 2026-03-22
[#1013] Add finalize workflow to prevent race conditions

### LINKED ISSUE #1013: Backup race condition: pgmoneta_backup releases repository lock too late
**Describe the bug**

Based on https://github.com/pgmoneta/pgmoneta/pull/1012#issuecomment-4050293129

There is a race condition in `pgmoneta_backup` caused by the order of operations at the end of the backup process. The server sends a success response to `pgmoneta-cli` before it has finished its background WAL compression/encryption tasks, and importantly, before it has released the server lock.

This causes a problem for automation scripts (like benchmark) that run back-to-back backups. Since the CLI returns success immediately, the script assumes the server is ready for the next operation, but the server will incorrectly reject the next backup request with Server primary is active.

Abdelrhmansersawy:
resolved #1013 

## Changes
1. Added a `Finalize` workflow step to guarantee repository locks are released.
2. Replaced shared global variables with ART nodes to pass data directly and prevent race conditions.
3. Centralized client responses to ensure they are sent only after cleanup finishes.

### DIFF
--- a/src/include/workflow.h
+++ b/src/include/workflow.h
@@ -62,9 +62,19 @@ extern "C" {
 
 #define CLEANUP_TYPE_RESTORE             0
 
+#define FINALIZE_TYPE_BACKUP             0
+#define FINALIZE_TYPE_RESTORE            1
+#define FINALIZE_TYPE_VERIFY             2
+
+#define WORKFLOW_RESULT_SUCCESS          0
+#define WORKFLOW_RESULT_ERROR            1
+#define WORKFLOW_RESULT_FATAL            2
+
 #define NODE_ALL                         "all"                 /* All the files in a manifest */
 #define NODE_BACKUP                      "backup"              /* The backup structure */
 #define NODE_COMBINE_AS_IS               "combine_as_is"       /* Whether to combine the backups as is*/
+#define NODE_FINALIZE_TYPE               "finalize_type"       /* The finalize type */
+#define NODE_FINALIZE_VALUE              "finalize_value"      /* The finalize value (locked state) */
 #define NODE_COPY_WAL                    "copy_wal"            /* Whether to copy WAL */
 #define NODE_BACKUP_BASE                 "backup_base"         /* The base directory of the backup */
 #define NODE_BACKUP_DATA                 "backup_data"         /* The data directory of the backup */
@@ -86,6 +96,12 @@ extern "C" {
 #define NODE_TARGET_BASE                 "target_base"         /* The target base directory */
 #define NODE_TARGET_FILE                 "target_file"         /* The target file */
 #define NODE_TARGET_ROOT                 "target_root"         /* The target root directory */
+#define NODE_CLIENT_FD                   "client_fd"           /* The client fd */
+#define NODE_SSL                         "ssl"                 /* The SSL connection */
+#define NODE_PAYLOAD                     "payload"             /* The JSON payload */
+#define NODE_COMPRESSION                 "compression"         /* The compression type */
+#define NODE_ENCRYPTION                  "encryption"          /* The encryption type */
+#define NODE_START_TIME                  "start_time"          /* The start time */
 
 /* Supplied by the user */
 #define USER_DIRECTORY  "directory"  /* The target root directory */
--- a/src/include/workflow_funcs.h
+++ b/src/include/workflow_funcs.h
@@ -202,6 +202,14 @@ pgmoneta_create_manifest(void);
  */
 struct workflow*
 pgmoneta_create_extra(void);
+
+/**
+ * Create a workflow for the finalize step
+ * @return The workflow
+ */
+struct workflow*
+pgmoneta_create_finalize(void);
+
 #ifdef __cplusplus
 }
 #endif
--- a/src/libpgmoneta/backup.c
+++ b/src/libpgmoneta/backup.c
@@ -38,32 +38,27 @@
 #include <network.h>
 #include <security.h>
 #include <utils.h>
-#include <wal.h>
 #include <workflow.h>
+#include <workflow_funcs.h>
 
 #define NAME "backup"
 
 void
 pgmoneta_backup(int client_fd, int server, uint8_t compression, uint8_t encryption, struct json* payload)
 {
-   bool active = false;
    char date_str[128];
    char* date = NULL;
-   char* elapsed = NULL;
    char* incremental = NULL;
    char* incremental_base = NULL;
    struct tm* time_info;
    struct timespec start_t;
-   struct timespec end_t;
    time_t curr_t;
-   double total_seconds;
    char* en = NULL;
    int ec = -1;
    int backup_index = -1;
    char* server_backup = NULL;
    char* root = NULL;
    char* d = NULL;
-   char* backup_data = NULL;
    bool backup_incremental = false;
    int number_of_backups = 0;
    struct backup** backups = NULL;
@@ -72,7 +67,6 @@ pgmoneta_backup(int client_fd, int server, uint8_t compression, uint8_t encrypti
    struct backup* backup = NULL;
    struct backup* child = NULL;
    struct json* req = NULL;
-   struct json* response = NULL;
    struct main_configuration* config;
 
    pgmoneta_start_logging();
@@ -92,21 +86,6 @@ pgmoneta_backup(int client_fd, int server, uint8_t compression, uint8_t encrypti
       pgmoneta_log_error("Backup: Server %s is not WAL streaming", config->common.servers[server].name);
       goto error;
    }
-
-   if (!atomic_compare_exchange_strong(&config->common.servers[server].repository, &active, true))
-   {
-      ec = MANAGEMENT_ERROR_BACKUP_ACTIVE;
-      pgmoneta_log_info("Backup: Server %s is active", config->common.servers[server].name);
-      pgmoneta_log_debug("Backup=%s, Restore=%s, Archive=%s, Delete=%s, Retention=%s",
-                         config->common.servers[server].active_backup ? "Yes" : "No",
-                         config->common.servers[server].active_restore ? "Yes" : "No",
-                         config->common.servers[server].active_archive ? "Yes" : "No",
-                         config->common.servers[server].active_delete ? "Yes" : "No",
-                         config->common.servers[server].active_retention ? "Yes" : "No");
-      goto error;
-   }
-
-   config->common.servers[server].active_backup = true;
 
 #ifdef HAVE_FREEBSD
    clock_gettime(CLOCK_MONOTONIC_FAST, &start_t);
@@ -129,13 +108,18 @@ pgmoneta_backup(int client_fd, int server, uint8_t compression, uint8_t encrypti
       goto error;
    }
 
+   pgmoneta_art_insert(nodes, NODE_FINALIZE_TYPE, (uintptr_t)FINALIZE_TYPE_BACKUP, ValueInt32);
+   pgmoneta_art_insert(nodes, NODE_CLIENT_FD, (uintptr_t)client_fd, ValueInt32);
+   pgmoneta_art_insert(nodes, NODE_PAYLOAD, (uintptr_t)payload, ValueRef);
+   pgmoneta_art_insert(nodes, NODE_COMPRESSION, (uintptr_t)compression, ValueUInt8);
+   pgmoneta_art_insert(nodes, NODE_ENCRYPTION, (uintptr_t)encryption, ValueUInt8);
+   pgmoneta_art_insert(nodes, NODE_START_TIME, (uintptr_t)&start_t, ValueRef);
+
    if (pgmoneta_workflow_nodes(server, date, nodes, &backup))
    {
       goto error;
    }
 
-   backup_data = (char*)pgmoneta_art_search(nodes, NODE_BACKUP_DATA);
-   server_backup = (char*)pgmoneta_art_search(nodes, NODE_SERVER_BACKUP);
    root = pgmoneta_get_server_backup_identifier(server, date);
 
    if (incremental != NULL)
@@ -155,6 +139,8 @@ pgmoneta_backup(int client_fd, int server, uint8_t compression, uint8_t encrypti
 
    if (backup_incremental)
    {
+      server_backup = (char*)pgmoneta_art_search(nodes, NODE_SERVER_BACKUP);
+
       if (pgmoneta_load_infos(server_backup, &number_of_backups, &backups))
       {
          ec = MANAGEMENT_ERROR_BACKUP_NOBACKUPS;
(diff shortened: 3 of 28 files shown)


Abdelrhmansersawy:
@jesperpedersen 
Could you give this a high-level review to check if I’m heading in the right direction?
I’m working on ensuring that the finalizer workflow does not affect the normal behavior of any workflows.

jesperpedersen:
@Abdelrhmansersawy Yes, this is the right direction

src/libpgmoneta/wf_finalize.c
```diff
@@ -0,0 +1,141 @@
...
+#include <pgmoneta.h>
+#include <logging.h>
+#include <management.h>
+#include <workflow.h>
+
+/* system */
+#include <stdatomic.h>
+#include <stdlib.h>
+
+static int finalize_type = -1;
```
Amr-Shams: it is a static global in file scope so calling pgmoneta_create_finalize() multiple times (backup, restore, verify) overwrites it each time. The finalize step will always use whatever type was set last. i would go and define the type per instance not the global

jesperpedersen: @Amr-Shams You can expand `struct server` to have the information needed

jesperpedersen: @Abdelrhmansersawy Cc:

Abdelrhmansersawy: @Amr-Shams 
Good Catch! 
I think we can do as @jesperpedersen suggested.

Abdelrhmansersawy:
A few updates:
1. Backup needs the repo lock held through WAL compress, so finalize can't be a workflow step - make it a utility function callers invoke before the workflow response is sent.
2. Verify workflow doesn't do a repositry lock, so we don't need to make finialize for it.

jesperpedersen:
1. There is encryption as well... I don't see why we can't set the flag based on a node value in `art*` in teardown
2. That sounds like a bug

Abdelrhmansersawy:
> 2. That sounds like a bug

No, I mean repo lock release don't happen in `pgmoneta_verify` but in `pgmoneta_sha512_verification`.
https://github.com/pgmoneta/pgmoneta/blob/414e90862faac863d3f0aac049feafaacc3c4051/src/libpgmoneta/verify.c#L541-L546

For consistent I think I can move this to replace finialize...

Abdelrhmansersawy:
@jesperpedersen PTAL

src/include/workflow.h
```diff
@@ -62,9 +62,14 @@ extern "C" {
...
 #define CLEANUP_TYPE_RESTORE             0
 
+#define FINALIZE_TYPE_BACKUP             0
+#define FINALIZE_TYPE_RESTORE            1
+#define FINALIZE_TYPE_VERIFY             2
+
 #define NODE_ALL                         "all"                 /* All the files in a manifest */
 #define NODE_BACKUP                      "backup"              /* The backup structure */
 #define NODE_COMBINE_AS_IS               "combine_as_is"       /* Whether to combine the backups as is*/
+#define NODE_FINALIZE_TYPE               "finalize_type"       /* The finalize type */
```
jesperpedersen, line 76: We need a `NODE_FINALIZE_VALUE` too

src/include/workflow_funcs.h
```diff
@@ -202,6 +202,15 @@ pgmoneta_create_manifest(void);
...
 struct workflow*
 pgmoneta_create_extra(void);
+
+/**
+ * Release the server lock based on the finalize type
+ * @param server The server index
+ * @param type The finalize type
+ */
+void
+pgmoneta_finalize_release(int server, int type);
```
jesperpedersen: This need to be a `struct workflow*` instead

src/libpgmoneta/backup.c
```diff
@@ -107,6 +109,7 @@ pgmoneta_backup(int client_fd, int server, uint8_t compression, uint8_t encrypti
    }
 
    config->common.servers[server].active_backup = true;
+   locked = true;
```
jesperpedersen: `locked` need to go into the finalize's setup

src/libpgmoneta/backup.c
```diff
@@ -129,6 +132,8 @@ pgmoneta_backup(int client_fd, int server, uint8_t compression, uint8_t encrypti
       goto error;
    }
 
+   pgmoneta_art_insert(nodes, NODE_FINALIZE_TYPE, (uintptr_t)FINALIZE_TYPE_BACKUP, ValueInt32);
```
jesperpedersen, line 111: Same

src/libpgmoneta/backup.c
```diff
@@ -271,6 +276,13 @@ pgmoneta_backup(int client_fd, int server, uint8_t compression, uint8_t encrypti
       ec = MANAGEMENT_ERROR_BACKUP_ERROR;
       goto error;
    }
+   pgmoneta_log_info("Backup: %s/%s (Elapsed: %s)", config->common.servers[server].name, date, elapsed);
+
+   pgmoneta_wal_server_compress_encrypt(server, NULL, NULL);
```
jesperpedersen: This looks like a bug as well - it need to be in the compression and encryption workflows

src/libpgmoneta/backup.c
```diff
@@ -271,6 +276,13 @@ pgmoneta_backup(int client_fd, int server, uint8_t compression, uint8_t encrypti
...
+
+   pgmoneta_finalize_release(server, (int)pgmoneta_art_search(nodes, NODE_FINALIZE_TYPE));
```
jesperpedersen: This should be in the finalize's teardown

src/libpgmoneta/backup.c
```diff
@@ -312,8 +317,10 @@ pgmoneta_backup(int client_fd, int server, uint8_t compression, uint8_t encrypti
 
 error:
 
-   config->common.servers[server].active_backup = false;
-   atomic_store(&config->common.servers[server].repository, false);
+   if (locked)
```
jesperpedersen: Same

jesperpedersen (changes requested):

jesperpedersen:
@Abdelrhmansersawy The comments for backup applies for the others as well

src/libpgmoneta/backup.c
Abdelrhmansersawy: @jesperpedersen 
Adding `pgmoneta_wal_server_compress_encrypt()` inside the compression execute functions will cause it to also run in `wf_post_rollup()` and `wf_archive()`, since they share the same compression step.

Is that intended? I’m not familiar with these, just wanted to ask.

jesperpedersen: @Abdelrhmansersawy There should be one compression step, and one encryption step. They should not be hidden inside other workflows
@sommye-ctr Cc: - feedback

jesperpedersen:
@Abdelrhmansersawy I think this depend on #1038 ...
@sommye-ctr Do you agree ?

Abdelrhmansersawy:
@jesperpedersen 
Do u see it's better to stop now, return work on it after https://github.com/pgmoneta/pgmoneta/pull/1038 is merged?

Or skip wal_compression_encryption part for now (as it's not PR focused) and open new PR later.

jesperpedersen:
@Abdelrhmansersawy Just leave it for now - we will circle back soon as #1038 is close

sommye-ctr:
Yes I think we can have this on hold till #1038 is settled

src/libpgmoneta/backup.c
Abdelrhmansersawy: @jesperpedersen 
Now, what's the correct place to put wal_compression _encryption

jesperpedersen: @Abdelrhmansersawy Nuke it - we need to use aes.h and compression.h only. Everything else that touches those areas need to be rewritten

Abdelrhmansersawy:
@jesperpedersen PTAL

src/libpgmoneta/backup.c
```diff
@@ -312,8 +292,15 @@ pgmoneta_backup(int client_fd, int server, uint8_t compression, uint8_t encrypti
 
 error:
 
-   config->common.servers[server].active_backup = false;
-   atomic_store(&config->common.servers[server].repository, false);
+   if (nodes != NULL && workflow != NULL)
```
jesperpedersen: Eh, what ?

src/libpgmoneta/restore.c
```diff
@@ -690,6 +672,16 @@ pgmoneta_restore(SSL* ssl, int client_fd, int server, uint8_t compression, uint8
 
 error:
 
+   if (nodes != NULL && (bool)pgmoneta_art_search(nodes, NODE_FINALIZE_VALUE))
```
jesperpedersen: Eh, what ?

src/libpgmoneta/verify.c
```diff
@@ -289,6 +300,20 @@ pgmoneta_verify(SSL* ssl, int client_fd, int server, uint8_t compression, uint8_
 
    pgmoneta_delete_directory((char*)pgmoneta_art_search(nodes, NODE_TARGET_BASE));
 
+   if (nodes != NULL && workflow != NULL)
```
jesperpedersen: Again...

jesperpedersen (changes requested):

jesperpedersen:
@Abdelrhmansersawy You need to start the backup/restore/verify workflow with the finalizer... Then in each workflow check that you have the lock. And in finalizer's teardown send the recorded result.

There are no loops inside a workflow node - that is handled by workflow.c

Abdelrhmansersawy:
@jesperpedersen 

1. Right now `pgmoneta_workflow_execute()` doesn’t run teardowns if a setup or execute step fails - it just returns `1`. If the finalizer runs first and acquires the lock in its setup, then a later step fails, the lock would never be released. Should I update `pgmoneta_workflow_execute()` to always run teardowns on error?

2. About “sending the recorded result in the finalizer’s teardown” - currently the response is built after the workflow finishes. For example in `backup.c`, things like `backup_size` are computed after `pgmoneta_workflow_execute()` returns, and `client_fd` / payload aren’t stored in the ART nodes. Should I pass these through the nodes so the finalizer can handle it, or did you have something else in mind?

jesperpedersen:
@Abdelrhmansersawy 

1. Yes, we need to change the workflow - we need a way to identify required workflow success (wf_finalizer), so if that fails an `INFO` (or `WARN`) is logged. The rest are logged as `DEBUG` for the backup, restore and verify flow. So, maybe 0, 1, 2... Of course that means that we have make guards in setup/execute/teardown in order to support this
2. For backup/restore/verify - stick everything you need in the ART and do it in wf_finalizer.c::teardown - remember to make those nodes mandatory as an assert

Abdelrhmansersawy:
@jesperpedersen 
If we move `finalize` to the start of the workflow chain, its teardown will actually run *first*, since `pgmoneta_workflow_execute()` processes teardowns in forward order.

That means the repository lock could be released too early - before the rest of the steps finish cleaning up. In that case, a new backup might grab the lock while the previous one is still freeing memory or removing files.

Do you think it would be better to run teardowns in reverse order (LIFO) inside `pgmoneta_workflow_execute()`, so the lock is always released last?

jesperpedersen:
@Abdelrhmansersawy Yes, it sounds correct to run setup in order, and teardown in reserve. Of course we have to go through the current teardown implementation will work with that

Abdelrhmansersawy:
@jesperpedersen PTAL

A few follow-up:
1. Moved client responses into `wf_finalize::teardown` - just stick what you need in the ART + mandatory assert.
2. Added `WORKFLOW_RESULT_FATAL` and `WORKFLOW_RESULT_ERROR` so failed workflows log at the right level.
3. Made sure teardowns always run, even when something fails.
4. Put the Finalize step at the head of the chain.
5. Made teardowns run in reverse order of setup - as `wf_finalize::teardown` run as the last step.
6. Added a guard in `pgmoneta_common_setup` to skip if finalize is part of the workflow but the lock was not acquired.

src/include/workflow.h
```diff
@@ -86,6 +96,12 @@ extern "C" {
 #define NODE_TARGET_BASE                 "target_base"         /* The target base directory */
 #define NODE_TARGET_FILE                 "target_file"         /* The target file */
 #define NODE_TARGET_ROOT                 "target_root"         /* The target root directory */
+#define NODE_CLIENT_FD                   "client_fd"           /* The client fd */
```
jesperpedersen, line 99: You need to sort these

src/libpgmoneta/backup.c
```diff
@@ -312,9 +237,6 @@ pgmoneta_backup(int client_fd, int server, uint8_t compression, uint8_t encrypti
 
 error:
 
-   config->common.servers[server].active_backup = false;
-   atomic_store(&config->common.servers[server].repository, false);
-
    pgmoneta_management_response_error(NULL, client_fd, config->common.servers[server].name,
```
jesperpedersen, line 240: This should be moved, right ?

src/libpgmoneta/restore.c
```diff
@@ -38,6 +38,7 @@
 #include <utils.h>
 #include <workers.h>
 #include <workflow.h>
+#include <workflow_funcs.h>
```
jesperpedersen, line 41: Why ?

src/libpgmoneta/restore.c
```diff
@@ -1117,7 +1060,7 @@ pgmoneta_rollup_backups(int server, char* newest_label, char* oldest_label)
    }
 
    workflow = pgmoneta_workflow_create(WORKFLOW_TYPE_POST_ROLLUP, tmp_backup);
-   if (carry_out_workflow(workflow, nodes) != RESTORE_OK)
+   if (pgmoneta_workflow_execute(workflow, nodes, &en, &ec))
```
jesperpedersen, line 1063: This looks wrong

src/libpgmoneta/restore.c
```diff
@@ -2479,8 +2424,9 @@ restore_backup_full(struct art* nodes)
    }
 
    workflow = pgmoneta_workflow_create(WORKFLOW_TYPE_RESTORE, backup);
-   if ((ret = carry_out_workflow(workflow, nodes) != RESTORE_OK))
+   if (pgmoneta_workflow_execute(workflow, nodes, &en, &ec))
```
jesperpedersen, line 2427: And this

src/libpgmoneta/restore.c
```diff
@@ -2615,8 +2563,9 @@ restore_backup_incremental(struct art* nodes)
       workflow = pgmoneta_workflow_create(WORKFLOW_TYPE_COMBINE_AS_IS, backup);
    }
 
-   if ((ret = carry_out_workflow(workflow, nodes) != RESTORE_OK))
+   if (pgmoneta_workflow_execute(workflow, nodes, &en, &ec))
```
jesperpedersen, line 2566: And this

src/libpgmoneta/verify.c
```diff
@@ -36,9 +36,11 @@
 #include <security.h>
 #include <utils.h>
 #include <workflow.h>
+#include <workflow_funcs.h>
```
jesperpedersen, line 39: Why ?

src/libpgmoneta/verify.c
```diff
@@ -144,132 +143,13 @@ pgmoneta_verify(SSL* ssl, int client_fd, int server, uint8_t compression, uint8_
...
-         {
-            goto error;
-         }
-
-         pgmoneta_json_append(all, (uintptr_t)j, ValueJSON);
-      }
-   }
-
-   if (pgmoneta_management_create_response(payload, server, &response))
+   if (pgmoneta_workflow_execute(workflow, nodes, &en, &ec))
```
jesperpedersen, line 146: This looks wrong as well...

src/libpgmoneta/verify.c
```diff
@@ -281,22 +161,16 @@ pgmoneta_verify(SSL* ssl, int client_fd, int server, uint8_t compression, uint8_
...
 
    free(real_directory);
-   free(elapsed);
 
    exit(0);
 
 error:
 
-   pgmoneta_delete_directory((char*)pgmoneta_art_search(nodes, NODE_TARGET_BASE));
+   pgmoneta_management_response_error(ssl, client_fd, config->common.servers[server].name,
```
jesperpedersen, line 169: This is wrong - its finalizer::teardown

jesperpedersen (changes requested):

src/libpgmoneta/backup.c
Abdelrhmansersawy: @jesperpedersen 
The problem that there are two types of errors that exist:

1. Early errors - things like invalid server or WAL streaming not enabled. These happen before `pgmoneta_workflow_execute` runs, so `finalize_teardown` never gets called. The error response in `backup.c` is needed here.

2. Workflow errors - when something fails inside `pgmoneta_workflow_execute`, it already calls `finalize_teardown`, which sends the error response. Then we fall through to the same `error:` label and send it again. So the client gets the response twice.

How should I handle this? I was thinking of checking if `NODE_ERROR_CODE` is already set to skip the duplicate, but I'm not sure if there's a better way. This also applies to `restore.c` and `verify.c`.

src/libpgmoneta/restore.c
Abdelrhmansersawy: @jesperpedersen 
I will revert `pgmoneta_rollup_backups` back to `carry_out_workflow`. But should I update `carry_out_workflow` to run teardowns in reverse order and also run teardowns when an error happens? Right now it skips teardowns on error and runs them forward.

src/libpgmoneta/backup.c
jesperpedersen: The things is that we need to move the logic up to workflow.c such that if there is an existing error code, then the execute isn't called. I think it is ok for all setup/teardown to be called.

finalizer should always send the repsonse.

For now, we have backup, restore and verify - but long term all workflows should use finalizer

src/libpgmoneta/restore.c
jesperpedersen: The main thing is that only workflow.c should call those - there might be pre checks inside the execute's that will fail, but that means there is a bug in the previous workflow nodes...

If we need to split the existing nodes up even more then be it

jesperpedersen:
@Abdelrhmansersawy Rebase, please

Abdelrhmansersawy:
@jesperpedersen 
About strict reponse on finialize workflow, I see that we can extract `pgmoneta_workflow_teardown` into a public function to guarantee teardowns always run. 

1. **Inside `pgmoneta_workflow_execute`:** It simply calls `pgmoneta_workflow_teardown`.
2. **On `error` labels:** I removed manual error responses. Workflows now exclusively call `pgmoneta_workflow_teardown(workflow, nodes)`.
   - **Early errors (workflow is NULL):** `pgmoneta_workflow_teardown` detects this and creates a temporary `finalizer` workflow to reliably send the error response.
   - **Execution errors:** It normally runs the teardown chain to send the response.

WDYT?

PR #1055, opened 2026-03-24
Fix verify response SHA512 string serialization

sksingh2005:
## Problem
When `verify` returned failed-file details, the response could include SHA512 hashes in fields like `Calculated` and `Original`.

The string serializer in `src/libpgmoneta/value.c` used a fixed `MISC_LENGTH` buffer for JSON string formatting. That buffer was too small for quoted SHA512 hex strings, so the response could be truncated and the client could fail while reading it.

`verify` now returns valid JSON responses even when failed-file details contain long SHA512 hash values.

### DIFF
--- a/src/libpgmoneta/value.c
+++ b/src/libpgmoneta/value.c
@@ -549,27 +549,25 @@ string_to_string_cb(uintptr_t data, int32_t format, char* tag, int indent)
 {
    char* ret = NULL;
    char* str = (char*)data;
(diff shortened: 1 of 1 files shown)


jesperpedersen:
Merged.

jesperpedersen:
@sksingh2005 Please, cross-port and remember the issue describing the problem

sksingh2005:
Ohh sorry forgot to file the issue first. Will remember this.

Issue #1051, opened 2026-03-22
Combine Search with Filtering in Interactive Mode


saraanbih:
Currently, we have search and filtering as two disconnected features: filters can only be set before launch and cannot be changed while browsing. The goal is to combine them into a single investigation workflow where the DBA can dynamically apply and stack filters while browsing WAL records across multiple files simultaneously, search within the filtered results to pinpoint the exact bad transaction, then automatically generate the walfilter rule file - YAML file - from the selection, connecting the investigation directly to the recovery process without manually writing configuration files or switching between tools.

PR #1056, opened 2026-03-24
[1051] filtering in interactive mode

saraanbih:
Fixes [#1051]

What I have done:
- Add filter option
- You can filter with one or more options
- The filtering is anded (if u filtered based on some option and make another filter, they will be combined)

### DIFF
--- a/src/walinfo.c
+++ b/src/walinfo.c
@@ -53,6 +53,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <time.h>
 #include <unistd.h>
 #include <ncurses.h>
 #include <sys/stat.h>
@@ -69,6 +70,8 @@
 #define COL_WIDTH_FPI_PCT       8
 #define COL_WIDTH_COMBINED_SIZE 14
 #define COL_WIDTH_COMBINED_PCT  10
+
+#define WAL_FILTER_MAX_TOKENS   32
 
 static int describe_walfile(char* path, enum value_type type, FILE* output, bool quiet, bool color, struct deque* rms, uint64_t start_lsn, uint64_t end_lsn, struct deque* xids, uint32_t limit, bool summary, char** included_objects);
 static int describe_walfile_internal(char* path, enum value_type type, FILE* out, bool quiet, bool color, struct deque* rms, uint64_t start_lsn, uint64_t end_lsn, struct deque* xids, uint32_t limit, bool summary, char** included_objects, struct column_widths* provided_widths);
@@ -102,6 +105,7 @@ enum column_index {
 /* WAL record wrapper for UI display */
 struct wal_record_ui
 {
+   char filename[512];
    char rmgr[32];
    uint64_t start_lsn;
    uint64_t end_lsn;
@@ -113,6 +117,7 @@ struct wal_record_ui
    bool verified;
    char verification_status[64];
    struct decoded_xlog_record* record; /* Pointer to actual record */
+   bool marked;                        /* For YAML file generation */
 };
 
 /* Field types for search queries */
@@ -147,6 +152,21 @@ struct wal_search_criteria
    bool has_end_lsn;
    bool has_xid;
    bool has_description;
+};
+
+/* Filter criteria for interactive mode */
+struct wal_filter_criteria
+{
+   char rmgr[256];
+   uint64_t start_lsn;
+   uint64_t end_lsn;
+   char xid_csv[256];
+   char relation[256];
+   bool has_rmgr;
+   bool has_start_lsn;
+   bool has_end_lsn;
+   bool has_xid;
+   bool has_relation;
 };
 
 static const char* wal_search_rmgr_values[] = {
@@ -198,8 +218,12 @@ struct ui_state
    char* wal_filename;
    struct wal_record_ui* records;
    size_t record_count;
+   size_t record_count_unfiltered;
    size_t record_capacity;
    struct walfile* wf;
+
+   /* Filter state */
+   struct wal_filter_criteria filters;
 
    /* Navigation */
    size_t current_row;
@@ -724,8 +748,7 @@ wal_interactive_load_records(struct ui_state* state, char* wal_filename)
       char* desc = get_simple_record_description(record, wf->magic_number);
       if (desc != NULL)
       {
-         strncpy(rec_ui->description, desc, sizeof(rec_ui->description) - 1);
-         rec_ui->description[sizeof(rec_ui->description) - 1] = '\0';
+         pgmoneta_snprintf(rec_ui->description, sizeof(rec_ui->description), "%s", desc);
          free(desc);
       }
       else
@@ -756,11 +779,10 @@ wal_interactive_load_records(struct ui_state* state, char* wal_filename)
       }
 
       rec_ui->verified = false;
-      strncpy(rec_ui->verification_status, "Unchecked",
-              sizeof(rec_ui->verification_status) - 1);
-      rec_ui->verification_status[sizeof(rec_ui->verification_status) - 1] = '\0';
+      pgmoneta_snprintf(rec_ui->verification_status, sizeof(rec_ui->verification_status), "Unchecked");
 
       rec_ui->record = record;
+      rec_ui->marked = false;
 
       state->record_count++;
    }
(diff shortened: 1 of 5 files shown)


saraanbih:
Hi Mr. @jesperpedersen PTAL
Note: I was trying to combine search and filter, and the result was native filtering, so I didn't see any benefit of the search, so I have separated them

saraanbih:
Hi Mr. @jesperpedersen, PTAL ;)

src/walinfo.c
```diff
@@ -779,6 +799,230 @@ wal_interactive_load_records(struct ui_state* state, char* wal_filename)
...
+         {
+            matches = false;
+         }
+      }
+
+      if (matches)
+      {
+         if (new_count != i)
+         {
+            state->records[new_count] = state->records[i];
```
Amr-Shams, line 1098: i would go for a visible array like this 
bool visible[state->record_capacity]; 
treating the records as immutable

Amr-Shams: filter pass sets visible[i] = true, fail sets visible[i] = false. The render loop skips invisible records. our data (records) are never moved or destroyed. and when clearing a filter is just resetting the visibility array

src/walinfo.c
```diff
@@ -149,6 +152,22 @@ struct wal_search_criteria
    bool has_description;
 };
 
+/* Filter criteria for interactive mode */
+struct wal_filter_criteria
+{
```
Amr-Shams, line 159: i would go for this for a better function call and reducing the code smells
struct wal_filter
  {
     int      (*apply)(struct wal_record_ui* record, void* ctx);
     void*    ctx;
     bool     enabled;
  };

src/walinfo.c
```diff
@@ -779,6 +799,230 @@ wal_interactive_load_records(struct ui_state* state, char* wal_filename)
    return 0;
 }
 
+/**
+ * Apply current filters to the records
+ * 
+ * @param state UI state structure
+ */
+static void
+wal_apply_filters(struct ui_state* state)
```
Amr-Shams, line 993: following the suggested struct above the code is much simpler now 
```
 struct wal_filter_chain
  {
     struct wal_filter filters[8];
     int               count;
  };

  // the func(apply_filters) is just 10 lines of code:
  for (size_t i = 0; i < state->record_count; i++)
  {
     state->records[i].visible = true;
     for (int f = 0; f < chain->count; f++)
     {
        if (chain->filters[f].enabled &&
            chain->filters[f].apply(&state->records[i], chain->filters[f].ctx) == 0)
        {
           state->records[i].visible = false;
           break;
        }
     }
  }
```

src/walinfo.c
saraanbih: @Amr-Shams, but when navigating, it will make a delay cuz the original navigation was through the records after filtering (such that I destroyed), and it will not affect cuz it is read-only

saraanbih: this may cause another navigation for the filter only

src/walinfo.c
saraanbih: I have a question here why this design? what difference will it make

src/walinfo.c
saraanbih: same question, it will reduce the code here, but will increase in others

src/walinfo.c
Amr-Shams: u are 1005 right. bool visible[] approach alone would have these issues.i would go for visible indices array  visible_indices[current_pos] gives you the actual record in O(1), no skipping needed.

src/walinfo.c
Amr-Shams: the main issue for me is that wal_apply_filters owns the matching logic for every single filter, which makes it a hot function, u can check the same pattern used as chain of resp (nginx and pg uses this pattern)

saraanbih:
CC: @jesperpedersen WDYT

saraanbih:
@jesperpedersen Kind Reminder ")

jesperpedersen:
@saraanbih How do I see the filters applied ? How do I edit/remove a filter ? How do I export a walfilter.yml based on the filters ?

Maybe give each filter a number - we can start with 1 to 9. If pressing 0 - open a window with showing each with the option to edit/remove with the one selected.

You can change the bottom bar to look like the top bar and then use the right side for filter display - like `1:Heap2 2:757 3:0/30097C0` -- 0 will have the full information

jesperpedersen:
@saraanbih You need documentation for the feature as well

jesperpedersen:
@saraanbih In bottom bar you can remove `Cmd:` to have a bit more space

saraanbih:
oh sorry, made with wrong while i was writing my comment

saraanbih:
Hi Mr. @jesperpedersen 

> @saraanbih How do I see the filters applied ? How do I edit/remove a filter ? How do I export a walfilter.yml based on the filters ?

- u can press `f` and type the field u wanna filter with
- if u wanna edit, u press `f` again and edit on the filtered records
- after filtering u can mark the records u wanna generate yaml file from and press `g` it will generated automatic

> Maybe give each filter a number - we can start with 1 to 9. If pressing 0 - open a window with showing each with the option to edit/remove with the one selected.
>

hmm, do u mean we should make a filter history with each filter user make? 

> You can change the bottom bar to look like the top bar and then use the right side for filter display - like 1:Heap2 2:757 3:0/30097C0 -- 0 will have the full information

we can leave the top bar as it it, the bottom will contain the commands to guide and ok at the right we can add filter history display

saraanbih:
@jesperpedersen 
> @saraanbih In bottom bar you can remove Cmd: to have a bit more space

but is it considered as a guide?

jesperpedersen:
@saraanbih I think we need some where the filters are displayed while we are working on them. It could be `F` as well...

`g` isn't documented in the UI.

And, the second `f` isn't clear...

What if I start with the latest XID, then I find out I need to use a Start LSN instead... How do I clear the XID ?

saraanbih:
@jesperpedersen 
> @saraanbih I think we need some where the filters are displayed while we are working on them. It could be F as well...
>
> g isn't documented in the UI.

OK, i am sorry, i have missed the documents -- Noted

> And, the second f isn't clear...
>

It's ANDed filtering -- u aren't just filtering once, u can do another filter and so on...

> What if I start with the latest XID, then I find out I need to use a Start LSN instead... How do I clear the XID ?

Oh, got it! -- Noted

jesperpedersen:
> It's ANDed filtering -- u aren't just filtering once, u can do another filter and so on...

Yes, but 2 XIDs filters would be an `OR`, right ? Same for `RMGR`

jesperpedersen:
@saraanbih And 2 `Start LSN`  would be the lowest, 2 `End LSN` would be the highest. And so on

jesperpedersen:
@saraanbih I think we need `<TAB>` completion on LSNs as well

saraanbih:
> Yes, but 2 XIDs filters would be an OR, right ?

yes yes, i have understood -- but how we can handle if we wanna make filter with RMGR ANDed? it will depend on the history of filtering and the field u will filter with

saraanbih:
@jesperpedersen 
> @saraanbih I think we need <TAB> completion on LSNs as well

We can refactor autocomplete at RMGR and add one for LSNs by making them with a trie, it doesn't make a significant impact at RMGR cuz the values are limited and defined, i think at LSNs it will be more efficient

jesperpedersen:
@saraanbih What do you mean ? A `RMGR` can only have one value. Think of it as

```
Filter = ((XID = 1 OR XID = 2) AND (RMGR = Heap OR RMGR = Heap2) AND (StartLSN = 0/30097C0) -- StartLSN = 0/30098DB was removed
```

You get the idea

saraanbih:
@jesperpedersen 

I mean if i filtered with `XID = 123453` and wanna filter on the filtered records with `RMGR = Heap` 

That's what I mean by ANDed

I think i have understood what you mean

You mean that between different fields make AND and with the same field make OR, right?

jesperpedersen:
@saraanbih Well, sort of - StartLSN is lowest, and EndLSN is highest. Then it is `OR`ed and `AND`ed

saraanbih:
Hi Mr. @jesperpedersen, I've worked on your comments -- PTAL

jesperpedersen:
@saraanbih You don't need the description in the fields themselves,

```
RMGR (comma=OR)   |
Start LSN (min)   |
End LSN (max)     |
XID (comma=OR)    |                           
Relation (comma=OR|
```

because you have that description below

jesperpedersen:
Merged.

Issue #1064, opened 2026-03-28
Manifest-driven S3 upload with parallel workers


Amr-Shams:
**Is your feature request related to a problem? Please describe.**
S3 uploads use recursive directory traversal, which prevents parallelization and provides no atomicity guarantee.

**Describe the solution you'd like**
Upload files from the backup manifest with worker pool support, uploading backup.manifest last as an atomic commit marker.

**Additional information**
Also extracts pgmoneta_manifest_get_paths() as a shared utility and refactors s3_download_files to use it.

PR #1065, opened 2026-03-28
#1064 feat: parallel s3 upload files

Amr-Shams:
- Add `pgmoneta_manifest_get_paths()` to read manifest CSV into a deque
 - Refactor S3 upload to be manifest-driven with worker pool support and atomic commit (manifest uploaded last) it is X(number of workers) faster
 - Refactor S3 download to use the same manifest API instead of manual CSV parsing


jesperpedersen:
Merged.

Issue #1067, opened 2026-03-29
S3 list returns empty with path-style endpoints when base_dir is empty


Amr-Shams:
using a path-style S3 endpoint (e.g., Garage) with no `s3_base_dir` configured

the `s3 ls` command was not passing the prefix argument from the CLI to the management request.

PR #1068, opened 2026-03-29
#1067 fix: fix s3 ls command to work with pfix

Amr-Shams:


Amr-Shams:
@jesperpedersen is this normal for ci to fail at the backup, i might need to look this up cause it is not related to the changes i have made

jesperpedersen:
@Amr-Shams I havn't looked yet - but it should be ok

jesperpedersen:
Merged.

Issue #1071, opened 2026-03-31
Add S3 storage CI tests using Garage


Amr-Shams:
**Is your feature request related to a problem? Please describe.**

there are no CI tests for the S3 storage engine. S3 operations (backup, list, restore, delete) are untested in the pipeline.

**Describe the solution you'd like**

add a separate CI workflow that sets up Garage (S3 compatible backend) and runs basic S3 storage tests: backup, list, restore, and delete. this ensures the S3 storage engine works end to end on every PR.

**Additional information**

a working prototype exists on the `ci/storage-tests` branch in my fork using Garage as the S3 backend on ubuntu with no docker dependency. restore test is currently failing, depends on #1069.

Issue #1070, opened 2026-03-31
S3 verify should query remote storage instead of local filesystem


Amr-Shams:
**Is your feature request related to a problem? Please describe.**

`verify` currently reads metadata (`backup.sha512`, `backup.info`) from the local filesystem only. when using S3 as the storage engine, the `data/` directory is deleted after upload, so verification against local files fails.

this also affects startup. after restarting pgmoneta, `verification_cb` runs and tries to hash local files that only exist on S3:

```
2026-03-31 17:54:01 ERROR verify.c:462 Verification: primary / Could not create hash for /tmp/pgmoneta/primary/backup/20260331175206/data/PG_VERSION.zstd
2026-03-31 17:54:01 ERROR verify.c:462 Verification: primary / Could not create hash for /tmp/pgmoneta/primary/backup/20260331175206/data/pg_ident.conf.zstd
2026-03-31 17:54:01 ERROR verify.c:462 Verification: primary / Could not create hash for /tmp/pgmoneta/primary/backup/20260331175206/data/postgresql.auto.conf.zstd
```

**Describe the solution you'd like**

`verify` should fetch metadata from S3 and validate backup integrity remotely when S3 is the storage engine. `s3_bootstrap` in `se_s3.c` already has the logic to download `backup.sha512` and `backup.info` from S3 and verify the SHA512 hash, this can be reused or extracted.

**Additional information**

per discussion with @jesperpedersen: `info` and `list-backup` can stay local (metadata is kept), but `verify` has to go to S3.

Akashx64:
Hi @Amr-Shams! I am a GSoC applicant mapping out the pgmoneta codebase. I'm very interested in working on this issue to familiarize myself with the S3 storage engine logic.

Could you point me to the source file or function where the S3 verification currently checks the local filesystem? I'd love to start tracing it on my local setup and see how we can point it to query the remote storage instead. Thanks!

Akashx64:
Hi @Amr-Shams and @jesperpedersen,
I've been studying the code and I can see two places in verify.c that need to handle S3:

Line ~399: fopen(sha512_path) fails because backup.sha512 doesn't exist locally when S3 is the storage engine
Line ~461: pgmoneta_create_sha512_file() fails for the same reason

I'm planning to add a check using config->storage_engine & STORAGE_ENGINE_S3 before these local file operations, similar to the pattern in wal.c:181. Would this approach work, or is there a preferred way to fetch files from S3 for verification?
[...]

PR #1075, opened 2026-04-02
#1071 feat: added integartion testing for s3

Amr-Shams:
- spin up garage on ubuntu 
- inject cli like list, delete,restore,backup for local pg setup 
- check the file count as a source of valid backup/list
- check the manifest file present 
- no validity check for now (once added by #1070 will be added into the integration testing)

### DIFF
--- a/.github/workflows/storage.yml
+++ b/.github/workflows/storage.yml
@@ -1,0 +1,335 @@
+name: Storage
+
+on:
+  pull_request:
+    branches:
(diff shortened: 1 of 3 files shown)


Amr-Shams:
right now the workflow is added whenever there is a pr to the main etc, i can make that workflow needs the ci workflow

jesperpedersen:
Merged.

PR #1076, opened 2026-04-02
#1071 ref:cleanup jobs for a better perforamnce

Amr-Shams:
- just clean up to the ci for a better flow
- now it only checks if the ci is completed 
- separate jobs from steps 
- use right vars and contexts


jesperpedersen:
Merged.

Issue #1080, opened 2026-04-05
Worker pool: no early abort on error and data race on outcome flag


Amr-Shams:
**Describe the bug**
When a worker encounters a fatal error and sets outcome = false, already-queued tasks continue executing. worker_do() in workers.c only checks worker_keepalive before executing tasks, it never checks outcome(shared state between workers).

Additionally, `bool outcome` in `workers.h:83` is not `_Atomic`, but is written by worker threads and read by the main thread without synchronization.
 
Found while investigating #1070.

Amr-Shams:
right now it works like continue-on-error(github actions), but no fast-fail config by the caller

Amr-Shams:
@jesperpedersen what you think of the worker pool? 
it should handle the error propagation and draine the workers from signals/error(fast-fail)/retery policy etc. 
i think it needs some work to be done there

jesperpedersen:
@Amr-Shams Yes, we probably need different policies (ALWAYS / FAST_FAIL). And, yes - we can use atomic.

Amr-Shams:
> [@Amr-Shams](https://github.com/Amr-Shams) Yes, we probably need different policies (ALWAYS / FAST_FAIL). And, yes - we can use atomic.

@jesperpedersen yep, that makes sense. i understood ALWAYS / FAST_FAIL like this:
- ALWAYS: process everything in the queue, then report final outcome.
- FAST_FAIL: first fatal worker error should stop the rest ASAP (graceful stop), not keep running queued tasks.
i already made policy caller-managed(#1087), but i think we should lock the behavior down a bit:
1. Worker task returns 0 on success, non-zero on failure.
2. Shared state is atomic (outcome + a shared stop flag-used when shutdown gracefully).
3. In FAST_FAIL, once a task fails:
   - set outcome=false
   - set stop flag
   - workers check stop flag before next task and exit cleanly.
then maybe as a follow-up, add a timeout safety net in case a worker gets stuck( it is more like the go-lang style of graceful-shutdown)

jesperpedersen:
@Amr-Shams Yes, for `FAST_FAIL` we need to process all tasks already started, but we can canceled the rest

Amr-Shams:
@jesperpedersen since FAST_FAIL now cancels queued-not-started tasks, we need a safe cleanup path for those work items/payloads (wc). freeing payloads generically in the pool is unsafe, but skipping cleanup
[...]

PR #1087, opened 2026-04-09
[#1080] Add fast fail support to worker pool

Amr-Shams:
Add fast fail support to the worker pool. When enabled, workers stop picking up new tasks once a failure is detected. Enabled for verify, S3 download, and S3 upload workers.

**Note**: The task caller is responsible for setting `workers->outcome = false` on failure. maybe a cahnge to switch task functions to return int so worker_do can manage outcome automatically along with the retery policy etc for the worker pool.

### DIFF
--- a/src/include/workers.h
+++ b/src/include/workers.h
@@ -37,8 +37,7 @@ extern "C" {
 #include <deque.h>
 
 #include <pthread.h>
-#include <stdio.h>
-#include <stdlib.h>
+#include <stdatomic.h>
 
 struct worker_common;
 
@@ -58,6 +57,7 @@ struct semaphore
 struct worker_task
 {
    void (*function)(struct worker_common*); /**< The task function */
+   void (*destroy)(struct worker_common*);  /**< The task destroy function */
    struct worker_common* wc;                /**< Pointer to the common data */
 };
 
@@ -80,7 +80,9 @@ struct workers
    volatile int number_of_working; /**< The number of workers */
    pthread_mutex_t worker_lock;    /**< The worker lock */
    pthread_cond_t worker_all_idle; /**< Are workers idle */
-   bool outcome;                   /**< Outcome of the workers */
+   atomic_bool fast_fail;          /**< fast fail to the workers*/
+   atomic_bool outcome;            /**< Outcome of the workers*/
+   atomic_bool queue_drained;      /**< Queued tasks have been canceled */
    struct deque* queue;            /**< The task queue using deque */
    struct semaphore* has_tasks;    /**< Semaphore for tasks dispatching*/
 };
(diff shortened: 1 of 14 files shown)


Amr-Shams:
this patch introduces the worker-pool support for cancel-safe cleanup via an optional task destructor, but existing callers still need to migrate to the new API. until they do, canceled queued payloads retain existing behavior

Amr-Shams:
@jesperpedersen PTAL

src/include/workers.h
```diff
@@ -162,6 +175,35 @@ int
 pgmoneta_create_worker_input(char* directory, char* from, char* to, int level,
                              struct workers* workers, struct worker_input** wi);
 
+/**
+ * Enable fast fail for workers
+ * When enabled, workers stop processing new tasks after outcome is set to false
+ * @param workers The workers
+ */
+void
+pgmoneta_set_fast_fail(struct workers* workers);
```
jesperpedersen: Should be init

src/include/workers.h
```diff
@@ -162,6 +175,35 @@ int
...
+
+/**
+ * mark the task as a failure
+ * @param workers The workers
+ */
+void
+pgmoneta_workers_mark_failure(struct workers* workers);
```
jesperpedersen, line 183: Do we need this ?

src/include/workers.h
```diff
@@ -162,6 +175,35 @@ int
...
+
+/**
+ * get the workers outcome 
+ * @param workers The workers
+ */
+bool
+pgmoneta_workers_outcome(struct workers* workers);
```
jesperpedersen, line 190: I guess this is ok

src/include/workers.h
```diff
@@ -162,6 +175,35 @@ int
...
+
+/**
+ * check if the worker can accpet or not new tasks 
+ * @param workers The workers
+ */
+bool
+pgmoneta_workers_should_accept(struct workers* workers);
```
jesperpedersen: Sounds like a static

jesperpedersen (changes requested):

jesperpedersen:
@Amr-Shams Lets start with these

src/include/workers.h
Amr-Shams: yes, because i want the outcome transition to be one-way from the api side, and will make the refactoring more explicit and easy to track

src/include/workers.h
Amr-Shams: Kept it separate so callers can opt in, not all need fast-fail. Also avoids breaking existing `pgmoneta_workers_initialize` call sites. or u mean renaming the func ?

jesperpedersen: @Amr-Shams I would add a `, bool fast_fail` here and let all existing call-sites use `false`

Amr-Shams (commented):
this was a workaround on a big failure #1070 handled here

PR #1088, opened 2026-04-10
[#516] Configuration generator

### LINKED ISSUE #516: Configuration generator
Create a tool that can generate an initial configuration file (`pgmoneta.conf`) based on user inputs.

Could be from console, or (more difficult) from command line parameters

Sakshii-27:
This pull request introduces `pgmoneta-config`, a new command-line tool that makes it easy to generate and manage `pgmoneta.conf` configuration files.

### What is included
* **Interactive Setup (`init`):** Users can run the tool to answer a few quick prompts, and it will automatically build a fully valid configuration file for them.
* **Quiet Mode (`-q`):** Quickly generate a default configuration template without asking any questions. Perfect for automated testing or CI/CD scripts.
* **Easy Editing (`get`, `set`, `del`, `ls`):** Read, add, or delete specific configuration values directly from the terminal without opening a text editor. When making edits, the tool safely preserves your existing file structure and comments. 
* **Safe File Management:** Instead of risking data loss if an error happens while saving, the tool writes to a temporary file first safely swapping it. This ensures configurations are never accidentally corrupted.
* **Secure Defaults:** Newly generated configuration files are strictly secured so that only the owner can read or write to them (`600` permissions) to protect your passwords.

Closes #516


Abdelrhmansersawy:
@Sakshii-27 
Hi! You might want to take a look at how pgagroal handles this. The `conf get` and `conf set` commands are already part of `pgagroal-cli`.

You could check `pgagroal_conf_get()` and `pgagroal_conf_set()` in `configuration.c`, and also how it’s handled in `cli.c`.

Sakshii-27:
@Abdelrhmansersawy, thanks for the comment! The pgagroal conf get/set approach works over the management socket against a running daemon but that's not what #516 is asking for. The issue is specifically about generating the initial config file from scratch, before the daemon can even be started. There's no running process to talk to at that point, so this tool works directly on the file, following the crudini-style approach that was suggested in the issue.

jesperpedersen:
@Sakshii-27 We need a `CONFIG.md` file, and a chapter in the manual for this

Sakshii-27:
done.

jesperpedersen:
Rebased, and merged.

jesperpedersen:
@Sakshii-27 I think this can be cross-ported to the other projects

Issue #1089, opened 2026-04-10
S3 restore: verify downloaded files and clean up partial backup on failure


Amr-Shams:
per discussion with @jesperpedersen: S3 restore should verify the integrity of downloaded files before considering the backup valid, and clean up the entire backup directory on any failure.

currently `s3_storage_restore` downloads metadata and data files, renames `.tmp` files, and returns success without verifying that the downloaded data matches the manifest hashes. if a file is corrupted during transfer or partially written, the backup appears valid but is not.

the restore flow should be: download from S3 → verify against manifest → restore (decompress/decrypt). if any step fails, delete the local backup directory entirely instead of leaving a partial or invalid backup on disk.

this reuses the existing verify workflow (`wf_verify.c`) after the S3 download step completes, chaining two workflow executions in `pgmoneta_restore_s3_objects`.

related to #1070.

PR #1090, opened 2026-04-10
[#1089] chain verify for the restore workflow of s3

Amr-Shams:
chain verification to the resotore of the s3 workflow 
separate workflow exectuion for the verification


jesperpedersen:
Rebased, and merged.

PR #1093, opened 2026-04-11
[#1089] fix the integration testing

Amr-Shams:
this should fix the integration testing for s3 
-s3 doesn't follow the normal path of verification like wf_restore does
- it need substeps from the original flow 
- extracted this flow into a helper method and injiected them dir to the s3 workflow of restoring
- s3 doesn't need the exclude nor the permission steps

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -167,13 +167,13 @@ jobs:
           path: /tmp/pgmoneta-test/log
           retention-days: 90
 
-  # storage:
-  #   needs: [format-check, build-linux]
-  #   strategy:
-  #     matrix:
-  #       platform: [ubuntu-latest]
-  #       pg_version: ["14"]
-  #   uses: ./.github/workflows/storage.yml
-  #   with:
-  #     platform: ${{ matrix.platform }}
-  #     pg_version: ${{ matrix.pg_version }}
+  storage:
+    needs: [format-check, build-linux]
+    strategy:
+      matrix:
+        platform: [ubuntu-latest]
+        pg_version: ["17"]
+    uses: ./.github/workflows/storage.yml
+    with:
+      platform: ${{ matrix.platform }}
+      pg_version: ${{ matrix.pg_version }}
--- a/src/libpgmoneta/s3.c
+++ b/src/libpgmoneta/s3.c
@@ -36,10 +36,12 @@
 #include <logging.h>
 #include <management.h>
 #include <network.h>
+#include <restore.h>
 #include <security.h>
 #include <utils.h>
 #include <wal.h>
 #include <workflow.h>
+#include <workflow_funcs.h>
 
 #define NAME "s3"
 
@@ -326,12 +328,14 @@ pgmoneta_restore_s3_objects(int client_fd, int server, char* prefix, uint8_t com
    struct timespec start_t;
    struct timespec end_t;
    double total_seconds;
-   char* base_dir = NULL;
+   char* directory = NULL;
+   char* position = NULL;
    struct art* nodes = NULL;
    struct workflow* workflow = NULL;
    struct main_configuration* config;
    struct backup* backup = NULL;
    char* local_root = NULL;
+   struct json* req = NULL;
 
    config = (struct main_configuration*)shmem;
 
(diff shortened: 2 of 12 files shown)


jesperpedersen:
@Amr-Shams We should have individual steps - its `pgmoneta_workflow_create_verify_chain` I don't like

Amr-Shams:
> @Amr-Shams We should have individual steps - its `pgmoneta_workflow_create_verify_chain` I don't like

done, used inline steps(maybe later when Azure, SSH ) needs the same steps we might introduce the a helper methods

jesperpedersen:
@Amr-Shams Check you want `storage` in `ci.yml` back again...

Amr-Shams:
> @Amr-Shams Check you want `storage` in `ci.yml` back again...

oh  it's in there, gated on format-check + build-linux so it only runs once the main ci is green (40df095)

jesperpedersen:
@Amr-Shams Yes, but https://github.com/pgmoneta/pgmoneta/commit/2656cf6190a155d398ed08b92d9aca2db72669c0 since it was failing

.github/workflows/ci.yml
```diff
@@ -167,13 +167,13 @@ jobs:
...
-  #   uses: ./.github/workflows/storage.yml
-  #   with:
-  #     platform: ${{ matrix.platform }}
-  #     pg_version: ${{ matrix.pg_version }}
+  storage:
+    needs: [format-check, build-linux]
+    strategy:
+      matrix:
+        platform: [ubuntu-latest]
+        pg_version: ["14"]
```
jesperpedersen: Is this because Ubuntu has PostgreSQL 14 ? Otherwise we should use 17+

Amr-Shams: not really, i was aiming for the least supported version in pg right now in the matrix, but sure will not make any diff now

src/libpgmoneta/s3.c
```diff
@@ -399,7 +416,64 @@ pgmoneta_restore_s3_objects(int client_fd, int server, char* prefix, uint8_t com
       pgmoneta_log_error("S3 restore: could not load info for %s", config->common.servers[server].name);
       goto error;
    }
-   workflow = pgmoneta_workflow_create(WORKFLOW_TYPE_VERIFY, backup);
+
+   {
```
jesperpedersen: I don't understand this - and remove the code block

jesperpedersen (changes requested):

jesperpedersen:
@Amr-Shams Cleanup any unnecessary changes as well

src/libpgmoneta/s3.c
Amr-Shams: the verification workflow (`wf-verify`) currently injects several additional steps,restore point creation, file exclusion, and permission handling. these make sense in a full restore scenario, but they’re unnecessary for the S3 verification step.

also backup metadata is strictly loaded from remote S3. while pulling from local metadata or files would simplify things, it would violate the design principle that all state must originate from remote storage.

cause of this, the verification process had to be implemented as a separate (chained) workflow at runtime. the workflow focuses only on what’s required for integrity checks: decompression, decryption, and validation. if any of these steps fail, the entire backup is treated as invalid and removed.

src/libpgmoneta/s3.c
```diff
@@ -399,7 +416,62 @@ pgmoneta_restore_s3_objects(int client_fd, int server, char* prefix, uint8_t com
       pgmoneta_log_error("S3 restore: could not load info for %s", config->common.servers[server].name);
       goto error;
    }
-   workflow = pgmoneta_workflow_create(WORKFLOW_TYPE_VERIFY, backup);
+
+   struct workflow* chain_head = NULL;
```
jesperpedersen: There is something wrong here - s3.c should only work on S3 stuff.

Encryption and compression should have been called by workflow.c before S3 fires...

Amr-Shams: yep, you hit it, so the flow i was thinking off was a little odd than the right restore workflow which requires a destniation, not in place(unless the ueser wants that), so i have removed this manual steps creation and used the right restore calls from the restore func

jesperpedersen:
Rebased, and merged.

PR #1094, opened 2026-04-11
[#1091] walinfo/Interactive: LSN completion

### LINKED ISSUE #1091: walinfo/Interactive: LSN completion
Allow for `<TAB>` completion on `Start LSN` and `End LSN` across the loaded WAL files

Sakshii-27:
Fixes #1091 

This adds `<TAB>` completion for the Start LSN and End LSN fields in the interactive search dialog (pgmoneta-walinfo -I).

### DIFF
--- a/src/walinfo.c
+++ b/src/walinfo.c
@@ -28,6 +28,7 @@
 
 /* pgmoneta */
 #include <achv.h>
+#include <art.h>
 #include <cmd.h>
 #include <configuration.h>
 #include <deque.h>
@@ -263,6 +264,8 @@ struct ui_state
    size_t current_match_index;
    size_t total_matches_in_field;
    enum wal_search_field current_search_field;
+
+   struct art* lsn_art;
    char current_search_query[256];
 };
 
@@ -283,9 +286,10 @@ static int wal_interactive_load_file(struct ui_state* state, const char* path);
 static int wal_interactive_load_records(struct ui_state* state, char* wal_filename);
 static void wal_set_window_theme(WINDOW* win, int color_pair, bool border);
 static void wal_reset_completion_cycle(bool* active, int* cycle_index, char cycle_seed[][128], int fields);
-static bool wal_input_has_letter(const char* input);
-static const char* const* wal_get_known_values_for_field(int field);
+static bool wal_input_has_content(const char* input);
+static const char* const* wal_get_known_values_for_field(struct ui_state* state, int field, const char* partial, char*** dynamic_array_out, char** field_inputs);
 static const char* wal_get_known_value_completion(const char* const* values, const char* partial, int cycle_index);
+static void wal_handle_tab_completion(struct ui_state* state, int current_field, char** field_inputs, size_t* field_input_sizes, bool* completion_active, int* completion_cycle, char completion_seed[][128]);
 
 static void
 wal_interactive_endwin(void)
@@ -376,7 +380,7 @@ wal_reset_completion_cycle(bool* active, int* cycle_index, char cycle_seed[][128
 }
 
 static bool
-wal_input_has_letter(const char* input)
+wal_input_has_content(const char* input)
 {
    if (input == NULL)
    {
@@ -385,7 +389,7 @@ wal_input_has_letter(const char* input)
 
    for (size_t i = 0; input[i] != '\0'; i++)
    {
-      if (isalpha((unsigned char)input[i]))
+      if (!isspace((unsigned char)input[i]))
       {
          return true;
       }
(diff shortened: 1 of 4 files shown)


jesperpedersen:
@saraanbih PTAL

src/walinfo.c
```diff
@@ -371,37 +369,6 @@ wal_reset_completion_cycle(bool* active, int* cycle_index, char cycle_seed[][128
...
-{
-   switch (field)
-   {
-      case 0:
-         return wal_search_rmgr_values;
-      default:
-         return NULL;
-   }
-}
-
```
saraanbih, line 408: why do you remove this?

src/walinfo.c
```diff
@@ -2732,6 +2704,91 @@ show_known_values_autocomplete(WINDOW* win, const char* const* values, const cha
    }
 }
 
+static void
+wal_build_lsn_values(struct ui_state* state, char*** start_lsns, size_t* start_count, char*** end_lsns, size_t* end_count)
```
saraanbih: the LSN values are added in the order they appear in the records, which is already chronological since WAL is sequential, so this is fine in practice but there is no explicit sort, if records were ever out of order, the autocomplete would show unsorted LSNs

src/walinfo.c
```diff
@@ -2732,6 +2704,91 @@ show_known_values_autocomplete(WINDOW* win, const char* const* values, const cha
...
+
+      if (s_lsn == NULL || e_lsn == NULL)
+      {
+         free(s_lsn);
+         free(e_lsn);
+         continue;
+      }
+
+      bool s_exists = false;
+      for (size_t j = 0; j < *start_count; j++)
```
saraanbih: it is O(n^2), in the worst case, it scans all existing LSNs linearly, assuming 100000 records - it will be a massive number of comparisons

--

u can use a hash set or at least sort + binary search - it will be more fine

src/walinfo.c
```diff
@@ -3128,6 +3232,20 @@ handle_filter_input(struct ui_state* state)
    int autocomplete_row = table_start_row + 8;
    int footer_row = height - 5;
 
+   char** start_lsns = NULL;
+   size_t start_lsn_count = 0;
+   char** end_lsns = NULL;
+   size_t end_lsn_count = 0;
+
+   wal_build_lsn_values(state, &start_lsns, &start_lsn_count, &end_lsns, &end_lsn_count);
```
saraanbih: what's after filtering?

src/walinfo.c
```diff
@@ -2732,6 +2704,91 @@ show_known_values_autocomplete(WINDOW* win, const char* const* values, const cha
...
+   {
+      *start_lsns = NULL;
+      *start_count = 0;
+      *end_lsns = NULL;
+      *end_count = 0;
+      return;
+   }
+
+   *start_lsns = calloc(state->record_count + 1, sizeof(char*));
+   *end_lsns = calloc(state->record_count + 1, sizeof(char*));
```
saraanbih: the two calloc calls are checked together in a single condition, which makes the failure handling unclear and harder to maintain

saraanbih:
Hi @Sakshii-27, good job thx, it will help us in the interactive mode
You can look at my review comments

Does the autocomplete work with you? -- I test it, and it doesn't work with me

Don't forget the autocomplete will be at the search and filter

Don't forget to update the documentation related to this part

src/walinfo.c
Sakshii-27: I removed those two functions to properly support the new LSN completion:
wal_get_known_values_for_field only worked on static arrays (like RMGR). Since our LSN arrays are dynamically generated, I replaced it with a localized array mapping that safely handles both static and dynamic fields.
wal_input_has_letter scanned for alphabet characters (isalpha). Since LSNs are mostly numeric (like 0/0), keeping this function blocked autocomplete for LSN fields.

src/walinfo.c
Sakshii-27: I completely rewrote this to fix both the sorting and the O(n^2) performance issue.

We now explicitly qsort() the raw uint64_t LSNs first, then do a fast, single linear pass to properly deduplicate them. This guarantees strict chronological order for the autocomplete and brings the performance down to O(N log N).

src/walinfo.c
Sakshii-27: Currently, wal_build_lsn_values explicitly reads from the full state->records array, so the autocomplete suggestions reflect ALL records present in the active WAL file.

This is intentional, if a user already has an active filter applied, we don't want to restrict their future autocomplete suggestions strictly to that subset. Showing the global master list of chronological LSNs guarantees they can always easily tab-complete to completely new LSN bounds at any time

Sakshii-27:
Hi @saraanbih, thank you for the review! I've addressed all your inline comments.

Regarding the autocomplete not working for you: I found exactly why it broke and pushed the fixes
1. Previously, the code had a strict `strlen == 0` check that prevented users from pressing `Tab` on an empty field. Removing this now allows `Tab` to seamlessly grab the first suggestion.
2. There was also a hidden UI bug: the popup windows were hardcoded to a height of 28/30 lines. On smaller default terminal screens, `ncurses` quietly threw an error and stripped away the `keypad()` bindings! This is why the UI felt unresponsive and navigation keys totally failed. I've added safety boundary checks, so it now renders perfectly on all screen sizes.

I have thoroughly tested it and can confirm that LSN autocompletion works smoothly in **both** the Search (`s`) and Filter (`f`) dialogs!

Finally, I have updated both the man page (`doc/man/pgmoneta-walinfo.1.rst`) and the user manual (`doc/manual/en/18-wal-tools.md`) to explicitly explain how users can trigger this new `Tab` completion feature.

src/walinfo.c
saraanbih: But those functions are related to RMGR completion, the static array cuz the values of RMGR are defined, and they are considered separated from LSN completion

src/walinfo.c
saraanbih: very well

jesperpedersen:
@Sakshii-27 @saraanbih Once everything is ready then PTAL me

saraanbih:
Hi @Sakshii-27, thanks for your work -- sorry for being late -- my laptop wasn't working, and I was fixing it

--

There is overflow at the screen -- make sure that the autocomplete doesn't go out of the borders
image

--

And after the finish, and go for another field, still the borders contain garbage values
image

saraanbih:
Hi Mr. @jesperpedersen 

> @Sakshii-27 @saraanbih Once everything is ready then PTAL me

Don't worry, surely after I see everything is ok and it is ready for merge, I will mention you

src/walinfo.c
Sakshii-27: That's a very fair point. Even though RMGR and LSN are conceptually separated, the issue was that they both share the exact same Tab completion event loop inside handle_search_input.

If I had kept those old RMGR helpers around, I would have had to clutter that generic loop with field-specific rules (like if (current_field == 0)). By removing them and mapping everything into a single field_known_values array, the Tab loop runs cleanly without needing to know or care which field it's actively processing.

Sakshii-27:
Hi @saraanbih, Great catch! Fixed both, suggestion values are now truncated to stay within the window borders, and the clearing width is adjusted to properly wipe the entire suggestion area when switching fields. No more overflow or leftover characters.

saraanbih:
@Sakshii-27 
> Hi @saraanbih, Great catch! Fixed both, suggestion values are now truncated to stay within the window borders, and the clearing width is adjusted to properly wipe the entire suggestion area when switching fields. No more overflow or leftover characters.

Did you test it local? because it is the same at my branch

Sakshii-27:
yes I did test it locally and it worked fine for me

image

saraanbih:
@Sakshii-27 
> yes I did test it locally and it worked fine for me

Hmm, I think because of the difference in screen scale

jesperpedersen:
@Sakshii-27 You need to make sure that the visible suggestions will fit the filter box

Sakshii-27:
@jesperpedersen I checked it for filter box too
image

jesperpedersen:
@Sakshii-27 Look like you are cutting off numbers. Only display what you can. Keep updating when input is given or <TAB> completion is done

saraanbih:
Hi @Sakshii-27 

I think it is better to make autocomplete with trie -- it is more efficient and used in real systems

This may help
https://ali-ibrahim137.github.io/software/engineering/2020/05/06/Autocomplete.html

Sakshii-27:
> @Sakshii-27 Look like you are cutting off numbers. Only display what you can. Keep updating when input is given or completion is done

Hi @jesperpedersen, fixed autocomplete truncation and layout.

image

image

Sakshii-27:
@saraanbih, thanks for the suggestion! A trie would indeed be more efficient for prefix-based autocomplete.
I looked into it, pgmoneta already has an ART (Adaptive Radix Tree) in libpgmoneta/art.c, which is essentially a more advanced trie. However, it currently doesn't have a prefix search/iteration API. So we'd either need to:
Add a prefix search function to the existing ART (libpgmoneta/art.c), or Build a standalone trie specifically for walinfo
@jesperpedersen what do you think we should do regarding this?

saraanbih:
@Sakshii-27 
> @saraanbih, thanks for the suggestion! A trie would indeed be more efficient for prefix-based autocomplete.
I looked into it, pgmoneta already has an ART (Adaptive Radix Tree) in libpgmoneta/art.c, which is essentially a more advanced trie. However, it currently doesn't have a prefix search/iteration API. So we'd either need to:
Add a prefix search function to the existing ART (libpgmoneta/art.c), or Build a standalone trie specifically for walinfo
@jesperpedersen what do you think we should do regarding this?

you don't need to build a standalone, see how we can use the existing one at our search and autocomplete

Sakshii-27:
Hi @jesperpedersen and @saraanbih, I’ve just updated the PR! I integrated the existing ART for prefix search as suggested. PTAL

saraanbih:
Hi @Sakshii-27 
> Hi @jesperpedersen and @saraanbih, I’ve just updated the PR! I integrated the existing ART for prefix search as suggested. PTAL

Great !

LGTM, CC: @jesperpedersen 

Many thanks for your work, means a lot !

jesperpedersen:
@Sakshii-27 We are close here, but the completion should stop as soon it doesn't "know" how to continue.

So, f.ex. if you have

```
0/2...
0/3...
```

doing a `<TAB>` should resolve to `0/`. The suggestions should keep being updated based on the current selection

saraanbih:
@jesperpedersen 
> @Sakshii-27 We are close here, but the completion should stop as soon it doesn't "know" how to continue.
>
> So, f.ex. if you have
>
> 0/2...
> 0/3...
>
> doing a <TAB> should resolve to 0/. The suggestions should keep being updated based on the current selection

It is already doing this
image

image

jesperpedersen:
@saraanbih Start with an empty field... it didn't work here

Sakshii-27:
Hi @saraanbih, I think @jesperpedersen is referring specifically to the **`<TAB>` completion behavior** (Longest Common Prefix) rather than the visual suggestions list at the bottom.

Currently, if matches are `0/2...` and `0/3...`, hitting `<TAB>` on `0` immediately fills in the first full match (`0/2...`). I believe he means that hitting `<TAB>` should only resolve to `0/` (the shared part), stopping there because the next character is ambiguous.

@jesperpedersen, could you confirm if that's what you meant? If so, I can update the Tab handler to resolve the longest common prefix first.

saraanbih:
@Sakshii-27 Ahhh okay -- great point

jesperpedersen:
@Sakshii-27 Yes, correct

Sakshii-27:
@jesperpedersen I’ve pushed the update for LCP-based Tab completion PTAL

jesperpedersen:
@Sakshii-27 There are still bugs in this... Try and create a filter with `Heap2`, then try and tab completion on `Start LSN`. LSNs that doesn't fulfill `Heap2` are being listed

jesperpedersen:
@Sakshii-27 Side note, you have to rebase before you force push

jesperpedersen:
@Sakshii-27 There is still something going on - create filter on Heap2, then create filter on Start LSN `<TAB>` -> `0/` (correct), then `0/3050<TAB>` - it completes to the first one and all suggestions disappear. I havn't tried End LSN yet - make sure that suggestions are always larger than Start LSN

Sakshii-27:
@jesperpedersen PTAL

jesperpedersen:
@Sakshii-27 @saraanbih We have to decide on some keys to go through the suggestions - `j` and `k` are standard - because it is still too difficult to get the right entry

jesperpedersen:
Of course we have use some keys that won't be used in the completions - maybe `<ARROW_LEFT>` and `<ARROW_RIGHT>` ?

Sakshii-27:
@jesperpedersen,
Sounds good, I'll implement <ARROW_LEFT> and <ARROW_RIGHT> to navigate through the suggestion grid, with <ENTER> to select the highlighted entry.

Sakshii-27:
@jesperpedersen, I've pushed the update with <ARROW_LEFT>/<ARROW_RIGHT> navigation and <ENTER> to select. I also added pagination so we don't lose the cursor in long lists.
I noticed the Search and Filter dialogs are almost identical (~80% duplication). Should I refactor those into one helper function to clean it up?

jesperpedersen:
@Sakshii-27 Yes, that sounds like a good idea

Sakshii-27:
@jesperpedersen PTAL

jesperpedersen:
Rebased, and merged.

jesperpedersen:
@Sakshii-27 Please, cross port the ART changes to pgagroal, pgexporter and pgvictoria

Jubilee101:
Hi @Sakshii-27 , nice work on adding prefix search support for ART. Can you also add unit test cases? Refer to https://github.com/armon/libart as it may already have some we can borrow.

PR #1097, opened 2026-04-14
[#1008] Expand backup Progress to include other workflow progress

### LINKED ISSUE #1008: Expand backup Progress to include other workflow progress
Backup progress is computed only from the **basebackup** workflow. We should also include the other workflows (encryption, compression, manifest, linking, sha512) when calculating the overall progress.

Right now, we can assign a weight e.g `back_backup = 85%` to each workflow and implement a progress report inside each workflow to compute its own progress based on previous workflow.

Abdelrhmansersawy:
resolved #1008 

Weights are now adaptive. If a previous backup exists, the system reads the elapsed time of each phase and computes weights accordingly.

**Full backup (no compression/encryption):**
`Base backup 77 | SHA512 14 | Manifest 7 | Link 2`

**Incremental + Compressed + Encrypted:**
`Base backup 84 | SHA512 4 | Manifest 7 | Compression 2 | Encryption 1 | Link 2`

## APIs

1. `pgmoneta_progress_setup` - calculates phase weights (store them on ART), and starts progress tracking for a workflow.

2. `pgmoneta_progress_next_phase` - Marks the end of the current phase, resets counters, and starts the next phase.

3. `pgmoneta_progress_report` - Generates a real-time progress summary (percent, elapsed time, ETA) for the CLI or logging metrics.

4. `pgmoneta_progress_increment` - Atomically adds amount (e.g., bytes/files) to the completed work of the active phase.

5. `pgmoneta_progress_teardown` - Finalizes state, records total elapsed time, and clears the shared memory space.


Abdelrhmansersawy:
@jesperpedersen 

Could you give this a high-level review to check if I’m heading in the right direction?

I'm working now on testing progress on large dataset backups to see the effect of each workflow and try to optimize the progress weights to be smoother - this may take some time.

src/include/aes.h
```diff
@@ -137,10 +137,11 @@ pgmoneta_decrypt(char* ciphertext, int ciphertext_length, char* password, size_t
  * @param d The directory
  * @param workers The optional workers
  * @param excludes The optional excludes
+ * @param server The server index for progress tracking, or -1 to disable
  * @return 0 upon success, otherwise 1
  */
 int
-pgmoneta_encrypt_directory(char* d, struct workers* workers, struct deque* excludes);
+pgmoneta_encrypt_directory(char* d, struct workers* workers, struct deque* excludes, int server);
```
jesperpedersen, line 144: Put `server` first

src/include/management.h
```diff
@@ -120,8 +120,8 @@ extern "C" {
 #define MANAGEMENT_ARGUMENT_BACKUPS               "Backups"
 #define MANAGEMENT_ARGUMENT_BACKUP_SIZE           "BackupSize"
 #define MANAGEMENT_ARGUMENT_BIGGEST_FILE_SIZE     "BiggestFileSize"
-#define MANAGEMENT_ARGUMENT_BYTES_DONE            "BytesDone"
-#define MANAGEMENT_ARGUMENT_BYTES_TOTAL           "BytesTotal"
+#define MANAGEMENT_ARGUMENT_DONE                  "BytesDone"
```
jesperpedersen, line 123: Why ?

src/include/pgmoneta.h
```diff
@@ -295,11 +303,15 @@ struct s3_configuration
  */
 struct backup_progress
 {
-   atomic_int state;         /**< The progress state */
-   atomic_llong bytes_done;  /**< The bytes transferred so far */
-   atomic_llong bytes_total; /**< The total bytes to transfer */
-   atomic_llong start_time;  /**< The start time */
-   atomic_llong elapsed;     /**< The elapsed time */
+   atomic_int state;            /**< The progress state */
+   atomic_llong done;           /**< Units of work completed (bytes or files) */
```
jesperpedersen, line 307: Hmm, ok

jesperpedersen (changes requested):

jesperpedersen:
@Abdelrhmansersawy Yes, this is the correct direction. `server` is first in parameters

src/libpgmoneta/compression.c
```diff
@@ -330,15 +334,29 @@ do_compression_operation(struct worker_common* wc)
       task->common.workers->outcome = false;
    }
 
+   if (task->progress_enabled)
```
Amr-Shams, line 337: just small follow up a
percentage interpolation logic is duplicated here and in wf_sha512.c and aes.c. Consider making into a dry helper.

src/libpgmoneta/aes.c
```diff
@@ -2406,6 +2425,20 @@ do_aes_operation(struct worker_common* wc)
       task->common.workers->outcome = false;
    }
 
+   if (task->progress_enabled)
```
Amr-Shams, line 2428: both the synchronous path in pgmoneta_encrypt_data and in do_aes_operation increment, if the dispatch logic ever changes, this becomes a double-count. i think keeping only one is enough

src/libpgmoneta/wf_bzip2.c
```diff
@@ -144,7 +145,13 @@ bzip2_execute_compress(char* name __attribute__((unused)), struct art* nodes)
       pgmoneta_deque_add(excludes, "backup.sha512.tmp", 0, ValueString);
       pgmoneta_deque_add(excludes, "backup.sha256", 0, ValueString);
 
-      if (pgmoneta_compress_directory(backup_base, COMPRESSION_CLIENT_BZIP2, workers, excludes))
+      if (pgmoneta_is_progress_enabled(server))
+      {
+         int file_count = pgmoneta_count_files(backup_base);
```
Amr-Shams, line 150: the count files can make use of the manifest, reducing the total ios by 100x at least, and also there is a possibility for a TOCTOU window, if the total is wrong the perc can exceed with no clamp

src/include/compression.h
```diff
@@ -144,10 +144,12 @@ pgmoneta_compress_file(char* from, char* to, int type, struct workers* workers);
  * @param directory The directory path
  * @param type The compression type
  * @param workers Optional worker pool. If NULL, runs synchronously.
+ * @param excludes Excluded file patterns
+ * @param server The server index for progress tracking, or -1 to disable
  * @return 0 on success, otherwise 1
  */
 int
-pgmoneta_compress_directory(char* directory, int type, struct workers* workers, struct deque* excludes);
+pgmoneta_compress_directory(char* directory, int type, struct workers* workers, struct deque* excludes, int server);
```
Amr-Shams, line 152: i am with jesper in this, but i don't think it should be passed a server, if possible to remove it, maybe setting total and passing as a pointer would be great option,

src/libpgmoneta/aes.c
```diff
@@ -28,6 +28,7 @@
 
 #include <pgmoneta.h>
 #include <aes.h>
+#include <backup.h>
```
Amr-Shams, line 31: aes shouldn't depend on backup, the pgmoneta_is_progress_enabled can be leveled up on the hierarchy.

src/libpgmoneta/workflow.c
```diff
@@ -436,6 +490,327 @@ pgmoneta_common_teardown(char* name, struct art* nodes)
...
+         return "Compression";
+      case PROGRESS_PHASE_ENCRYPTION:
+         return "Encryption";
+      default:
+         return "Unknown";
+   }
+}
+
+static int
+progress_phase_from_workflow_name(char* name)
```
Amr-Shams, line 763: why not adding a new field in the workflow called phase? would it reduce the fragile of this func?

src/libpgmoneta/prometheus.c
```diff
@@ -2659,43 +2659,121 @@ general_information(prometheus_metrics_container_t* container)
...
+   data = pgmoneta_append(data, "#HELP pgmoneta_backup_progress_elapsed_seconds The elapsed seconds since backup started\n");
+   data = pgmoneta_append(data, "#TYPE pgmoneta_backup_progress_elapsed_seconds gauge\n");
+   for (int i = 0; i < config->common.number_of_servers; i++)
+   {
+      data = pgmoneta_append(data, "pgmoneta_backup_progress_elapsed_seconds{name=\"");
       data = pgmoneta_append(data, config->common.servers[i].name);
-      data = pgmoneta_append(data, "\", type=\"remaining\"} ");
-      data = pgmoneta_append_int(data, (int)remaining);
+      data = pgmoneta_append(data, "\"} ");
+      data = pgmoneta_append_ulong(data, (unsigned long)atomic_load(&config->common.servers[i].backup_progress.elapsed));
```
Amr-Shams, line 2771: The ETA is now computed purely from percentage and elapsed. when the first is 0 at the wormup the later would always be 0, thinking about supressing it till the prct >0

Amr-Shams (commented):
this is a huge pr, it is a clear hard working, just some follow ups, i tried to help here, a geunine question, is there a better way than injecting almost all the core funs with the progress?, or is it fine for the level of the workflow prgress? like each phase/step can be completed done or not and update the progress accordingly?

src/libpgmoneta/prometheus.c
Abdelrhmansersawy: @Amr-Shams 
I'm thinking about the case where the first update is late. For example, base-backup sends the request, then computes the total tablespace size, so the first 1% takes time. The operation has already started, but progress is still 0%.
So the user should be aware of this, because if it stays at 0% for a long time, it might look like a problem (e.g. an internet issue with S3 progress).

src/libpgmoneta/workflow.c
Abdelrhmansersawy: I’m thinking that progress is optional, so we don’t need to enforce it across the workflow. I think it makes sense to keep it on ART as a node.

WDYT?

src/libpgmoneta/aes.c
Abdelrhmansersawy: Good catch ! Since we’re now putting the infrastructure for the progress framework, it makes sense to move all functions to the workflow level.

src/include/compression.h
Abdelrhmansersawy: That problem is that progress is optional, so before competing anything related to it - we need to check first if it's enabled so we need to aware of server index.

jesperpedersen: Progress should be global and per server override

Amr-Shams: well chcking if enabled or not, can be a utility or a helper method, cause if it is confiugred later on teh sever level then this would make a min change

Amr-Shams (commented):
great work bro !!, just final follow up from myside here

src/include/compression.h
Abdelrhmansersawy: @Amr-Shams `pgmoneta_is_progress_enabled(int server)` is already check if progress is enabled for this server or not (allow global and per server override)

Abdelrhmansersawy:
@jesperpedersen 
As a few follow-ups:
1. Improved progress phase weighting - it is now adaptive (if a previous backup exists, it reads the elapsed time of each phase and computes weights based on that).
2. Moved progress to be at workflow level instead of backup.c
3. Tried to make the API stable and more generalized so it can be reused by other commands later.

Abdelrhmansersawy:
I think this CI [build-linux (clang, Debug, 14)](https://github.com/pgmoneta/pgmoneta/actions/runs/24639991832/job/72042176126?pr=1097#logs) takes too long to run (~2 hours), so it fails, which is unrelated to my changes.

src/libpgmoneta/utils.c
```diff
@@ -1810,6 +1812,88 @@ pgmoneta_directory_size(char* directory)
    return total_size;
 }
 
+static int count_files_in_dir(char* directory);
+
+int
+pgmoneta_count_files(char* directory)
```
Amr-Shams, line 1816: you can simplify the logic here by using `pgmoneta_manifest_get_paths` in manifest.c

jesperpedersen:
@Abdelrhmansersawy @Amr-Shams PTAL me once this done

Abdelrhmansersawy:
@Abdelrhmansersawy 
You can take a look now.

src/include/management.h
jesperpedersen: Sorting

src/include/pgmoneta.h
```diff
@@ -357,13 +370,13 @@ struct server
    int workers;                                                   /**< The number of workers */
    int max_rate;                                                  /**< Maximum backup rate in bytes per second. */
    int number_of_extra;                                           /**< The number of source directory*/
-   int progress;                                                  /**< The Backup progress status */
+   int progress_enabled;                                          /**< The progress status */
```
jesperpedersen, line 352: bool ?

src/include/workflow.h
```diff
@@ -87,6 +86,13 @@ extern "C" {
 #define NODE_TARGET_FILE                 "target_file"         /* The target file */
 #define NODE_TARGET_ROOT                 "target_root"         /* The target root directory */
 
+#define NODE_PROGRESS_LIMIT_BACKUP       "progress_limit_backup"      /* Backup phase limit */
```
jesperpedersen: Sort

src/libpgmoneta/compression.c
```diff
@@ -27,6 +27,7 @@
  */
 
 #include <aes.h>
+#include <backup.h>
```
jesperpedersen: Why ?

jesperpedersen (changes requested):

jesperpedersen:
@Abdelrhmansersawy Lets start with those

src/include/pgmoneta.h
Abdelrhmansersawy: we need it to be as int, because:
1. `-1` : server progress not configured
2. `0` : server progress is set to false
2. `1` : server progress is set to true.

So we used it when compared with `bool progress` under pgmoneta section.

jesperpedersen: Yes, but default should `false`

jesperpedersen: We can make per-server overrides

Abdelrhmansersawy: @jesperpedersen

> We can make per-server overrides

Yes, that part is already in place. We have `pgmoneta_is_progress_enabled(int server)` which decides whether progress is enabled for a server based on both the global `config->progress` and `servers[server].progress_enabled`.

> Yes, but default should false

The only reason I went with `-1` as a default is to represent “not configured” for that specific server.

If we use a `bool` with default `false`, we won’t be able to tell the difference between these two cases:

```
[pgmoneta]
progress = on

[primary]
progress = off
```
progress shouldn't be run on `primary` server

and

```
[pgmoneta]
progress = on

[primary]
```
progress should be run on `primary` server

In both, `false` would mean two different things (explicitly disabled vs not set). That’s why I used a third state (`-1`) to keep that distinction clear.

Abdelrhmansersawy:
@jesperpedersen 
PTAL - failed testcases are unrelated to my changes (due of last merged commit)

src/include/pgmoneta.h
jesperpedersen: @Abdelrhmansersawy Yes, the `progress_enabled` in configuration is `bool`. The one in server is `int`

src/include/workflow.h
```diff
@@ -172,6 +178,97 @@ pgmoneta_common_setup(char* name, struct art* nodes);
 int
 pgmoneta_common_teardown(char* name, struct art* nodes);
 
+/**
```
jesperpedersen, line 209: Lets do a progress.h | .c - and move the `#define`s

Abdelrhmansersawy:
@jesperpedersen PTAL

src/include/progress.h
```diff
@@ -0,0 +1,179 @@
...
+#define PHASE_ENCRYPTION                6
+
+#define NODE_PROGRESS_LIMIT_BACKUP      "progress_limit_backup"      /* Backup phase limit */
+#define NODE_PROGRESS_LIMIT_COMPRESSION "progress_limit_compression" /* Compression phase limit */
+#define NODE_PROGRESS_LIMIT_ENCRYPTION  "progress_limit_encryption"  /* Encryption phase limit */
+#define NODE_PROGRESS_LIMIT_LINK        "progress_limit_link"        /* Link phase limit */
+#define NODE_PROGRESS_LIMIT_MANIFEST    "progress_limit_manifest"    /* Manifest phase limit */
+#define NODE_PROGRESS_LIMIT_SHA512      "progress_limit_sha512"      /* SHA512 phase limit */
+
+struct art;
```
jesperpedersen: `#include`

src/include/progress.h
```diff
@@ -0,0 +1,179 @@
...
+
+#define PHASE_NONE                      0
+#define PHASE_BASEBACKUP                1
+#define PHASE_MANIFEST                  2
+#define PHASE_SHA512                    3
+#define PHASE_LINKING                   4
+#define PHASE_COMPRESSION               5
+#define PHASE_ENCRYPTION                6
+
+#define NODE_PROGRESS_LIMIT_BACKUP      "progress_limit_backup"      /* Backup phase limit */
```
jesperpedersen, line 47: Doxygen doc

src/include/progress.h
```diff
@@ -0,0 +1,179 @@
...
+ */
+int
+pgmoneta_progress_phase_from_workflow_name(const char* name);
+
+/**
+ * ART node key storing the cumulative percentage limit for a phase
+ * @param phase The phase constant
+ * @return The key string, or NULL if unknown
+ */
+const char*
```
jesperpedersen: We don't use `const`

src/include/progress.h
```diff
@@ -0,0 +1,179 @@
...
+ */
+void
+pgmoneta_progress_teardown(int server);
+
+/**
+ * Get display name for a progress phase
+ * @param phase The phase constant
+ * @return The display name
+ */
+const char*
```
jesperpedersen: Same

src/libpgmoneta/prometheus.c
```diff
@@ -2708,43 +2710,95 @@ general_information(prometheus_metrics_container_t* container)
    add_metric_to_art(container->general_metrics, "pgmoneta_extension_pgmoneta_ext", data, NULL, NULL, 0);
    free(data);
    data = NULL;
-   data = pgmoneta_append(data, "#HELP pgmoneta_backup_progress The backup progress for a server\n");
-   data = pgmoneta_append(data, "#TYPE pgmoneta_backup_progress gauge\n");
+   data = pgmoneta_append(data, "#HELP pgmoneta_progress_percentage The backup progress percentage for a server\n");
```
jesperpedersen: We need to think about this... because this won't work. Maybe we can just remove "backup".

There is a progress for restore as well - it should be reflected...

src/libpgmoneta/prometheus.c
```diff
@@ -2708,43 +2710,95 @@ general_information(prometheus_metrics_container_t* container)
...
+   }
+   data = pgmoneta_append(data, "\n");
 
-      data = pgmoneta_append(data, "pgmoneta_backup_progress{");
-      data = pgmoneta_append(data, "name=\"");
+   add_metric_to_art(container->general_metrics, "pgmoneta_progress_percentage", data, NULL, NULL, 0);
+   free(data);
+   data = NULL;
+
+   data = pgmoneta_append(data, "#HELP pgmoneta_progress_elapsed_time The elapsed seconds since backup started\n");
```
jesperpedersen: Same issue

src/libpgmoneta/status.c
```diff
@@ -592,10 +592,22 @@ pgmoneta_progress(SSL* ssl, int client_fd, uint8_t compression, uint8_t encrypti
...
+   pgmoneta_json_put(response, MANAGEMENT_ARGUMENT_PROGRESS_STATE, (uintptr_t)atomic_load(&config->common.servers[srv].progress.state), ValueInt32);
+   pgmoneta_json_put(response, MANAGEMENT_ARGUMENT_DONE, (uintptr_t)atomic_load(&config->common.servers[srv].progress.done), ValueInt64);
+   pgmoneta_json_put(response, MANAGEMENT_ARGUMENT_TOTAL, (uintptr_t)atomic_load(&config->common.servers[srv].progress.total), ValueInt64);
+   pgmoneta_json_put(response, MANAGEMENT_ARGUMENT_ELAPSED, (uintptr_t)atomic_load(&config->common.servers[srv].progress.elapsed), ValueInt64);
+
+   int pct = atomic_load(&config->common.servers[srv].progress.percentage);
+   int64_t elapsed_secs = atomic_load(&config->common.servers[srv].progress.elapsed);
+   pgmoneta_time_t remaining = PGMONETA_TIME_DISABLED;
+
+   if (pct > 0 && pct < 100)
```
jesperpedersen: We need an `else`

src/libpgmoneta/workflow.c
```diff
@@ -329,18 +331,64 @@ pgmoneta_workflow_execute(struct workflow* workflow, struct art* nodes,
...
+
+      if (progress_enabled)
+      {
+         int phase = pgmoneta_progress_phase_from_workflow_name(current->name());
+         if (phase > 0)
+         {
+            const char* key = pgmoneta_progress_limit_node_key(phase);
+            if (key != NULL && pgmoneta_art_contains_key(nodes, (char*)key))
+            {
+               struct main_configuration* config = (struct main_configuration*)shmem;
```
jesperpedersen: Resolve at the beginning of the function

src/libpgmoneta/workflow.c
```diff
@@ -436,6 +484,42 @@ pgmoneta_common_teardown(char* name, struct art* nodes)
    return 0;
 }
 
+const char*
+pgmoneta_workflow_name(int workflow_type)
+{
+   switch (workflow_type)
+   {
+      case WORKFLOW_TYPE_BACKUP:
+         return "Backup";
```
jesperpedersen: We need `#define` for this - or better use its `name()`

jesperpedersen (changes requested):

jesperpedersen:
@Abdelrhmansersawy Next round - look across the entire patch

Abdelrhmansersawy:
@jesperpedersen PTAL.

src/libpgmoneta/configuration.c
```diff
@@ -3784,7 +3784,7 @@ add_servers_configuration_response(struct json* res)
       pgmoneta_json_put(server_conf, CONFIGURATION_ARGUMENT_HOT_STANDBY_TABLESPACES, (uintptr_t)config->common.servers[i].hot_standby_tablespaces, ValueString);
       pgmoneta_json_put(server_conf, CONFIGURATION_ARGUMENT_WORKERS, (uintptr_t)config->common.servers[i].workers, ValueInt64);
       pgmoneta_json_put(server_conf, CONFIGURATION_ARGUMENT_MAX_RATE, (uintptr_t)config->common.servers[i].max_rate, ValueInt64);
-      pgmoneta_json_put(server_conf, CONFIGURATION_ARGUMENT_PROGRESS, (uintptr_t)config->common.servers[i].progress, ValueInt64);
+      pgmoneta_json_put(server_conf, CONFIGURATION_ARGUMENT_PROGRESS, (uintptr_t)config->common.servers[i].progress_enabled, ValueInt64);
```
jesperpedersen: ValueBool

src/libpgmoneta/utils.c
```diff
@@ -1810,6 +1812,42 @@ pgmoneta_directory_size(char* directory)
...
+
+   memset(manifest_path, 0, MAX_PATH);
+   pgmoneta_snprintf(manifest_path, MAX_PATH, "%s/backup.manifest", directory);
+
+   if (pgmoneta_exists(manifest_path))
+   {
+      if (!pgmoneta_manifest_get_paths(manifest_path, &paths))
+      {
+         num = pgmoneta_deque_size(paths);
+         pgmoneta_deque_destroy(paths);
```
jesperpedersen: One cleanup place, one return place

jesperpedersen (changes requested):

jesperpedersen:
@Abdelrhmansersawy You need to fix all the workflow::name callbacks to use the `#define`s

Abdelrhmansersawy:
@jesperpedersen PTAL

src/include/pgmoneta.h
```diff
@@ -45,6 +45,8 @@ extern "C" {
 #include <sys/types.h>
 #include <openssl/ssl.h>
 
+#include <progress.h>
```
jesperpedersen: Why ?

src/include/progress.h
```diff
@@ -0,0 +1,174 @@
...
+extern "C" {
+#endif
+
+/* system */
+#include <stdatomic.h>
+#include <stdbool.h>
+#include <stdint.h>
+
+/* pgmoneta */
+#include <art.h>
```
jesperpedersen: Ours are first

src/libpgmoneta/progress.c
```diff
@@ -0,0 +1,569 @@
...
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT
+ * OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR
+ * TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/* pgmoneta */
+#include <pgmoneta.h>
+#include <progress.h>
```
jesperpedersen: Sort

src/libpgmoneta/progress.c
```diff
@@ -0,0 +1,569 @@
...
+   atomic_store(&config->common.servers[server].progress.state, WORKFLOW_PROGRESS_RUNNING);
+   atomic_store(&config->common.servers[server].progress.workflow_type, workflow_type);
+   atomic_store(&config->common.servers[server].progress.current_phase, phases[0]);
+   atomic_store(&config->common.servers[server].progress.done, 0);
+   atomic_store(&config->common.servers[server].progress.total, 0);
+   atomic_store(&config->common.servers[server].progress.elapsed, 0);
+   atomic_store(&config->common.servers[server].progress.start_time, time(NULL));
+   atomic_store(&config->common.servers[server].progress.percentage, 0);
+
+   pgmoneta_log_info("Progress: Workflow \"%s\" started for server %s",
```
jesperpedersen: debug

src/libpgmoneta/progress.c
```diff
@@ -0,0 +1,569 @@
...
+   if (nodes != NULL && phase > 0)
+   {
+      char* key = pgmoneta_progress_limit_node_key(phase);
+      assert(key);
+      assert(pgmoneta_art_contains_key(nodes, key));
+
+      atomic_store(&p->phase_limit, (int)(uintptr_t)pgmoneta_art_search(nodes, key));
+   }
+
+   pgmoneta_log_info("Progress: Phase \"%s\" started for server %s",
```
jesperpedersen: debug

src/libpgmoneta/progress.c
```diff
@@ -0,0 +1,569 @@
...
+{
+   struct main_configuration* config;
+   struct progress* p;
+   int workflow_type;
+
+   config = (struct main_configuration*)shmem;
+   p = &config->common.servers[server].progress;
+   workflow_type = atomic_load(&p->workflow_type);
+
+   pgmoneta_log_info("Progress: Workflow \"%s\" completed for server %s",
```
jesperpedersen: debug

src/libpgmoneta/utils.c
```diff
@@ -36,6 +36,7 @@
 #include <logging.h>
 #include <shmem.h>
 #include <utils.h>
+#include <manifest.h>
```
jesperpedersen: Sort

src/libpgmoneta/wf_encryption.c
```diff
@@ -28,6 +28,7 @@
 
 /* pgmoneta */
 #include <pgmoneta.h>
+#include <progress.h>
```
jesperpedersen: Sort

src/libpgmoneta/wf_gzip.c
```diff
@@ -28,6 +28,7 @@
 
 /* pgmoneta */
 #include <pgmoneta.h>
+#include <progress.h>
```
jesperpedersen: Sort

src/libpgmoneta/wf_lz4.c
```diff
@@ -28,6 +28,7 @@
 
 /* pgmoneta */
 #include <pgmoneta.h>
+#include <progress.h>
```
jesperpedersen: Sort

src/libpgmoneta/wf_sha512.c
```diff
@@ -28,6 +28,7 @@
 
 /* pgmoneta */
 #include <pgmoneta.h>
+#include <progress.h>
```
jesperpedersen: Sort

src/libpgmoneta/wf_zstd.c
```diff
@@ -28,6 +28,7 @@
 
 /* pgmoneta */
 #include <pgmoneta.h>
+#include <progress.h>
```
jesperpedersen: Sort

src/libpgmoneta/workflow.c
```diff
@@ -28,9 +28,12 @@
 
 /* pgmoneta */
 #include <pgmoneta.h>
+#include <progress.h>
```
jesperpedersen: Sort

src/libpgmoneta/workflow.c
```diff
@@ -329,18 +332,62 @@ pgmoneta_workflow_execute(struct workflow* workflow, struct art* nodes,
       current = current->next;
    }
 
+   int server = -1;
```
jesperpedersen: Move to start

test/testcases/test_progress.c
```diff
@@ -0,0 +1,596 @@
...
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT
+ * OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR
+ * TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ */
+
+#include <pgmoneta.h>
+#include <progress.h>
```
jesperpedersen: Sort

jesperpedersen (changes requested):

jesperpedersen:
@Abdelrhmansersawy You should have caught most of these...

src/include/pgmoneta.h
Abdelrhmansersawy: @jesperpedersen 
Because `struct progress` is defined on `progress.sh` which include on `struct server`.

Should I move `struct progress` to be at `pgmoneta.h` ?

jesperpedersen: @Abdelrhmansersawy No, just leave it then

Abdelrhmansersawy:
@jesperpedersen PTAL

Abdelrhmansersawy:
@jesperpedersen 

About "Sorting of include files", I'm thinking about having a Python script that detects them and provides a `--fix` option (since we couldn’t do this through `clang-format` as we have our specific order). I believe this could help save review time on small things.

I have a draft Python script - if you think this makes sense, I could open a PR for it.

jesperpedersen:
@Abdelrhmansersawy I think we should stick to `clang-format` - there are some options to do sorting - we need to explore that

Abdelrhmansersawy:
> @Abdelrhmansersawy I think we should stick to `clang-format` - there are some options to do sorting - we need to explore that

I would navigate it and find the most suitable option for our purpose.

jesperpedersen:
```
test/include/tscommon.h:167: error: Member test_home[MAX_PATH] (variable) of struct test_encryption_env is not documented.
test/include/tscommon.h:168: error: Member original_home (variable) of struct test_encryption_env is not documented.
test/include/tscommon.h:169: error: Member original_config_home[MAX_PATH] (variable) of struct test_encryption_env is not documented.
test/include/tscommon.h:170: error: Member original_encryption (variable) of struct test_encryption_env is not documented.
test/include/tscommon.h:171: error: Member shmem_locally_allocated (variable) of struct test_encryption_env is not documented.
```

Please, compile with documentation on

Abdelrhmansersawy:
@jesperpedersen PTAL

jesperpedersen:
Merged.

jesperpedersen:
@Abdelrhmansersawy Please, create an issue for each of the TODOs

jesperpedersen:
@Amr-Shams Please, go through this

Abdelrhmansersawy:
@jesperpedersen 
About cross-porting to pgvictoria, should we do it now or later? (Ping me when needed)

jesperpedersen:
@Abdelrhmansersawy Lets wait with pgvictoria

Issue #1092, opened 2026-04-11
walinfo/Interactive: Description search


jesperpedersen:
Improve description search using `-m` / `-t` from within the UI.

We need to expand the documentation around these 2 options as well

saraanbih:
Can i take it?

jesperpedersen:
@saraanbih Go for it !

PR #1102, opened 2026-04-16
[#1092] Fix OID mapping translation

saraanbih:
- fixes OID-to-name translation in pgmoneta-walinfo when using -m mapping.json. 
- updated docs for this part

Fixes [#1092]

Before 
before

---

After
after


jesperpedersen:
@saraanbih Excellent, but we need more documentation on how to make the `-m` file

saraanbih:
@jesperpedersen 
> @saraanbih Excellent, but we need more documentation on how to make the -m file

Done

jesperpedersen:
Merged.

jesperpedersen:
@saraanbih This is way too difficult - we need a command in walinfo that can generate this... Please, investigate and type it up

saraanbih:
@jesperpedersen 
> @saraanbih This is way too difficult - we need a command in walinfo that can generate this... Please, investigate and type it up

any thoughts?

jesperpedersen:
@saraanbih You have to add a subcommand - maybe `pgmoneta-walinfo generate` ?

Also, we have to look at if we really need both `-m` and `-t`... Can we make that simpler ?

PR #1108, opened 2026-04-17
[#871] Add HTTP DELETE method for S3 storage engine

Hanu-man12:
Add PGMONETA_HTTP_DELETE definition to support S3 delete operations. This is part of the Storage Engine Operations Parity work to enable list/read/delete operations for remote backends.

Changes:
- Add PGMONETA_HTTP_DELETE (3) to http.h
- Add DELETE case to http_method_to_string() in http.c

### DIFF
--- a/src/include/http.h
+++ b/src/include/http.h
@@ -47,6 +47,7 @@ extern "C" {
 #define PGMONETA_HTTP_GET  0
 #define PGMONETA_HTTP_POST 1
 #define PGMONETA_HTTP_PUT  2
+#define PGMONETA_HTTP_DELETE 3
 
 /* HTTP status codes */
 #define PGMONETA_HTTP_STATUS_OK    0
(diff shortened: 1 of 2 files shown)


Amr-Shams (commented):
there is no need for delete request, it is not used in the s3 protocol
to delete an object we create a put request

Hanu-man12:
Thanksfor the review I understand S3 batch delete uses POST with XML body.However, the delete methodmight be useful for:
1. Single object deleteion via DELETE /bucket/object
2. Future Azure Blob Storage support(uses DELETE)
3. Consistency with other HTTP-based backends 


Should I modify this PR to focus on azure support instead, or close it?

Amr-Shams:
> Thanksfor the review I understand S3 batch delete uses POST with XML body.However, the delete methodmight be useful for:
> 
> 1. Single object deleteion via DELETE /bucket/object
> 2. Future Azure Blob Storage support(uses DELETE)
> 3. Consistency with other HTTP-based backends
> 
> Should I modify this PR to focus on azure support instead, or close it?

well, a single object deletion is not smth a backup needs to do, i think this might be closed for now and reintroudeced when working with azure in the future

Hanu-man12:
Thanks for the clarification!

Understood — I will close this PR for now and continue working on other parts of the storage engine.

Thanks!

Issue #980, opened 2026-03-04
FIPS


jesperpedersen:
For PostgreSQL 18+ if pgacrypto is installed then we need to report its status.

If < 18 see if we can figure out a way to report it.

Default is `false`

Amr-Shams:
Can’t we just check pg_extension per DB and return true/false? If it fails, default to false. Not sure we need PG 18 specific logic unless something changed. I have checked pg docs and found it supported since 9.1+v

jesperpedersen:
@Amr-Shams I have no idea what you are talking about. pgcrypto in PostgreSQL 18 has a function to report, otherwise we have to figure it out on our own

Amr-Shams:
I went straight for the solution: on PG 18+, call the new pgcrypto reporting function for real runtime/FIPS status; on older versions, fallback to checking pg_extension or return false if undetectable.

jesperpedersen:
@Amr-Shams Yes, for PostgreSQL 18+ we first check that pgcrypto is installed. If so, we call `fips_mode()`. In all other scenarios we need to do the checks that `fips_mode` does - if possible

trxvorr:
@jesperpedersen for pg < 18 we can just test the runtime status by intentionally executing a dummy MD5 hash via pgcrypto:

SELECT digest('test', 'md5');

if pgcrypto isnt installed at all, or if the query succeeds, we just default to false

jesperpedersen:
@trxvorr We know it from `SELECT * FROM pg_extension;`

trxvorr:
@jesperpedersen ah so something like

SELECT * FROM pg_extension WHERE extname = 'pgcrypto';

then

if pg >= 18: Execute SELECT fips_mode();

if pg < 18: Execute SELECT digest('test', 'md5');

jesperpedersen:
@trxvorr And, then for < 18 do all the checks manually. The requirements has to be documented

sksingh2005:
Can I pick this one?  Was going through it for a while.

jesperpedersen:
@sksingh2005 Go for it !

jesperpedersen:
@sksingh2005 Look at https://github.com/pgexporter/pgexporter/commit/e4fa1192148c30ba9a74632ebeecf8b6087bab56 and related commits

PR #1112, opened 2026-04-18
[#980] Add FIPS mode reporting

sksingh2005:


sksingh2005:
@jesperpedersen sir I think the CI failure is due to a Rocky Linux 10 repo inconsistency  `libarchive` and `libarchive-devel` are at mismatched versions across BaseOS and AppStream mirrors, causing some matrix jobs to fail at the  install dependencies step while others pass. Should I file a separate issue for this ?

src/libpgmoneta/fips.c
```diff
@@ -0,0 +1,236 @@
...
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+   return EVP_default_properties_is_fips_enabled(NULL) == 1;
+#elif OPENSSL_VERSION_NUMBER >= 0x10100000L
+   return FIPS_mode() == 1;
+#endif
+   return false;
+}
+
+int
+pgmoneta_fips_server(int server, bool* status)
```
Amr-Shams: why introudcing a new ssl/auth when called from `pgmoneta_server_info()` which already has active ssl/socket?

src/libpgmoneta/fips.c
```diff
@@ -0,0 +1,236 @@
...
+}
+
+static int
+query_fips_ext(SSL* ssl, int socket, struct query_response** qr)
+{
+   return query_execute(ssl, socket, "SELECT pgmoneta_ext_fips();", qr);
+}
+
+static int
+query_execute(SSL* ssl, int socket, char* qs, struct query_response** qr)
```
Amr-Shams, line 156: use the extension query execute func, already there

Amr-Shams (commented):
Great PR, nice work !! i just have a few genuine questions that might help clarify some parts from my side

src/libpgmoneta/fips.c
sksingh2005: @Amr-Shams `query_execute()` in `extension.c` is static, not accessible from `fips.c`. `server.c` and `extension.c` both have their own static copies of this helper. Making it public would mean chaging `extension.h/extension.c`

src/libpgmoneta/fips.c
sksingh2005: Good point the caller connection is a regular SQL connection to postgres and could be reused. I tried to follow the same pattern as `pgmoneta_detect_server_extensions()`  which also creates its own connection from the same call site. 
@jesperpedersen sir should I refactor and prefer passing ssl/socket through ?

jesperpedersen: @sksingh2005 We should use one connection across all the startup checks

jesperpedersen:
Rebased, and merged.

PR #1132, opened 2026-05-04
[#1133] Handle ErrorResponse during SCRAM authentication

### LINKED ISSUE #1133: Segfault when handling ErrorResponse (‘E’) during SCRAM authentication
**Describe**
During SCRAM authentication, the client may receive an ErrorResponse ('E')
message from the server. However, the current implementation continues
to parse SCRAM attributes even when the message type is 'E'.

This results in NULL values (e.g., base64_salt) being used, which can
lead to segmentation faults.

ankushT369:
fixed #1133 Avoid parsing SCRAM attributes when server returns 'E', which could lead to NULL values (e.g. base64_salt) and segfaults.

### DIFF
--- a/src/libpgmoneta/security.c
+++ b/src/libpgmoneta/security.c
@@ -505,6 +505,10 @@ pgmoneta_remote_management_scram_sha256(char* username, char* password, int serv
    }
 
    sasl_continue = pgmoneta_copy_message(msg);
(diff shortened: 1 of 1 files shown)


jesperpedersen:
@ankushT369 You need your name in the authors files - see DEVELOPERS.md

Issue #1119, opened 2026-04-25
Test cases: Tablespace


Jubilee101:
Add test cases for tablespaces. 

This should work for both full backup and incremental backup with Postgres 17+ (Postgres 14 - 16 incremental backup doesn't support tablespaces for now).

The tablespace dir in backup data should contain correct symlinks to the corresponding tablespace directories after either backup or restore.

sksingh2005:
Can i pick this one ?

Jubilee101:
@sksingh2005 Of course~ Have fun with that!

PR #1139, opened 2026-05-08
[#1119] Add test cases for tablespaces

sksingh2005:


jesperpedersen:
@ashu3103 Please, help here

ashu3103:
> @ashu3103 Please, help here

Sure, on it

jesperpedersen:
@ashu3103 Can you go through this ?

src/libpgmoneta/aes.c
```diff
@@ -535,11 +535,11 @@ decrypt_data(char* d, struct workers* workers, struct deque* excludes)
 
    while ((entry = readdir(dir)) != NULL)
    {
-      if (entry->d_type == DT_DIR || entry->d_type == DT_LNK)
+      if (entry->d_type == DT_DIR)
       {
          char path[1024];
 
-         if (strcmp(entry->d_name, ".") == 0 || strcmp(entry->d_name, "..") == 0)
+         if (strcmp(entry->d_name, ".") == 0 || strcmp(entry->d_name, "..") == 0 || strcmp(entry->d_name, "pg_tblspc") == 0)
```
Jubilee101: Why? pg_tblspc is a dir with nothing but symlinks the previous check on DT_LNK should work

src/libpgmoneta/aes.c
```diff
@@ -535,11 +535,11 @@ decrypt_data(char* d, struct workers* workers, struct deque* excludes)
 
    while ((entry = readdir(dir)) != NULL)
    {
-      if (entry->d_type == DT_DIR || entry->d_type == DT_LNK)
```
Jubilee101, line 538: Why removing DT_LNK?

src/libpgmoneta/archive.c
```diff
@@ -320,6 +327,7 @@ pgmoneta_receive_archive_files(int srv, SSL* ssl, int socket, struct stream_buff
             if (pgmoneta_compare_string(tup->data[1], tblspc->path))
             {
                tblspc->oid = atoi(tup->data[0]);
+               pgmoneta_snprintf(manifest_prefix, sizeof(manifest_prefix), "pg_tblspc/%d", tblspc->oid);
```
Jubilee101: Is it because the untar process isn't reading the file path from pg_tblspc dir correclty and causing manifest verification failure? Can you giev an example?

src/libpgmoneta/archive.c
```diff
@@ -342,6 +350,17 @@ pgmoneta_receive_archive_files(int srv, SSL* ssl, int socket, struct stream_buff
...
+         msg = (struct message*)malloc(sizeof(struct message));
+         if (msg == NULL)
+         {
+            goto error;
+         }
+         memset(msg, 0, sizeof(struct message));
+      }
+
       // get the copy out response
       while (msg == NULL || msg->kind != 'H')
```
Jubilee101: Hmm, did test fail because of this? Nice catch. Something is wrong with this, this looks like a regression issue.

src/libpgmoneta/archive.c
```diff
@@ -342,6 +350,17 @@ pgmoneta_receive_archive_files(int srv, SSL* ssl, int socket, struct stream_buff
          pgmoneta_log_error("Could not create archive tar file");
          goto error;
       }
+
+      if (msg == NULL)
```
Jubilee101: This is a bit too ad hoc. I think we can reuse the same msg throughout the entire process. So at the end of out most iteration we should just do `memset(msg, 0, ...)` instead of `pgmoneta_free_message(msg);`. That'll save us some malloc.

test/check.sh
```diff
@@ -86,6 +86,34 @@ detect_container_engine() {
   fi
 }
 
+make_tablespace_cleanup_writable() {
```
Jubilee101, line 89: Does this have to be run time? Can we put this into Dockerfile? And should we?

Jubilee101 (commented):
Amazing work! Very thorough and looks like it indeed caught some issues. Left some questions. My only concern is that some fixes seems a bit ad hoc.

test/testcases/test_tablespaces.c
Jubilee101, line 1: I wonder if it's better to move the test cases to test_backup and test_restore, respectively.

src/libpgmoneta/aes.c
sksingh2005: You’re right. Hard-coding `pg_tblspc` was not needed. The real issue is avoiding symlink traversal during decrypt. I will update this to skip `DT_LNK` entries directly, so we no longer need the `pg_tblspc` name check.

src/libpgmoneta/aes.c
sksingh2005: Ohh sorry, removing `DT_LNK` was too broad. I will change it to keep symlink handling explicit, but skip symlink traversal before recursing. I think that will then prevents duplicate decrypt work through `pg_tblspc/<oid>`.

src/libpgmoneta/archive.c
sksingh2005: Yes, the PG14 failure matched this path. In PG14  multiple tar streams. The old code freed/reset `msg` after one stream, and the next stream could hit bad state. The log showed `Connection reset by peer`, then the next test failed because the server stayed active.

src/libpgmoneta/archive.c
sksingh2005: Agreed. I will update this to reuse the same `msg` and just call:

`memset(msg, 0, sizeof(struct message));`

at the end of each tar iteration instead of freeing/reallocating it.

test/testcases/test_tablespaces.c
sksingh2005: That’s a valid option. I kept them in `test_tablespaces.c` because the tests share tablespace setup, symlink checks, and cleanup, and it lets us run only this coverage with `-m tablespaces`. I can move backup checks to `test_backup.c` and restore checks to `test_restore.c` if you want . Let me know :)

src/libpgmoneta/archive.c
sksingh2005: Yes, this is for manifest verification. Tablespace tar entries are relative, for example:
`PG_17_202406281/16385/16390`

But the backup manifest stores the same file as:

`pg_tblspc/<oid>/PG_17_202406281/16385/16390`

So without prefixing `pg_tblspc/<oid>`, checksum/size verification uses the wrong key.

test/check.sh
sksingh2005: I think Dockerfile alone won’t help here because `/conf` is bind-mounted from `$PGCONF_DIRECTORY`, so image-time permissions are hidden by the host mount. This cleanup handles stale host-side tablespace dirs from previous runs, where files may be owned by the container postgres user and block `rm -Rf "$PGMONETA_ROOT_DIR"`. We can keep creation in `run-postgresql`, but cleanup still needs to happen in `check.sh`. What are your suggestions ?

src/libpgmoneta/aes.c
Jubilee101: Hmm, doesn't sound right. Decryption through symlink is the correct way to go (I added that commit a year ago). Can you elaborate on how duplicate decryption happened? You should be removing the other one, this is the right place where encryption should happen for tablespace dirs.

src/libpgmoneta/archive.c
Jubilee101: Thanks for catching this. This while loop condition is still wrong, I have pushed a commit to fix this, kindly rebase.

src/libpgmoneta/archive.c
Jubilee101: Edit: Or maybe let's just pass in a `tblspc_oid` directly. The key is to move the prefix building inside the extraction function. 

---------------------------------------------------------------------------------------------------------------------------------

Nice catch. But instead of `manifest_prefix` I'd prefer passing in a `backup_tar_name` to `pgmoneta_extract_backup_tar_file`. For main backup it's "base" (feel free to pass in NULL), for tablespaces the name is the OID. Then inside `pgmoneta_extract_backup_tar_file` we can assemble the prefix when name isn't "base" (or NULL). That'll make more sense for this function. 

This aligns with the official document:
> ...
> Write the output as tar files in the target directory. The main data directory's contents will be written to a file named base.tar, and each other tablespace will be written to a separate tar file named after that tablespace's OID.
> ...

src/libpgmoneta/compression.c
```diff
@@ -431,6 +431,11 @@ process_directory_operation(int server, char* directory, int type, struct worker
 
       pgmoneta_snprintf(full_path, sizeof(full_path), "%s/%s", directory, entry->d_name);
 
+      if (strcmp(entry->d_name, "pg_tblspc") == 0)
```
Jubilee101: compression should go into pg_tblspc (and recursively compress whatever symlink points to), why skipping it?

test/testcases/test_tablespaces.c
Jubilee101: Let's split them to backup and restore. If you have functionality to reuse, feel free to add a new module in test/include and test/libpgmonetatest respectively. cmake should be able to pick it up automatically.

src/libpgmoneta/compression.c
Jubilee101: Like encryption, if something is causing duplicate compression/decryption, it's the other one that needs to be removed.

test/check.sh
Jubilee101: In that case we should perhaps create the tablespace dir locally inside `$PGCONF_DIRECTORY` and set it to 777 before mounting it to `/conf ` in the container, what do you think?

Jubilee101 (changes requested):
Sorry for the late response @sksingh2005 , got caught up by other stuff... I'm a bit concerned about the compression/encryption changes, please take another look, thanks.

src/libpgmoneta/archive.c
```diff
@@ -905,6 +916,11 @@ pgmoneta_extract_backup_tar_file(char* file_path, char* destination, struct art*
          pgmoneta_snprintf(dst_path, sizeof(dst_path), "%s/%s", destination, entry_path);
       }
 
+      if (type == AE_IFLNK && pgmoneta_ends_with(dst_path, "/"))
```
Jubilee101, line 923: Please add a comment as to why it's necessary to remove the trailing `/`. Otherwise it's quite confusing.

Jubilee101:
Also the CI breaks, can you take a look? @sksingh2005 Let me know if you need me to also take a look.

src/libpgmoneta/aes.c
sksingh2005: You’re right. I re-checked this and reverted my earlier approach. Decryption through the `pg_tblspc/<oid>` symlink is intentional and should remain the path for tablespace files. The duplicate processing concern came from the hot standby path, but the correct fix is not to skip symlinks here. I’ve restored `DT_DIR || DT_LNK` traversal.

src/libpgmoneta/compression.c
sksingh2005: You were right. I removed the `pg_tblspc` skip, so compression now recurses through `pg_tblspc` and follows the symlinked tablespace path.

The duplicate processing was from the other path: the top-level extracted `tblspc_*` directory. I now exclude those top-level tablespace dirs from compression/encryption workflows, so tablespace files are processed through `data/pg_tblspc/<oid>` only.

src/libpgmoneta/wf_bzip2.c
```diff
@@ -144,6 +144,7 @@ bzip2_execute_compress(char* name __attribute__((unused)), struct art* nodes)
       pgmoneta_deque_add(excludes, "backup.sha512", 0, ValueString);
       pgmoneta_deque_add(excludes, "backup.sha512.tmp", 0, ValueString);
       pgmoneta_deque_add(excludes, "backup.sha256", 0, ValueString);
+      pgmoneta_workflow_exclude_tablespaces(backup, excludes);
```
Jubilee101, line 147: Ah, looks like we are compressing the parent dir containing main data + tablespaces + meta data info. And we are excluding a lot of things because of that. That doesn't feel right. We should go to data/ directly and compress tablesapces through symlinks directly. Cc @jesperpedersen

Considering the purpose of this PR, let's keep your changes and do refactoring later.

Jubilee101: Created #1157 to track this.

src/libpgmoneta/archive.c
```diff
@@ -989,6 +1009,12 @@ pgmoneta_extract_backup_tar_file(char* file_path, char* destination, struct art*
          }
          while (asize > 0);
 
+         if (backup_tar_name != NULL && strlen(backup_tar_name) > 0 && strcmp(backup_tar_name, "base"))
```
Jubilee101, line 1012: Use pgmoneta_compare_string to skip the NULL/empty string checks.

Jubilee101:
@sksingh2005 There's a memleak in error path. I should have patched it, please rebase.

Reading the [CI logs](https://github.com/pgmoneta/pgmoneta/actions/runs/26418982930/artifacts/7204811531), there seems to be the same duplicated comrpession/encryption problem in hotstandby:

```
2026-05-25 20:53:37 ERROR aes.c:1381 encrypt_file: destination file /tmp/pgmoneta-test/standby/primary/pg_tblspc/16393/PG_17_202406281/16385/16397.zstd already exists
2026-05-25 20:53:38 WARN  utils.c:2412 pgmoneta_delete_file:  (No such file or directory)
2026-05-25 20:53:38 WARN  aes.c:674 do_decrypt_file: /tmp/pgmoneta-test/standby/primary/pg_tblspc/16393/PG_17_202406281/16385/16397.zstd.aes -> /tmp/pgmoneta-test/standby/primary/pg_tblspc/16393/PG_17_202406281/16385/16397.zstd
Backtrace:
#0  0x7f42cd845825 in pgmoneta_backtrace_string
	at /__w/pgmoneta/pgmoneta/src/libpgmoneta/utils.c:5286
#1  0x7f42cd6f5a81 in pgmoneta_log_line
	at /__w/pgmoneta/pgmoneta/src/libpgmoneta/logging.c:245
#2  0x7f42cd913512 in pgmoneta_workers_log_failures
	at /__w/pgmoneta/pgmoneta/src/libpgmoneta/workers.c:191
#3  0x7f42cd9136cd in pgmoneta_workers_transfer_failures
	at /__w/pgmoneta/pgmoneta/src/libpgmoneta/workers.c:207
#4  0x7f42cd8f08da in hot_standby_execute
	at /__w/pgmoneta/pgmoneta/src/libpgmoneta/wf_hot_standby.c:338
#5  0x7f42cd921248 in pgmoneta_workflow_execute
	at /__w/pgmoneta/pgmoneta/src/libpgmoneta/workflow.c:360
#6  0x7f42cd6240e9 in pgmoneta_backup
	at /__w/pgmoneta/pgmoneta/src/libpgmoneta/backup.c:226
#7  0x55aa0c6c6638 in accept_mgt_cb
	at /__w/pgmoneta/pgmoneta/src/main.c:1164
#8  0x7f42ce540cab in ev_invoke_pending
	at ??:?
#9  0x7f42ce54484f in ev_run
	at ??:?
#10  0x55aa0c6c4673 in ev_loop
	at /usr/include/ev.h:841
#11  0x55aa0c6bc2d6 in main
	at /__w/pgmoneta/pgmoneta/src/main.c:96
```

src/libpgmoneta/wf_bzip2.c
Jubilee101: On second thought, why does this matter? We have switched to streamer during backup which doesn't even use the workflow anymore.

jesperpedersen:
@sksingh2005 Lets focus on PostgreSQL 17 first

src/include/achv.h
```diff
@@ -87,10 +87,11 @@ pgmoneta_receive_archive_stream(int srv, SSL* ssl, int socket, struct stream_buf
  * @param destination The destination to extract to
  * @param checksums [out] The file checksums
  * @param sizes [out] The file sizes
+ * @param backup_tar_name The pg_basebackup tar name, base or tablespace OID
  * @return 0 upon success, otherwise 1
  */
 int
-pgmoneta_extract_backup_tar_file(char* file_path, char* destination, struct art* file_checksums, struct art* file_sizes);
+pgmoneta_extract_backup_tar_file(char* file_path, char* destination, struct art* file_checksums, struct art* file_sizes, char* backup_tar_name);
```
jesperpedersen, line 94: What is this for ? And an output parameter is `char**`

src/include/workflow.h
```diff
@@ -37,6 +37,8 @@ extern "C" {
 #include <art.h>
 #include <info.h>
 
+struct deque;
```
jesperpedersen, line 40: Remove, and `#include`

src/include/workflow.h
```diff
@@ -215,6 +217,14 @@ pgmoneta_common_setup(char* name, struct art* nodes);
 int
 pgmoneta_common_teardown(char* name, struct art* nodes);
 
+/**
+ * Add top-level tablespace directories to an exclude list
+ * @param backup The backup
+ * @param excludes The exclude list
+ */
+void
+pgmoneta_workflow_exclude_tablespaces(struct backup* backup, struct deque* excludes);
```
jesperpedersen, line 226: Why exclude ?

src/libpgmoneta/aes.c
```diff
@@ -184,11 +184,16 @@ pgmoneta_encrypt_data(int server, char* d, struct workers* workers, struct deque
 
    while ((entry = readdir(dir)) != NULL)
    {
-      if (entry->d_type == DT_DIR)
+      if (entry->d_type == DT_DIR || entry->d_type == DT_LNK)
```
jesperpedersen, line 187: Something is wrong...

src/libpgmoneta/archive.c
```diff
@@ -394,14 +404,12 @@ pgmoneta_receive_archive_files(int srv, SSL* ssl, int socket, struct stream_buff
       fclose(file);
 
       // extract the file
-      if (pgmoneta_extract_backup_tar_file(file_path, directory, file_checksums, file_sizes))
+      if (pgmoneta_extract_backup_tar_file(file_path, directory, file_checksums, file_sizes, backup_tar_name))
```
jesperpedersen, line 407: Yeah, this is `char**`

src/libpgmoneta/archive.c
```diff
@@ -394,14 +404,12 @@ pgmoneta_receive_archive_files(int srv, SSL* ssl, int socket, struct stream_buff
...
       {
          goto error;
       }
       remove(file_path);
-      pgmoneta_free_message(msg);
-
-      msg = NULL;
+      memset(msg, 0, sizeof(struct message));
```
jesperpedersen, line 412: pgmoneta_memory_free

src/libpgmoneta/archive.c
```diff
@@ -481,6 +489,7 @@ pgmoneta_receive_archive_stream(int srv, SSL* ssl, int socket, struct stream_buf
    char file_path[MAX_PATH];
    char directory[MAX_PATH];
    char link_path[MAX_PATH];
+   char backup_tar_name[MAX_PATH];
```
jesperpedersen, line 492: `char*`

src/libpgmoneta/archive.c
```diff
@@ -907,6 +920,13 @@ pgmoneta_extract_backup_tar_file(char* file_path, char* destination, struct art*
          pgmoneta_snprintf(dst_path, sizeof(dst_path), "%s/%s", destination, entry_path);
       }
 
+      if (type == AE_IFLNK && pgmoneta_ends_with(dst_path, "/"))
```
jesperpedersen, line 923: pgmoneta_remove_suffix

src/libpgmoneta/archive.c
jesperpedersen, line 1012: pgmoneta_compare_string

src/libpgmoneta/restore.c
```diff
@@ -3194,7 +3194,8 @@ copy_tablespaces_hotstandby(int server, char* from, char* to, char* tblspc_mappi
             memset(copied_tblspc_mappings, 0, strlen(tblspc_mappings) + 1);
             memcpy(copied_tblspc_mappings, tblspc_mappings, strlen(tblspc_mappings));
 
-            token = strtok(copied_tblspc_mappings, ",");
+            char* saveptr = NULL;
```
jesperpedersen, line 3197: Move up

jesperpedersen (changes requested):

jesperpedersen:
@sksingh2005  pgmoneta_workflow_exclude_tablespaces looks weird. Lets try and get the patch as minimal as possible to begin with

jesperpedersen:
@sksingh2005 I'll take this one, so I'll close

PR #1163, opened 2026-05-30
[#1162] Propagate extraction failures to worker outcome deque

### LINKED ISSUE #1162: Improve error reporting in extraction layer
make better use of the worker pool in extraction and related restore paths, including clearer collection of failures from worker threads via the outcome deque..

shashidhar-bm:
- closes #1162

this pr adds an optional workers handle to the extraction and rfile interfaces so extraction failures can be recorded in the worker outcome and reported to the user,

example result: 

```
Outcome:
  Error: 107
  Status: false
  WorkerErrors:
    - File copy failed: /tmp/pgmoneta-test/base/backup/primary/backup/20260605003432/data/global/pg_control
    - File copy failed: /tmp/pgmoneta-test/base/backup/primary/backup/20260605003432/data/pg_wal/archive_status/000000010000000000000007.done
  Workflow: Hot standby
Request:
```

### DIFF
--- a/src/include/workers.h
+++ b/src/include/workers.h
@@ -37,6 +37,7 @@ extern "C" {
 #include <deque.h>
 
 #include <pthread.h>
+#include <stdarg.h>
 #include <stdio.h>
 #include <stdlib.h>
 
@@ -150,12 +151,12 @@ bool
 pgmoneta_workers_outcome_ok(struct workers* workers);
 
 /**
- * Record a failure message on the workers outcome deque (thread-safe).
- * @param workers The workers
- * @param message Message to append (copied)
+ * Record a formatted failure message into a failure deque (thread-safe).
+ * @param failures The failure deque
+ * @param fmt printf-style format string
  */
 void
-pgmoneta_workers_record_failure(struct workers* workers, char* message);
+pgmoneta_record_failure(struct deque* failures, char* fmt, ...) __attribute__((format(printf, 2, 3)));
 
 /**
  * Log all failure messages collected in the outcome deque.
--- a/src/libpgmoneta/rfile.c
+++ b/src/libpgmoneta/rfile.c
@@ -34,6 +34,7 @@
 #include <pgmoneta.h>
 #include <rfile.h>
 #include <utils.h>
+#include <workers.h>
 
 /* system */
 #include <stdint.h>
@@ -78,7 +79,7 @@ error:
 }
 
 int
-pgmoneta_rfile_create(int server, char* label, char* relative_dir, char* base_file_name, int encryption, int compression, struct rfile** rfile)
+pgmoneta_rfile_create(int server, char* label, char* relative_dir, char* base_file_name, int encryption, int compression, struct deque* failures, struct rfile** rfile)
 {
    struct rfile* rf = NULL;
    char* extracted_file_path = NULL;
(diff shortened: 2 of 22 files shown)


shashidhar-bm:
@jesperpedersen 
I think all changes from patch are covered after this PR. The only thing left out is `pgmoneta_deque_add_all`, which i'll add it in a follow-up once there's a concrete use case for it.

src/libpgmoneta/rfile.c
```diff
@@ -176,36 +185,60 @@ pgmoneta_incremental_rfile_initialize(int server, char* label, char* relative_di
    if (nread != sizeof(uint32_t))
    {
       pgmoneta_log_error("rfile initialize: incomplete file header at %s, cannot read magic number", rf->filepath);
+      msg = pgmoneta_format_and_append(msg, "rfile_initialize: corrupt header in %s (label %s) - cannot read magic", rf->filepath, label);
+      pgmoneta_workers_record_failure(workers, msg);
```
Jubilee101: might be an idea to make pgmoneta_workers_record_failure variadic. Would save a lot of lines.

Jubilee101:
Is the context only useful to workers? In single thread situation do we have any error visibility problem? If we do maybe it's worthwhile to decouple the context part, it does feel a bit strange when we pass in workers only to do some bookkeeping instead of actual work.

Jubilee101:
One thing to think about is the streamer, even without the worker it's hard to return the file streaming error in the middle.

shashidhar-bm:
> Is the context only useful to workers? In single thread situation do we have any error visibility problem? If we do maybe it's worthwhile to decouple the context part, it does feel a bit strange when we pass in workers only to do some bookkeeping instead of actual work.

@Jubilee101 , with decouple, did you mean something like..
1) If the function creates / uses the thread pool --> pass `struct workers* workers` (pool + failure collection).
2) If it only records failures (no threading) --> pass `struct deque* failures` (failure collection).

shashidhar-bm:
> One thing to think about is the streamer, even without the worker it's hard to return the file streaming error in the middle.

for this, may be we can add a "local error carrier" on `struct streamer`?

Jubilee101:
> If the function creates / uses the thread pool --> pass struct workers* workers (pool + failure collection).
> If it only records failures (no threading) --> pass struct deque* failures (failure collection).

Yes, if you are only bookkeeping, maybe just pass in what's keeping track. Does this sound alright to you?

Jubilee101:
> for this, may be we can add a "local error carrier" on struct streamer?

This is a little out of the scope for this PR. Let's address it in another issue later. The thing is to me what you are describing here is somewhat very similar to what we have in workers. So we can maybe have a dedicated structure for bookkeeping such issues which can be placed in structs we need.

shashidhar-bm:
> Yes, if you are only bookkeeping, maybe just pass in what's keeping track. Does this sound alright to you?

Makes sense, I'll update the PR..

src/libpgmoneta/workers.c
```diff
@@ -167,6 +167,35 @@ pgmoneta_workers_record_failure(struct workers* workers, char* message)
...
+   if (msg == NULL)
+   {
+      return;
+   }
+
+   va_start(ap, fmt);
+   vsnprintf(msg, size_needed, fmt, ap);
+   va_end(ap);
+
+   pgmoneta_deque_add(failures, NULL, (uintptr_t)msg, ValueString);
```
Jubilee101, line 185: You need to free msg. Remember deque makes a copy.

src/include/workers.h
```diff
@@ -157,6 +158,14 @@ pgmoneta_workers_outcome_ok(struct workers* workers);
 void
 pgmoneta_workers_record_failure(struct workers* workers, char* message);
```
Jubilee101: Is this still used? Should we remove it?

shashidhar-bm: oh yes, it's still being used. since we're introducing the new helper function in this PR, I'll just replace the remaining usages too

Jubilee101:
Rebased and merged, thanks for your contribution!

Issue #1166, opened 2026-05-31
Add non-recursive S3 backup prefix listing


Amr-Shams:
pgmoneta currently lists S3 objects recursively under a backup prefix.
scheduled S3 verification needs to list known backup labels under the server backup root, for example:`<s3_base_dir>/<server>/backup/`
add a non-recursive S3 listing path using `delimiter=/` with ListObjectsV2 and parse `CommonPrefixes` from the response. this should return backup label prefixes without walking every object in every backup. 

The verification path can use this to decide which locally known backups have S3 as their authoritative data residence.

PR #1167, opened 2026-05-31
[#1166]feat(s3): impelemnt bucket listing for later intergartion

Amr-Shams:


src/include/storage.h
```diff
@@ -64,6 +64,16 @@ pgmoneta_storage_create_ssh(int workflow_type);
...
 
+/**
+ * List backup labels available in the configured storage engine.
+ *
+ * @param server The server index
+ * @param labels The resulting deque of backup labels
+ * @return 0 on success, otherwise 1
+*/
+int
+pgmoneta_storage_list_backup_labels(int server, struct deque** labels);
```
Amr-Shams, line 75: @Jubilee101 what u think exposing this for all storage sounds the right direction

Jubilee101: Eventually there will be a hybrid of storage options. Users can choose to store backups in S3, locally, in another machine or all of them. For us we should always keep a metadata file for each backup locally, as a virtual backup. The real location doesn't matter when it comes to listing all the backups, we'll just go through local metadata files.

That said, this should be useful to keep an access to the source of truth, after all the storage engines themselves know better what backups they have. CC @jesperpedersen

Amr-Shams: i agree. loal metadata should stay the catalog for normal `list-backup`.
the reason i wanted this expos at the storage level is for verify/repair paths, not for replacing metadata listing. i.e, with S3 the daemon can see a backup locally through `backup.info`, while the materialized files may exist only remotely. Then the current path walks the local backup directory and fails. #1070 has more context.

yhe storage function answers a different question: which backup labels does this backend physically contain? S3, that means listing first-level prefixes under the backup root. Verify can then match local metadata against the S3 inventory and choose the S3 verification path for those backups.

it also gives us an early signal that the backend is reachable and the credentials work.  in a hybrid setup, though, a label should probably record which storage backend owns the materialized data. Otherwise verify has to infer ownership by probing each enabled backend.

PR #1168, opened 2026-06-01
test: add ability to start and verify a restored backup

### LINKED ISSUE #1154: Test Infra: Ability to start a restored backup
We can't ensure 100% data integrity unless we are able to start the backup we restored and actually query the rows we backed up. For that we need to be able to start the restored backup, ideally in a container of course. Maybe it's an idea to mount the directory (or directories, if there are tablespaces) to container directly.

justin212407:
Closes #1154

## Summary

Adds test infrastructure to start a restored PostgreSQL backup inside a Podman container and verify data integrity by querying the restored server directly.

Previously, the test suite could take and restore a backup but had no way to confirm the restored data directory was actually startable or that the backed-up rows were intact. This PR closes that gap.

## Changes

**test/libpgmonetatest/tscommon.c**
- Added `start_restored_backup(restore_dir, port)`: ensures all required PostgreSQL subdirectories exist, writes a trust-auth
  `pg_hba.conf`, handles `recovery.signal`/`standby.signal` via `PGMONETA_TEST_RECOVERY_SIGNAL`, starts a Podman container with the restored directory mounted as PGDATA, and waits up to 30 seconds for `pg_isready` to succeed before returning the port.
- Added `stop_restored_backup()`: removes the container, tolerating errors if it does not exist.

**test/include/tscommon.h**
- Declared `start_restored_backup()` and `stop_restored_backup()`.
- Added `RESTORED_BACKUP_DEFAULT_PORT` (15432) so test cases do not hardcode the port.

**test/testcases/test_restored_backup.c**
- New end-to-end test `test_pgmoneta_restored_backup_start` that:
  1. Seeds a table with 3 rows on the primary.
  2. Takes a full backup.
  3. Restores it to `TEST_RESTORE_DIR`.
  4. Starts the restored directory in a container via `start_restored_backup()`.
  5. Runs `psql` inside the container to verify the 3 rows are present.
  6. Cleans up the container and restore directory unconditionally.

## Notes

- Uses rootless Podman with `:Z` volume flag for SELinux compatibility.
- Container image is resolved from `TEST_PG_VERSION` env var, falling back to `17`, matching the pattern used by the rest of the test suite.
- Tablespace support is out of scope for this PR. A `TODO` comment is left in `start_restored_backup()` noting that tablespace directories would require additional volume mounts.
- Setting `PGMONETA_TEST_RECOVERY_SIGNAL=1` before running will create `recovery.signal` in the restored directory if needed.

### DIFF
--- a/test/include/tscommon.h
+++ b/test/include/tscommon.h
@@ -39,8 +39,9 @@ extern "C" {
 #include <stdbool.h>
 #include <openssl/ssl.h>
 
-#define PRIMARY_SERVER   0
-#define ENV_VAR_BASE_DIR "PGMONETA_TEST_BASE_DIR"
+#define PRIMARY_SERVER               0
+#define ENV_VAR_BASE_DIR             "PGMONETA_TEST_BASE_DIR"
+#define RESTORED_BACKUP_DEFAULT_PORT 15432
 
 extern char TEST_CONFIG_SAMPLE_PATH[MAX_PATH];
 extern char TEST_RESTORE_DIR[MAX_PATH];
@@ -172,6 +173,24 @@ pgmoneta_test_resolve_binary_path(const char* binary_name, char* out);
  */
 int
 pgmoneta_test_exec_command(const char* command, char** output, int* exit_code);
+
+/**
+ * Start a PostgreSQL server from a restored backup directory.
+ * Ensures required subdirectories exist, keeps the original pg_hba.conf
+ * from the backup, and waits for pg_isready to succeed.
+ * Detects whether to use a container or local pg_ctl based on environment.
+ * @param restore_dir Path to restored PGDATA directory
+ * @param port Host port to listen on (<=0 uses RESTORED_BACKUP_DEFAULT_PORT)
+ * @return The port on success, otherwise -1
+ */
+int
+pgmoneta_start_restored_backup(const char* restore_dir, int port);
+
+/**
+ * Stop the restored backup server, ignoring errors.
+ */
+void
+pgmoneta_stop_restored_backup(void);
 
 /**
  * Load a configuration file into shared memory.
--- a/test/libpgmonetatest/tscommon.c
+++ b/test/libpgmonetatest/tscommon.c
@@ -48,13 +48,15 @@
 #include <sys/wait.h>
 #include <unistd.h>
 
-#define ENV_VAR_CONF_PATH        "PGMONETA_TEST_CONF"
-#define ENV_VAR_CONF_SAMPLE_PATH "PGMONETA_TEST_CONF_SAMPLE"
-#define ENV_VAR_USER_CONF        "PGMONETA_TEST_USER_CONF"
-#define ENV_VAR_RESTORE_DIR      "PGMONETA_TEST_RESTORE_DIR"
-#define ENV_VAR_RETROSPECT_DIR   "PGMONETA_TEST_RETROSPECT_DIR"
-#define ENV_VAR_HOT_STANDBY_DIR  "PGMONETA_TEST_HOT_STANDBY_DIR"
-#define ENV_VAR_EXECUTABLE_DIR   "PGMONETA_TEST_EXECUTABLE_DIR"
+#define ENV_VAR_CONF_PATH            "PGMONETA_TEST_CONF"
+#define ENV_VAR_CONF_SAMPLE_PATH     "PGMONETA_TEST_CONF_SAMPLE"
+#define ENV_VAR_USER_CONF            "PGMONETA_TEST_USER_CONF"
+#define ENV_VAR_RESTORE_DIR          "PGMONETA_TEST_RESTORE_DIR"
+#define ENV_VAR_RETROSPECT_DIR       "PGMONETA_TEST_RETROSPECT_DIR"
+#define ENV_VAR_HOT_STANDBY_DIR      "PGMONETA_TEST_HOT_STANDBY_DIR"
+#define ENV_VAR_EXECUTABLE_DIR       "PGMONETA_TEST_EXECUTABLE_DIR"
+
+#define RESTORED_BACKUP_WAIT_SECONDS 30
 
 char TEST_CONFIG_SAMPLE_PATH[MAX_PATH];
 char TEST_RESTORE_DIR[MAX_PATH];
@@ -87,7 +89,6 @@ pgmoneta_test_environment_create(void)
 
    conf_path = getenv(ENV_VAR_CONF_PATH);
    assert(conf_path != NULL);
-   // Create the shared memory for the configuration
    size = sizeof(struct main_configuration);
    ret = pgmoneta_create_shared_memory(size, HUGEPAGE_OFF, &shmem);
    assert(!ret);
@@ -95,17 +96,14 @@ pgmoneta_test_environment_create(void)
    ret = pgmoneta_init_main_configuration(shmem);
    assert(!ret);
 
-   // Try reading configuration from the configuration path
    ret = pgmoneta_read_main_configuration(shmem, conf_path);
    assert(!ret);
 
-   // Validate the configuration is valid
    ret = pgmoneta_test_validate_configuration(shmem);
    assert(!ret);
 
    config = (struct main_configuration*)shmem;
 
-   // some validations just to be safe
    memcpy(&config->common.configuration_path[0], conf_path, MIN(strlen(conf_path), MAX_PATH - 1));
    assert(config->common.number_of_servers > 0);
    assert(pgmoneta_compare_string(config->common.servers[0].name, "primary"));
@@ -135,7 +133,6 @@ pgmoneta_test_environment_create(void)
 
    pgmoneta_start_logging();
 
-   // Try reading the users configuration path
    ret = pgmoneta_read_users_configuration(shmem, user_conf_path);
    assert(!ret);
 }
@@ -166,14 +163,12 @@ pgmoneta_test_validate_configuration(void* shmem)
       return 1;
    }
 
-   // Use the existing validation function for comprehensive checks
    return pgmoneta_validate_main_configuration(shmem);
 }
 
 int
 pgmoneta_test_add_backup(void)
 {
-   /* Ensure server is online before backup */
    if (pgmoneta_tsclient_mode("primary", "online", 0))
    {
       return 1;
@@ -189,7 +184,6 @@ pgmoneta_test_add_backup(void)
 int
 pgmoneta_test_add_backup_chain(void)
 {
-   /* Ensure server is online before backup */
    if (pgmoneta_tsclient_mode("primary", "online", 0))
    {
       return 1;
@@ -220,7 +214,6 @@ pgmoneta_test_basedir_cleanup(void)
    char* wal_dir = NULL;
    bool restart = false;
    struct main_configuration* config;
-   int ret = 0;
 
    config = (struct main_configuration*)shmem;
 
@@ -321,7 +314,6 @@ pgmoneta_test_config_restore(void)
 int
 pgmoneta_test_backup(const char* server_name, const char* backup_name)
 {
-   /* Cast const away as pgmoneta_tsclient_backup doesn't modify the strings */
    return pgmoneta_tsclient_backup((char*)server_name, (char*)backup_name, 0);
 }
 
@@ -331,7 +323,6 @@ pgmoneta_test_execute_query(int srv, SSL* ssl, int skt, char* query, struct quer
    struct message* msg = NULL;
    struct query_response* response = NULL;
 
-   /* create and execute the query */
    pgmoneta_create_query_message(query, &msg);
    if (pgmoneta_query_execute(ssl, skt, msg, &response) || response == NULL)
    {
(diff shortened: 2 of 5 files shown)


jesperpedersen:
@shashidhar-bm Can you start with this ?

shashidhar-bm:
@jesperpedersen , sure, on it.

shashidhar-bm:
Hi @justin212407 , please squash this into one commit and also check with https://github.com/pgmoneta/pgmoneta/blob/main/doc/DEVELOPERS.md#authors

test/include/tscommon.h
```diff
@@ -188,18 +174,15 @@ pgmoneta_test_load_conf(char* conf_path);
...
-   char test_home[MAX_PATH];            /**< The test home directory path */
-   char* original_home;                 /**< The original home environment variable value */
-   char original_config_home[MAX_PATH]; /**< The original configuration home directory path */
-   int original_encryption;             /**< The original encryption setting */
-   bool shmem_locally_allocated;        /**< Flag indicating if shared memory was allocated locally */
+   char test_home[MAX_PATH];
+   char* original_home;
+   char original_config_home[MAX_PATH];
+   int original_encryption;
+   bool shmem_locally_allocated;
```
shashidhar-bm: I think these changes aren’t related to this PR?

test/libpgmonetatest/tscommon.c
```diff
@@ -614,3 +572,211 @@ pgmoneta_test_teardown_encryption_env(struct test_encryption_env* env)
...
+   version_env = getenv("TEST_PG_VERSION");
+   version = (version_env != NULL && version_env[0] != '\0') ? version_env : "17";
+
+   if (pgmoneta_snprintf(image, sizeof(image), "pgmoneta-test-postgresql%s-rocky10", version) <= 0)
+   {
+      return -1;
+   }
+
+   if (pgmoneta_snprintf(command, sizeof(command),
+                         "podman run -d --name %s -p %d:5432 -v \"%s:/pgdata:Z\" "
```
shashidhar-bm: use `$CONTAINER_ENGINE` from check.sh

test/testcases/test_restored_backup.c
```diff
@@ -0,0 +1,157 @@
...
+
+   /* --- backup and restore --- */
+   MCTF_ASSERT(pgmoneta_test_add_backup() == 0, cleanup,
+               "backup failed - check server is online and backup configuration");
+
+   MCTF_ASSERT(pgmoneta_tsclient_restore("primary", "newest", "current", 0) == 0, cleanup,
+               "restore operation failed");
+
+   /* --- start restored server in container --- */
+   restored_port = start_restored_backup(TEST_RESTORE_DIR, RESTORED_BACKUP_DEFAULT_PORT);
```
shashidhar-bm: it should be `TEST_RESTORE_DIR/primary-<label>`, and this might also be the reason for the CI failure.

You can find some references in `test_cli.c` on how to get the restored `PGDATA`.

test/include/tscommon.h
justin212407: This is an unintentional change caused by clang-format. I have restored the comments.

test/libpgmonetatest/tscommon.c
justin212407: I have made the change you suggested. `start_restored_backup` and `stop_restored_backup` in `tscommon.c` now read `CONTAINER_ENGINE` from the environment and fall back to podman if unset, matching the pattern in `check.sh`. The same change has been applied to the psql exec command in `test_restored_backup.c`.

test/testcases/test_restored_backup.c
justin212407: The test now calls `pgmoneta_tsclient_list_backup` after the restore to retrieve the backup label, then constructs the correct path as `TEST_RESTORE_DIR/primary-<label>` before passing it to `start_restored_backup`, following the same pattern used in `test_cli.c`.

shashidhar-bm:
@justin212407 ,there are some build errors, pleasee make sure `./test/check.sh` works fine locally, check [doc/TEST.md](https://github.com/pgmoneta/pgmoneta/blob/main/doc/TEST.md#:~:text=Single%20test%20or,or%20is%20aborted.)

justin212407:
> @justin212407 ,there are some build errors, pleasee make sure `./test/check.sh` works fine locally, check [doc/TEST.md](https://github.com/pgmoneta/pgmoneta/blob/main/doc/TEST.md#:~:text=Single%20test%20or,or%20is%20aborted.)

on it.

justin212407:
@shashidhar-bm I have fixed the failing build can you take a look now.

test/include/tscommon.h
```diff
@@ -102,17 +103,12 @@ pgmoneta_test_teardown(void);
 
 /**
  * Snapshot the current shared-memory configuration.
- * Call this at the start of a test (or in MCTF_TEST_SETUP) to preserve
- * the original config before a test modifies it.
```
Jubilee101, line 106: What's wrong with this and below comments that you are removing?

test/libpgmonetatest/tscommon.c
```diff
@@ -614,3 +604,483 @@ pgmoneta_test_teardown_encryption_env(struct test_encryption_env* env)
...
+   if (pgmoneta_snprintf(path, sizeof(path), "%s/%s", restore_dir, subdir) <= 0)
+   {
+      return 1;
+   }
+
+   return pgmoneta_mkdir(path);
+}
+
+static int
+write_restore_pg_hba(const char* restore_dir)
```
Jubilee101: I don't get this, if we are restoring from a backup we made only a few seconds ago, we should already have a working conf, including hba. Can you explain why we need it?

The same question for write_restore_postgresql_conf

Jubilee101:
Also this won't work in CI, as Github Action has very poor support for containers.. So we'll need an option to start the backup locally. We, including @shashidhar-bm have learned it the hard way I believe..

Thanks for working on this @justin212407

test/include/tscommon.h
justin212407: Those doc comments were unintentionally removed during a refactor, I have restored them now. Thanks for pointing that out.

test/libpgmonetatest/tscommon.c
justin212407: The restored backup does include copies of both files from the original server, but they're not directly usable for this test:

`write_restore_pg_hba`: the original `pg_hba.conf` requires the primary's auth method (SCRAM/password), so we overwrite it with trust auth so the test can connect to the restored server without credentials.
`write_restore_postgresql_conf`: rather than overwriting, we now append overrides. The original file has the primary's port and a restricted listen_addresses, which would conflict with the already-running primary server. We append listen_addresses = '*' and a non-conflicting port at the end of the file, since PostgreSQL applies the last value when a setting is repeated this preserves the rest of the original configuration.

I've added these explanations as code comments above each function as well.

justin212407:
> Also this won't work in CI, as Github Action has very poor support for containers.. So we'll need an option to start the backup locally. We, including @shashidhar-bm have learned it the hard way I believe..
> 
> Thanks for working on this @justin212407

I really appreciate you flagging this before it caused issues in CI. I've reworked the implementation to start the restored backup locally using `pg_ctl` start instead of a container, so it no longer depends on container support in GitHub Actions. Also wanted to mention apologies for the gap in activity on this, I was under the weather for a couple of weeks. Back on it now and the updated approach is passing locally, both the isolated test and the full suite. Let me know if this looks right.

test/include/tscommon.h
```diff
@@ -173,6 +174,23 @@ pgmoneta_test_resolve_binary_path(const char* binary_name, char* out);
...
+/**
+ * Start a PostgreSQL server from a restored backup directory.
+ * Ensures required subdirectories exist, updates pg_hba.conf for trust auth,
+ * and waits for pg_isready to succeed.
+ * @param restore_dir Path to restored PGDATA directory
+ * @param port Host port to listen on (<=0 uses RESTORED_BACKUP_DEFAULT_PORT)
+ * @return The port on success, otherwise -1
+ */
+int
+start_restored_backup(const char* restore_dir, int port);
```
Jubilee101: Prefix with `pgmoenta_` as this is public API now.

test/include/tscommon.h
```diff
@@ -173,6 +174,23 @@ pgmoneta_test_resolve_binary_path(const char* binary_name, char* out);
...
+
+/**
+ * Stop the restored backup server, ignoring errors.
+ */
+void
+stop_restored_backup(void);
```
Jubilee101: Same comment as above.

test/libpgmonetatest/tscommon.c
Jubilee101: I see. I agree with the postgres conf change. But for hba conf, the password auth had better be part of the test. There's no use of a backup if user can't log in anymore upon restore, so we need to cover that. Unless you are facing some architectural issue that you can't easily impletment that?

test/libpgmonetatest/tscommon.c
```diff
@@ -614,3 +602,318 @@ pgmoneta_test_teardown_encryption_env(struct test_encryption_env* env)
...
+      pgmoneta_snprintf(pg_ctl, sizeof(pg_ctl), "pg_ctl");
+   }
+
+   pgmoneta_snprintf(pg_isready, sizeof(pg_isready), "/usr/pgsql-%s/bin/pg_isready", version);
+   if (access(pg_isready, X_OK) != 0)
+   {
+      pgmoneta_snprintf(pg_isready, sizeof(pg_isready), "pg_isready");
+   }
+
+   if (pgmoneta_snprintf(command, sizeof(command),
```
Jubilee101: You have to consider both situations, when running locally we still wish to run within docker. It's just in CI that you can directly run locally.

Jubilee101 (changes requested):

Jubilee101:
> I was under the weather for a couple of weeks.

No worries, thanks for letting us know, and take care~

test/libpgmonetatest/tscommon.c
Jubilee101: Also I triggered CI for you, and seems like something is still not right. You can temporarily modify https://github.com/pgmoneta/pgmoneta/blob/main/.github/workflows/ci.yml to have CI enabled in your forked repo. Then you can debug CI directly.

test/include/tscommon.h
justin212407: Renamed to `pgmoneta_start_restored_backup`.

test/include/tscommon.h
justin212407: Renamed to `pgmoneta_start_restored_backup`.

test/libpgmonetatest/tscommon.c
justin212407: I have removed `write_restore_pg_hba` entirely, the original `pg_hba.conf` from the restored backup is now kept as-is. The `psql` verification command now uses `PGPASSWORD=mypass` with the actual test user credentials, so password authentication is properly exercised.

justin212407: > You have to consider both situations, when running locally we still wish to run within docker. It's just in CI that you can directly run locally.

I have added `is_local_restore()` which checks for ``CI=true (set automatically by GitHub Actions) or `PGMONETA_TEST_LOCAL_RESTORE=1`. In CI or when that flag is set, the server starts locally via `pg_ctl`. Otherwise it uses the pre-built `pgmoneta-test-postgresql<version>-rocky10` container image with the existing podman `--uidmap +26:0:1 --gidmap +26:0:1` approach for local development. Both paths use `PGPASSWORD=mypass` for psql verification.

justin212407: > Also I triggered CI for you, and seems like something is still not right. You can temporarily modify https://github.com/pgmoneta/pgmoneta/blob/main/.github/workflows/ci.yml to have CI enabled in your forked repo. Then you can debug CI directly.

Will certainly do that and add the push trigger on my branch.

jesperpedersen:
@justin212407 @shashidhar-bm Latest update on this ?

test/libpgmonetatest/tscommon.c
```diff
@@ -614,3 +602,425 @@ pgmoneta_test_teardown_encryption_env(struct test_encryption_env* env)
...
+   {
+      return 1;
+   }
+
+   if (pgmoneta_snprintf(path, sizeof(path), "%s/%s", restore_dir, subdir) <= 0)
+   {
+      return 1;
+   }
+
+   return pgmoneta_mkdir(path);
```
shashidhar-bm, line 621: With this change, we are automatically creating those folders if they don’t already exist. However, the test should be able to fail when the folders are missing.

In the current behavior, if the folders are missing, the user has to create them manually before starting PostgreSQL with the restored data, which I believe shouldn’t be the case.

shashidhar-bm (commented):
Hi @justin212407 , LGTM! Just one small change, please refer to #1261. This particular case isn’t covered in the current tests.

And for the storage engines, we can introduce this test case later along with #1261.

test/libpgmonetatest/tscommon.c
jesperpedersen: All of this should be transparent to the user - so no create/delete for the user

PR #1185, opened 2026-06-16
[#1184]  Retain local backup when local is combined with a remote engine

### LINKED ISSUE #1184: Local backup isn't retained when local is configured alongside another storage engine
For example, with `storage_engine = local, s3`, the backup should be retained both locally and in S3. Currently, however, the local backup data is deleted after it is uploaded to S3.

shashidhar-bm:
- fixes #1184

### DIFF
--- a/src/libpgmoneta/se_s3.c
+++ b/src/libpgmoneta/se_s3.c
@@ -39,6 +39,7 @@
 #include <manifest.h>
 #include <progress.h>
 #include <security.h>
+#include <storage.h>
 #include <utils.h>
 #include <value.h>
 #include <vfile.h>
@@ -549,11 +550,16 @@ s3_storage_teardown(char* name __attribute__((unused)), struct art* nodes)
 
    pgmoneta_log_debug("S3 storage engine (teardown): %s/%s", config->common.servers[server].name, label);
 
-   root = pgmoneta_get_server_backup_identifier_data(server, label);
+   if (!pgmoneta_is_storage_engine_enabled(STORAGE_ENGINE_LOCAL))
+   {
+      root = pgmoneta_get_server_backup_identifier_data(server, label);
 
-   pgmoneta_delete_directory(root);
-
-   free(root);
+      if (root != NULL)
+      {
+         pgmoneta_delete_directory(root);
+         free(root);
+      }
+   }
 
    return 0;
 }
--- a/src/include/pgmoneta.h
+++ b/src/include/pgmoneta.h
@@ -141,6 +141,7 @@ extern "C" {
 #define COMPRESSION_IS_CLIENT(t)     (!!((t) & COMPRESSION_TYPE_CLIENT))
 #define COMPRESSION_ALGORITHM(t)     ((t) & 0x0F)
 
+#define STORAGE_ENGINE_NONE          0
 #define STORAGE_ENGINE_LOCAL         1 << 0
 #define STORAGE_ENGINE_SSH           1 << 1
 #define STORAGE_ENGINE_S3            1 << 2
--- a/src/include/storage.h
+++ b/src/include/storage.h
@@ -117,6 +117,15 @@ pgmoneta_sftp_wal_open(int server, char* filename, int segsize, sftp_file* file)
  */
 int
 pgmoneta_sftp_wal_close(int server, char* filename, bool partial, sftp_file* file);
+
+/**
+ * Check if a storage engine is enabled in the current configuration
+ * @param engine The storage engine flag (e.g. STORAGE_ENGINE_LOCAL, STORAGE_ENGINE_S3)
+ * @return true if enabled, otherwise false
+ */
+bool
+pgmoneta_is_storage_engine_enabled(int engine);
+
 #ifdef __cplusplus
 }
 #endif
--- a/src/libpgmoneta/configuration.c
+++ b/src/libpgmoneta/configuration.c
@@ -5463,7 +5463,7 @@ error:
 static int
 as_storage_engine(char* str)
 {
-   int STORAGE_ENGINE_TYPES = STORAGE_ENGINE_LOCAL;
+   int STORAGE_ENGINE_TYPES = STORAGE_ENGINE_NONE;
    char* token = NULL;
    char* delimiter = ",";
    int i = 0, j = 0;
--- a/src/libpgmoneta/se_azure.c
+++ b/src/libpgmoneta/se_azure.c
@@ -31,6 +31,7 @@
 #include <http.h>
 #include <logging.h>
 #include <security.h>
+#include <storage.h>
 #include <utils.h>
 #include <workflow.h>
 
@@ -194,13 +195,18 @@ azure_storage_teardown(char* name __attribute__((unused)), struct art* nodes)
    server = (int)pgmoneta_art_search(nodes, NODE_SERVER_ID);
    label = (char*)pgmoneta_art_search(nodes, NODE_LABEL);
 
-   root = pgmoneta_get_server_backup_identifier_data(server, label);
+   if (!pgmoneta_is_storage_engine_enabled(STORAGE_ENGINE_LOCAL))
+   {
+      root = pgmoneta_get_server_backup_identifier_data(server, label);
 
-   pgmoneta_delete_directory(root);
+      if (root != NULL)
+      {
+         pgmoneta_delete_directory(root);
+         free(root);
+      }
+   }
 
    pgmoneta_log_debug("Azure storage engine (teardown): %s/%s", config->common.servers[server].name, label);
-
-   free(root);
 
    return 0;
 }
--- a/src/libpgmoneta/storage.c
+++ b/src/libpgmoneta/storage.c
@@ -1,0 +1,39 @@
+/*
+ * Copyright (C) 2026 The pgmoneta community
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
+ * are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this list
+ * of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice, this
+ * list of conditions and the following disclaimer in the documentation and/or other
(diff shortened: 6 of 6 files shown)


src/libpgmoneta/se_s3.c
```diff
@@ -549,11 +549,15 @@ s3_storage_teardown(char* name __attribute__((unused)), struct art* nodes)
 
    pgmoneta_log_debug("S3 storage engine (teardown): %s/%s", config->common.servers[server].name, label);
 
-   root = pgmoneta_get_server_backup_identifier_data(server, label);
+   /* Keep the local data when the local engine is also configured */
+   if (!(config->storage_engine & STORAGE_ENGINE_LOCAL))
```
jesperpedersen: Shouldn't this be STORAGE_ENGINE_S3 w/o the !

jesperpedersen: Hmm, right

jesperpedersen: I think we need a function to deal with the different combinations - maybe `pgmoneta_is_storage_engine_enabled`

jesperpedersen:
## Overall

- **Readiness**: The change set is not ready for merge due to critical logic errors and resource management issues across three files, resulting in Rejected verdicts for configuration, Azure, and S3 storage engines.
- **Risk**: High risk of memory leaks and resource exhaustion, as teardown functions for Azure and S3 storage engines now incorrectly skip cleanup based on the state of the local storage engine.
- **Risk**: Significant risk of null pointer dereferences and undefined behavior in `se_azure.c` and `utils.c` due to missing NULL checks on pointers returned by helper functions before use or free.
- **Theme**: A consistent semantic error exists where storage engine teardowns are coupled to the enabled status of the local storage engine, violating separation of concerns and likely breaking cleanup logic.
- **Theme**: Code quality issues include the use of magic numbers instead of symbolic constants in `configuration.c` and missing input validation or documentation in `utils.c`.

## Code

- **src/libpgmoneta/configuration.c:5466**: Magic number 0 used instead of the symbolic constant STORAGE_ENGINE_LOCAL, reducing code readability and maintainability.
- **src/libpgmoneta/se_azure.c:197-203**: Logic error: `pgmoneta_get_server_backup_identifier_data` is called but `root` is never freed if the function returns NULL or fails, leading to a potential memory leak or null pointer dereference on subsequent lines if not handled correctly (though `pgmoneta_delete_directory` likely handles NULL, the free is skipped). More critically, the logic is inverted or incomplete: if the local storage engine is *disabled*, it proceeds to delete the directory; if it is *enabled*, it does nothing. This seems semantically wrong for a teardown function (usually you want to clean up regardless, or perhaps only if the engine is enabled?). Assuming the intent was to skip cleanup for local storage, the condition `!pgmoneta_is_storage_engine_enabled` skips the cleanup when local is *disabled*. If the intent was to skip cleanup when local is *enabled*, the condition should be `pgmoneta_is_storage_engine_enabled`.
- **src/libpgmoneta/se_azure.c:197**: Potential Null Pointer Dereference: `pgmoneta_get_server_backup_identifier_data` might return NULL. If it does, `pgmoneta_delete_directory(root)` is called with NULL. While this might be safe depending on the implementation of `pgmoneta_delete_directory`, it is risky without explicit NULL checking.
- **src/libpgmoneta/se_azure.c:197-203**: Inconsistent indentation: The block inside the `if` statement is indented with 6 spaces (3 levels of 2 spaces), but the `if` statement itself is at 3 spaces. Standard C style usually aligns with 2 or 4 spaces per level. The existing code uses 2 spaces. The new block uses 2 spaces for the `if`, then 4 spaces for the body, then 6 spaces for the nested statements. This is consistent with the previous lines (e.g., line 194 uses 6 spaces for `server = ...`), so the indentation is actually correct relative to the surrounding context. However, the logic is the primary issue.
- **src/libpgmoneta/se_azure.c:197-203**: Logic Ambiguity: The change introduces a conditional execution path for teardown. If the local storage engine is enabled, the Azure storage engine teardown does *nothing* (no directory deletion). This implies that local storage handles its own cleanup or that Azure storage is not responsible for cleaning up when local is the primary engine. This requires careful verification of the system's
- **src/libpgmoneta/se_s3.c**: VERDCT: REJECT
- **src/libpgmoneta/se_s3.c:552-557**: The logic change prevents S3 storage teardown from cleaning up directories if the local storage engine is disabled, likely causing resource leaks or stale data.
- **src/libpgmoneta/se_s3.c:552**: The condition `!pgmoneta_is_storage_engine_enabled(STORAGE_ENGINE_LOCAL)` appears semantically incorrect for an S3 storage teardown function, as it couples S3 cleanup to the state of a different storage engine.
- **src/libpgmoneta/se_s3.c:552**: The new conditional block creates a dependency on the local storage engine's status for S3 operations, which violates the separation of concerns between storage backends.
- **src/libpgmoneta/se_s3.c:556**: `free(root)` is now conditional; if the condition is false, the pointer returned by `pgmoneta_get_server_backup_identifier_data` is leaked.
- **src/libpgmoneta/utils.c:5936**: Missing header inclusion for `bool` type (likely `<stdbool.h>`) if not already included elsewhere in the file.
- **src/libpgmoneta/utils.c:5936**: No input validation for the `engine` parameter, though bitwise AND is safe, documentation should clarify expected values.

## Security

- **src/libpgmoneta/se_azure.c:197-204**: Logic Error - `pgmoneta_delete_directory` is called without checking if `root` is non-NULL; if `pgmoneta_get_server_backup_identifier_data` returns NULL, this leads to undefined behavior or a crash.
- **src/libpgmoneta/se_azure.c:197-204**: Logic Error - The code now skips cleanup entirely if `STORAGE_ENGINE_LOCAL` is enabled. If Azure storage requires teardown regardless of local engine status, this results in resource leaks.
- **src/libpgmoneta/se_azure.c:202**: Logic Error - `free(root)` is executed even if `pgmoneta_delete_directory` fails or if `root` was NULL (depending on implementation), potentially causing double-free or use-after-free if the delete function does not nullify the pointer.
- **src/libpgmoneta/se_azure.c:197**: Logic Error - The condition `!pgmoneta_is_storage_engine_enabled(STORAGE_ENGINE_LOCAL)` implies that if the local engine is enabled, teardown is skipped. This is likely incorrect for Azure storage teardown, which should happen regardless of other engines.

## Memory

- **src/libpgmoneta/se_azure.c:197-204**: Memory leak introduced; `pgmoneta_get_server_backup_identifier_data` allocates memory that is now only freed when the local storage engine is disabled, causing a leak when the local engine is enabled.
- **src/libpgmoneta/utils.c:5935**: Potential undefined behavior: If `engine` is 0, the bitwise AND result is 0 (false), which is likely correct, but if `engine` is not a valid bit mask, the logic may silently fail or behave unexpectedly.
- **src/libpgmoneta/utils.c:5935**: Implicit assumption: The code assumes `config->storage_engine` is a bitmask compatible with `engine`, but no validation is performed on the `engine` argument or the config structure's state.

## Performance

- **src/libpgmoneta/se_azure.c:197-202**: Logic error causing Azure teardown to be skipped when Local storage is disabled, likely preventing cleanup of Azure resources.
- **src/libpgmoneta/se_azure.c:201**: `free(root)` is now conditional; if the check fails, the memory allocated by `pgmoneta_get_server_backup_identifier_data` is leaked.
- **src/libpgmoneta/se_azure.c:200**: `pgmoneta_delete_directory(root)` is skipped when Local storage is disabled, leaving orphaned directories if the logic is intended to be unconditional for Azure.

## Test Suite

- **src/libpgmoneta/se_azure.c:197-203**: Logic error: `pgmoneta_delete_directory` and `free` are now skipped when `STORAGE_ENGINE_LOCAL` is enabled, causing resource leaks and incomplete teardown.
- **src/libpgmoneta/utils.c:5931-5935**: Missing test coverage for the new function `pgmoneta_is_storage_engine_enabled`.

## Documentation

- **src/libpgmoneta/configuration.c:5466**: Magic number `0` used instead of `STORAGE_ENGINE_LOCAL` (or equivalent enum value) reduces code readability and maintainability; the original code explicitly named the default value.
- **src/libpgmoneta/se_s3.c:552-557**: Missing comment explaining why local storage engine teardown is skipped or why this conditional logic is necessary for S3 teardown.
- **src/libpgmoneta/utils.c:5931-5935**: Missing function documentation (Doxygen comments) describing parameters, return value, and purpose.

## Conclusion

**orangu rejects this patch**

- Rejected: **src/libpgmoneta/configuration.c**
- Rejected: **src/libpgmoneta/se_azure.c**
- Rejected: **src/libpgmoneta/se_s3.c**
- Rejected: **src/libpgmoneta/utils.c**

Generated by: **orangu 0.8.0** (bartowski/Qwen_Qwen3.6-35B-A3B-GGUF)

jesperpedersen:
@shashidhar-bm See if you can use something of this - otherwise PTAL me

shashidhar-bm:
> * reducing code readability and maintainability.

@jesperpedersen , i've considered couple of  points, thank you

src/include/utils.h
```diff
@@ -1562,6 +1562,14 @@ pgmoneta_time_is_valid(pgmoneta_time_t t);
 int
 pgmoneta_time_format(pgmoneta_time_t t, enum pgmoneta_time_format_t fmt, char** output);
 
+/**
+ * Check if a storage engine is enabled in the current configuration
+ * @param engine The storage engine flag (e.g. STORAGE_ENGINE_LOCAL, STORAGE_ENGINE_S3)
+ * @return true if enabled, otherwise false
+ */
+bool
+pgmoneta_is_storage_engine_enabled(int engine);
```
jesperpedersen: We have storage.h

src/libpgmoneta/utils.c
```diff
@@ -5929,3 +5929,10 @@ pgmoneta_get_lsn(uint32_t hi, uint32_t lo)
 {
    return ((uint64_t)hi << 32) | lo;
 }
+
+bool
```
jesperpedersen: Stick this in storage.c

jesperpedersen (changes requested):

jesperpedersen:
Merged.

PR #1188, opened 2026-06-17
[#1175] Implement health check

### LINKED ISSUE #1175: Health Check
cross port from pgagroal

trxvorr:
Port the periodic health check worker from pgagroal: probes each server with SELECT 1, tracks per-server health and auth type via the pgmoneta_server_health metric, with the health_check* configuration parameters.

Closes #1175


src/main.c
```diff
@@ -904,6 +906,12 @@ main(int argc, char** argv)
    ev_periodic_init(&verification, verification_cb, 0., pgmoneta_time_convert(config->verification, FORMAT_TIME_S), 0);
    ev_periodic_start(main_loop, &verification);
 
+   /* Start the health check worker */
+   if (config->health_check)
```
jesperpedersen, line 910: @trxvorr Why is this not a periodic ?
@Userfrom1995 Cc: Feedback ?

jesperpedersen:
@trxvorr I think we will wait with this until @Userfrom1995 is done with the new I/O layer

src/main.c
trxvorr: oops I ported pgagroal's forked worker model directly. I'll switch it to an ev_periodic after the new i/O layer

jesperpedersen:
@trxvorr Coordinate with @Userfrom1995

PR #1190, opened 2026-06-20
Harden SCRAM iteration count

trxvorr:
The SCRAM-SHA-256 iteration count from the server-first-message and the stored verifier was parsed with atoi() and used directly as the PBKDF2 loop count. atoi() ignores trailing garbage, accepts signs, and is undefined on overflow, and a large value lets a peer force unbounded PBKDF2 work.

Add scram_parse_iterations(), which accepts only a positive decimal value that fits in an int and is at most SCRAM_MAX_ITERATIONS (100000, well above PostgreSQL's default of 4096), and use it at every call site that previously parsed the count with atoi().

### DIFF
--- a/src/include/security.h
+++ b/src/include/security.h
@@ -39,6 +39,10 @@ extern "C" {
 #include <stdlib.h>
 #include <openssl/ssl.h>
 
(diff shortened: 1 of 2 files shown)


jesperpedersen:
Merged.

PR #1197, opened 2026-06-27
[#1196] Read complete SCRAM messages before parsing

### LINKED ISSUE #1196: SCRAM messages split across network segments can underflow the length-prefixed parsers
A SCRAM authentication message (`AuthenticationSASLContinue` / `SASLFinal`) that arrives split across multiple TCP segments is parsed after the first `read()` returns. The parser then computes `sasl_continue->length - 9` (and similar offsets) on a partial message; because `length` is unsigned, a short message underflows to a huge value that reaches `get_scram_attribute()` / `base64_decode()`, risking a crash. The same single-read assumption exists in every SCRAM exchange (server-side and client-side).

This is the counterpart of pgagroal#834.

### Fix

Add a `read_complete_message()` that keeps reading until the byte count declared in the message header has arrived and rejects a malformed declared length; use it for the SCRAM exchanges; guard every SCRAM parse site against short messages and missing attributes so a malformed message fails authentication cleanly instead of crashing.

### Acceptance

New MCTF regression tests covering single-segment, split-segment, truncated-by-close, and invalid-declared-length cases.

fheikens:
## Problem

A SCRAM message that arrives split across multiple network segments is parsed after the first `read()` returns. `sasl_continue->length - 9` (and the other offset arithmetic) then underflows because `length` is unsigned, and the huge value reaches `get_scram_attribute()` / `pgmoneta_base64_decode()`, risking a crash. The single-read assumption exists in every SCRAM exchange, server-side and client-side.

## Fix

Add `pgmoneta_read_complete_message()`, which keeps reading until the byte count declared in the message header has arrived and rejects a header whose declared length is malformed. Use it for the SCRAM reads. Guard every SCRAM parse site against short messages and missing attributes so a malformed message fails authentication cleanly instead of crashing.

This is the pgmoneta counterpart of pgagroal#834 / pgagroal#932.

## Testing

New MCTF regression tests (`test/testcases/test_message_complete.c`) over a `socketpair`:

- single segment is returned as-is
- a message split across segments is assembled before parsing
- a peer that closes mid-message is an error, not a partial message
- a header declaring an impossible length is rejected

All four pass locally; `clang-format` clean.

close #1196

### DIFF
--- a/src/include/message.h
+++ b/src/include/message.h
@@ -85,6 +85,21 @@ int
 pgmoneta_read_block_message(SSL* ssl, int socket, struct message** msg);
 
 /**
+ * Read a complete typed message in blocking mode. The message must carry the
+ * standard header (1 byte kind + 4 byte length); the read continues until the
+ * number of bytes declared in the header has arrived, so a message split
+ * across multiple network segments is never returned partially. A message
+ * whose declared length is malformed (less than 4 bytes or larger than the
+ * read buffer) is an error.
+ * @param ssl The SSL struct
+ * @param socket The socket descriptor
+ * @param msg The resulting message
+ * @return One of MESSAGE_STATUS_ZERO, MESSAGE_STATUS_OK or MESSAGE_STATUS_ERROR
+ */
+int
+pgmoneta_read_complete_message(SSL* ssl, int socket, struct message** msg);
+
+/**
  * Read a message with a timeout
  * @param ssl The SSL struct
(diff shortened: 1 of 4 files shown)


jesperpedersen:
@fheikens Please, squash

fheikens:
@jesperpedersen Done — it's a single commit now, rebased onto current `main`. Ready when you are.

Issue #1199, opened 2026-06-28
Backup worflow for tablespaces (v14) gives segfault


ashu3103:
**Describe the bug**

On executing the full backup command with Tablespaces (no Tablespaces no bug) the server gives segmentation fault. The error originates from `pgmoneta_receive_archive_files` function in `archive.c` which I think is not well tested for lower PostgreSQL versions with Tablespaces. (@Jubilee101 @jesperpedersen can you confirm this)

pgmoneta Error Log:
```
/home/ashu3103/Desktop/pgmoneta/src/libpgmoneta/archive.c:362:19: runtime error: member access within null pointer of type 'struct message'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /home/ashu3103/Desktop/pgmoneta/src/libpgmoneta/archive.c:362:19
```

PostgreSQL Error Log:
```
2026-06-28 22:08:52.649 IST [200404] STATEMENT:  BASE_BACKUP LABEL 'pgmoneta_20260628220851' FAST WAL NOWAIT MANIFEST 'yes' MANIFEST_CHECKSUMS 'SHA512';
2026-06-28 22:08:52.649 IST [200404] ERROR:  base backup could not send data, aborting backup
2026-06-28 22:08:52.649 IST [200404] STATEMENT:  BASE_BACKUP LABEL 'pgmoneta_20260628220851' FAST WAL NOWAIT MANIFEST 'yes' MANIFEST_CHECKSUMS 'SHA512';
2026-06-28 22:08:52.649 IST [200404] LOG:  could not send data to client: Broken pipe
2026-06-28 22:08:52.649 IST [200404] STATEMENT:  BASE_BACKUP LABEL 'pgmoneta_20260628220851' FAST WAL NOWAIT MANIFEST 'yes' MANIFEST_CHECKSUMS 'SHA512';
2026-06-28 22:08:52.649 IST [200404] FATAL:  connection to client lost
```

**PostgreSQL**

PostgreSQL Version 14

jesperpedersen:
@ashu3103 Yeah, this is untested so there are likely corner-cases that we need to implement to cover 14-16

PR #1200, opened 2026-06-29
[#1199] Backup worflow for tablespaces (v14) gives segfault

ashu3103:
Fix: Memory of a struct was allocated outside while loop and freed each time inside the while loop. This would work if the while loop is executed only once which is the case of no tablespaces (only main data directory to iterate on).


jesperpedersen:
@ashu3103 This is ok, but we need the `if (msg == NULL)` check after `malloc`

jesperpedersen:
Merged.

PR #1212, opened 2026-07-05
[#1211]feat: unify remote storage engines into one parallel upload path

### LINKED ISSUE #1183: Failed backup leaves junk in the cloud
the failure cleanup only deletes the local side.,,nd it doesn't remove what was already uploaded to the remote backends.

### LINKED ISSUE #1211: feat: parallel fan-out for remote storage engines
Add a unified remote upload node that runs S3, Azure, and SSH uploads in parallel threads
also Add MCTF integration tests that bring up all three backends together and prove the uploads actually run in parallel

shashidhar-bm:
- closes #1211 
- also fixes #1183 

changes includes:
- Unify the separate `ssh`/`s3`/`azure` upload paths into one shared node (`se_remote.c`) that runs all three in parallel threads 
- Uses plain `pthread`s (one per engine) for the fan-out, keeping it simple and independent of the worker pool
- Add `mctf_se_up_all()` plus 5 integration tests proving the fan-out lands on every backend and genuinely runs in parallel

### DIFF
--- a/src/libpgmoneta/se_azure.c
+++ b/src/libpgmoneta/se_azure.c
@@ -184,7 +184,6 @@ azure_storage_teardown(char* name __attribute__((unused)), struct art* nodes)
 {
    int server = -1;
    char* label = NULL;
-   char* root = NULL;
    struct main_configuration* config;
 
    config = (struct main_configuration*)shmem;
@@ -198,17 +197,6 @@ azure_storage_teardown(char* name __attribute__((unused)), struct art* nodes)
 
    server = (int)pgmoneta_art_search(nodes, NODE_SERVER_ID);
    label = (char*)pgmoneta_art_search(nodes, NODE_LABEL);
-
-   if (!pgmoneta_is_storage_engine_enabled(STORAGE_ENGINE_LOCAL))
-   {
-      root = pgmoneta_get_server_backup_identifier_data(server, label);
-
-      if (root != NULL)
-      {
-         pgmoneta_delete_directory(root);
-         free(root);
-      }
-   }
 
    pgmoneta_log_debug("Azure storage engine (teardown): %s/%s", config->common.servers[server].name, label);
 
@@ -682,4 +670,21 @@ azure_add_request_headers(struct http_request* request, char* auth_value, char* 
    }
 
    return 0;
+}
+
+int
+azure_upload(int server, char* label, int compression __attribute__((unused)), int encryption __attribute__((unused)))
+{
+   char* local_root = NULL;
+   char* azure_root = NULL;
+   int rc;
+
+   local_root = pgmoneta_get_server_backup_identifier(server, label);
+   azure_root = azure_get_basepath(server, label);
+
+   rc = azure_upload_files(local_root, azure_root, "");
+
+   free(local_root);
+   free(azure_root);
+   return rc;
 }
--- a/src/libpgmoneta/se_remote.c
+++ b/src/libpgmoneta/se_remote.c
@@ -1,0 +1,362 @@
+/*
+ * Copyright (C) 2026 The pgmoneta community
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
+ * are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this list
+ * of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice, this
+ * list of conditions and the following disclaimer in the documentation and/or other
+ * materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its contributors may
+ * be used to endorse or promote products derived from this software without specific
+ * prior written permission.
+ *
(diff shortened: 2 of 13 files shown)


jesperpedersen:
@shashidhar-bm You should be using `struct worker` for this - not a direct `pthread`

src/libpgmoneta/se_azure.c
```diff
@@ -199,17 +198,6 @@ azure_storage_teardown(char* name __attribute__((unused)), struct art* nodes)
    server = (int)pgmoneta_art_search(nodes, NODE_SERVER_ID);
    label = (char*)pgmoneta_art_search(nodes, NODE_LABEL);
 
-   if (!pgmoneta_is_storage_engine_enabled(STORAGE_ENGINE_LOCAL))
```
Amr-Shams, line 202: nice, the problem of teardown, when se_* happens the data dir gets deleted nice,

src/libpgmoneta/se_s3.c
```diff
@@ -3234,3 +3222,44 @@ s3_label_from_common_prefix(char* prefix)
...
+   base_dir = pgmoneta_get_server_backup(server);
+   s3_root = s3_get_basepath(server, label);
+
+   if (pgmoneta_load_info(base_dir, label, &bck))
+   {
+      pgmoneta_log_error("S3 upload: failed to load backup info for %s/%s", base_dir, label);
+      goto done;
+   }
+
+   rc = s3_upload_files(local_root, s3_root, server, bck->compression, bck->encryption);
```
Amr-Shams: this one spawn workers creating a nesting workers, i think it needs to be tested, but the best i can think of is to cap the workers (maybe half the workers enabled)

src/libpgmoneta/se_remote.c
```diff
@@ -0,0 +1,357 @@
...
+         continue;
+      }
+
+      tasks[n].ops = &remote_ops[i];
+      tasks[n].server = server;
+      tasks[n].label = label;
+      tasks[n].result = 1;
+      tasks[n].elapsed = 0.0;
+
+      if (pthread_create(&tasks[n].tid, NULL, backend_thread, &tasks[n]) == 0)
```
Amr-Shams: use  pgmoneta_workers_initialize/pgmoneta_workers_add/pgmoneta_workers_wait from workers

src/libpgmoneta/se_s3.c
```diff
@@ -3234,3 +3222,44 @@ s3_label_from_common_prefix(char* prefix)
...
+   char* base_dir = NULL;
+   char* s3_root = NULL;
+   struct backup* bck = NULL;
+   int rc = 1;
+
+   local_root = pgmoneta_get_server_backup_identifier(server, label);
+   base_dir = pgmoneta_get_server_backup(server);
+   s3_root = s3_get_basepath(server, label);
+
+   if (pgmoneta_load_info(base_dir, label, &bck))
```
Amr-Shams: redundant as the se_remote already has the backup struct on NODE_BACKUP via pgmoneta_art_search(nodes, NODE_BACKUP).

Amr-Shams (commented):
se_ops at se_remote.c:85-93
hardcodes upload/cleanup as function pointers. each new operation adds another field. duplicates the storage engine mapping the workflow pipeline already has.
what if u think about this design? with capabilities:
```
#define STORAGE_CAP_ATOMIC_RENAME  (1 << 0)  /* SSH */
#define STORAGE_CAP_RANGE_GET      (1 << 1)  /* S3, Azure */
#define STORAGE_CAP_BATCH_DELETE   (1 << 2)  /* S3 */
#define STORAGE_CAP_MULTIPART      (1 << 3)  /* S3 */
#define STORAGE_CAP_PARALLEL_SAFE  (1 << 4)  /* S3, Azure */

struct storage_engine {
    int protocol;
    int capabilities;
    char name[32];
    struct workflow* (*create)(int workflow_type);
};
```
- 
- bitmask scales without struct changes
- create(workflow_type) calls the existing pgmoneta_storage_create_* functions directly
- no raw pthreads: each workflow step checks PARALLEL_SAFE, runs in-thread or submits to a shared dispatcher
- replaces se_ops, remote_ops[], backend_thread(), and the whole se_remote node

shashidhar-bm:
> what if u think about this design? with capabilities:

@Amr-Shams ,Really liked this approach. but wondering, since capabilities are defined per engine, how would parallel-safe engines avoid concurrently modifying the shared nodes tree f.ex. s3 list writes to it?

Amr-Shams:
> > what if u think about this design? with capabilities:
> 
> @Amr-Shams ,Really liked this approach. but wondering, since capabilities are defined per engine, how would parallel-safe engines avoid concurrently modifying the shared nodes tree f.ex. s3 list writes to it?

the engines don't modify the shared tree during parallel upload, and for workflows that do, the capabilities bitmask lets se_remote choose serial execution per engine. The thin-wrapper design handles both cases without adding complexity.

shashidhar-bm:
@Amr-Shams , @jesperpedersen , PTAL,

jesperpedersen:
Rebased and merged.

PR #1228, opened 2026-07-28
[#1213]feat: parallel Azure blob upload using worker pool

### LINKED ISSUE #1213: Parallel Azure blob upload using worker pool
With #1212, we introduced parallelism across different storage engines. However, file uploads within the Azure storage engine are still performed one at a time. 
This issue brings worker-pool-based parallel file upload to Azure (same as how s3 upload works today).

shashidhar-bm:
- closes #1213 

benchmark results:
```
case         : backup_azure
iterations   : 5 (median)
build        : Release
machine      : 8 cores, Linux
baseline     : bench-base @ d2e85f2
candidate    : bench-cand @ 334c261

phase                     baseline     candidate  change
--------------------  ------------  ------------  ------------
total                      12690ms        8487ms  1.50x faster
basebackup                  3274ms        3383ms  no change
manifest                      29ms          32ms  no change
hash                          41ms          40ms  no change
linking                       21ms          19ms  no change

phase (emulated)          baseline     candidate  change
--------------------  ------------  ------------  ------------
remote_azure                8514ms        4889ms  1.74x faster

Emulated phases upload to a local container (Azurite, Garage, SFTP),
not a real object store. They show the direction of a change, not a
figure you can quote for production.

Differences below 10% are reported as no change.
```

what changes? 
- Replaced Azure’s serial directory walk with manifest-based uploads and a worker pool, matching S3’s data-first, metadata-last model.
- Moved `SHA512` generation before remote uploads so `backup.sha512` is available before cloud uploads begin.
- Made timestamp formatting thread-safe with `gmtime_r` for parallel Azure and S3 uploads.

### DIFF
--- a/benchmarks/cases/backup_azure.c
+++ b/benchmarks/cases/backup_azure.c
@@ -1,0 +1,54 @@
+/*
+ * Copyright (C) 2026 The pgmoneta community
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
+ * are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this list
+ * of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice, this
+ * list of conditions and the following disclaimer in the documentation and/or other
+ * materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its contributors may
+ * be used to endorse or promote products derived from this software without specific
+ * prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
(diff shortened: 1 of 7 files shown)


Amr-Shams (commented):
Yo, PR looks solid. two things before merging:
1. metadata files (backup.info etc) get uploaded alongside data files right now. same pattern S3 uses at line 1608 as commit markers.
2. also .pgmoneta empty-dir marker: the fopen happens after the upload call, so you're uploading an unopened file. i think this is a n old bud but maybe u can fix it along the way

shashidhar-bm:
> 2\. also .pgmoneta empty-dir marker: the fopen happens after the upload call, so you're uploading an unopened file. i think this is a n old bud but maybe u can fix it along the way

@Amr-Shams , With the current implementation, we use manifest-based uploads, similar to S3, with data files uploaded by workers and metadata uploaded last. We no longer rely on `.pgmoneta` markers. WDYT on this?

jesperpedersen:
Merged.

PR #1233, opened 2026-08-08
[#1232] Fix thread-safe CSV parsing for parallel manifest reads

### LINKED ISSUE #1232: Parallel manifest parsing fails under concurrent reads
While working on Azure parallel workers and remote fan-out [#1211](https://github.com/pgmoneta/pgmoneta/issues/1211), the remote tests failed when `S3` and `Azure` were used concurrently. The tests for` S3` and `Azure` continued to pass when each backend was tested independently.

The manifest is parsed using the CSV reader, whose implementation is not thread-safe when accessed concurrently by multiple threads. This can result in parsing errors and cause uploads to fail even when the manifest file itself is valid.

shashidhar-bm:
- fixes #1232 

Each CSV reader keeps its own saveptr and uses `strtok_r` instead of `strtok`, so parallel manifest parsing (e.g. S3 and Azure at once) no longer shares hidden global state and corrupts columns.

### DIFF
--- a/src/include/csv.h
+++ b/src/include/csv.h
@@ -46,6 +46,7 @@ struct csv_reader
 {
    FILE* file;     /**< The file */
    char line[512]; /**< The line */
+   char* saveptr;  /**< Per-reader strtok_r state (thread-safe) */
 };
 
 /** @struct csv_writer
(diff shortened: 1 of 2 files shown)


jesperpedersen:
Merged.

PR #1235, opened 2026-08-17
[#1234] feat: introduce azure restore path on a shared object storage implementation

### LINKED ISSUE #1234: feat: add restore support for Azure Blob Storage

shashidhar-bm:
- closes #1234 

what changes? 
Extract the object storage restore flow into common functions shared by the engines, and add the azure restore path on top.

### DIFF
--- a/src/libpgmoneta/azure.c
+++ b/src/libpgmoneta/azure.c
@@ -1,0 +1,222 @@
+/*
+ * Copyright (C) 2026 The pgmoneta community
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
+ * are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this list
+ * of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice, this
+ * list of conditions and the following disclaimer in the documentation and/or other
+ * materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its contributors may
+ * be used to endorse or promote products derived from this software without specific
+ * prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+ * OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+ * THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT
+ * OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR
+ * TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/* pgmoneta */
+#include <pgmoneta.h>
+#include <art.h>
+#include <azure.h>
+#include <info.h>
+#include <logging.h>
+#include <management.h>
+#include <network.h>
+#include <progress.h>
+#include <restore.h>
+#include <storage.h>
+#include <utils.h>
+#include <workflow.h>
+#include <workflow_funcs.h>
+
+/* system */
+#include <stdlib.h>
+#include <string.h>
+
+#define NAME "azure"
+
+static bool
+azure_is_safe_label(char* label)
+{
+   if (label == NULL || strlen(label) == 0)
+   {
+      return false;
+   }
+
+   if (pgmoneta_contains(label, "/") || pgmoneta_contains(label, "."))
+   {
+      return false;
+   }
+
+   return true;
+}
+
+void
+pgmoneta_restore_azure_objects(int client_fd, int server, char* label, uint8_t compression, uint8_t encryption, struct json* payload)
+{
+   char* elapsed = NULL;
+   char* en = NULL;
+   int ec = -1;
+   struct timespec start_t;
+   struct timespec end_t;
+   double total_seconds;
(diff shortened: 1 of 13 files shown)


jesperpedersen:
@shashidhar-bm PTAL us when ready for review

src/libpgmoneta/azure.c
```diff
@@ -0,0 +1,225 @@
...
+   clock_gettime(CLOCK_MONOTONIC_RAW, &start_t);
+#endif
+
+   req = (struct json*)pgmoneta_json_get(payload, MANAGEMENT_CATEGORY_REQUEST);
+   position = (char*)pgmoneta_json_get(req, MANAGEMENT_ARGUMENT_POSITION);
+   directory = (char*)pgmoneta_json_get(req, MANAGEMENT_ARGUMENT_DIRECTORY);
+
+   if (!azure_is_safe_prefix(prefix))
+   {
+      ec = MANAGEMENT_ERROR_RESTORE_S3_ERROR;
```
Amr-Shams: maybe it is a typo, but this should be azure

src/libpgmoneta/workflow.c
```diff
@@ -1316,7 +1350,7 @@ get_error_code(int type, int flow, struct art* nodes)
          return -1;
       }
    }
-   else if (type == WORKFLOW_TYPE_S3_RESTORE)
+   else if (type == WORKFLOW_TYPE_S3_RESTORE || type == WORKFLOW_TYPE_AZURE_RESTORE)
```
Amr-Shams: also this would result in the same s3 errors instead of azure

Amr-Shams (commented):
i can see there are like 70% or even more shared between the s3 and azure, while the signature is diff, everything else looks the same, which makes me wonder can we reduce that code? cause gcs would look the same

src/libpgmoneta/azure.c
shashidhar-bm: Yes, it was just a try at how far the S3 restore implementation could be reused, so the S3 codes came along with it, which clearly isn't correct for the Azure path. I'll PTAL once the proper restore implementation is done.

shashidhar-bm:
> i can see there are like 70% or even more shared between the s3 and azure, while the signature is diff, everything else looks the same, which makes me wonder can we reduce that code? cause gcs would look the same

sure @Amr-Shams , I see lot of scope here, i'll think about it :)

shashidhar-bm:
hi @Amr-Shams , @jesperpedersen , please PTAL

src/libpgmoneta/azure.c
```diff
@@ -0,0 +1,222 @@
...
+#include <workflow_funcs.h>
+
+/* system */
+#include <stdlib.h>
+#include <string.h>
+
+#define NAME "azure"
+
+static bool
+azure_is_safe_label(char* label)
```
Amr-Shams: this func is the same as s3 safe label, we can make it in utils file

src/libpgmoneta/se_azure.c
```diff
@@ -834,3 +1006,135 @@ azure_upload(int server, char* label, int compression __attribute__((unused)), i
...
+
+   info_tmp = pgmoneta_append(pgmoneta_append(NULL, local_root), "backup.info.tmp");
+   info_final = pgmoneta_append(pgmoneta_append(NULL, local_root), "backup.info");
+
+   if (pgmoneta_object_bootstrap(&azure_ops, azure_root, server, local_root))
+   {
+      goto error;
+   }
+
+   if (pgmoneta_move_file(info_tmp, info_final))
```
Amr-Shams: phantom backup should be after the full download

src/libpgmoneta/se_azure.c
```diff
@@ -63,13 +66,15 @@ struct azure_transfer_task
...
 static int azure_send_upload_request(char* local_root, char* azure_root, char* relative_path);
-static int azure_add_request_headers(struct http_request* request, char* auth_value, char* utc_date);
+static int azure_add_request_headers(struct http_request* request, char* auth_value, char* utc_date, bool blob_type);
 static void do_upload_file(struct worker_common* wc);
 static int azure_create_transfer_task(int server, char* azure_root, char* remote_path,
                                       char* local_root, char* local_path,
                                       struct workers* workers, struct azure_transfer_task** task);
 static int azure_upload_one_file(struct azure_transfer_task* task);
 
+static int azure_send_get_request(char* azure_root, char* relative_path, struct http_response** response);
```
Amr-Shams, line 76: signing block looks the same for both the azure_send_upload_request and the function

src/libpgmoneta/se_azure.c
```diff
@@ -834,3 +1006,135 @@ azure_upload(int server, char* label, int compression __attribute__((unused)), i
...
+   {
+      pgmoneta_log_error("Azure restore: could not rename %s to %s", info_tmp, info_final);
+      goto error;
+   }
+
+   base_dir = pgmoneta_get_server_backup(server);
+
+   if (pgmoneta_load_info(base_dir, label, &backup))
+   {
+      pgmoneta_log_error("Azure restore: failed to load backup.info from %s", local_root);
```
Amr-Shams: use base_dir

Amr-Shams (commented):
i am not sure staging in place is the right call here, but i know it is fine with the current usage

src/libpgmoneta/se_azure.c
shashidhar-bm: I think we should fix this in S3 as well?

shashidhar-bm:
> i am not sure staging in place is the right call here, but i know it is fine with the current usage

@Amr-Shams, did you mean to have a seperate staging dir for s3/azure?

Amr-Shams:
> > i am not sure staging in place is the right call here, but i know it is fine with the current usage
> 
> @Amr-Shams, did you mean to have a seperate staging dir for s3/azure?

yes, temp dir is a way to load till finish, the thing is deleting or overwriting might in the future be buggy when another process uses the same backup dir

src/libpgmoneta/se_azure.c
Amr-Shams: mmmm, yes, but it is not simple change cause in s3 we are loading the backup info which means we need that move, the change is needed but not straight

shashidhar-bm:
along with review changes, i've added testcase that starts PostgreSQL on the restored Azure backup

Amr-Shams (approved):

src/libpgmoneta/se_azure.c
```diff
@@ -548,82 +708,34 @@ azure_send_upload_request(char* local_root, char* azure_root, char* relative_pat
...
-
-   if (pgmoneta_base64_decode(config->azure_shared_key, strlen(config->azure_shared_key), (void**)&signing_key, &signing_key_length))
+   if (azure_build_authorization(signed_headers_prefix, azure_path, utc_date, &auth_value))
    {
       goto error;
    }
 
-   if (pgmoneta_generate_string_hmac_sha256_hash(signing_key, signing_key_length, string_to_sign, strlen(string_to_sign), &signature_hmac, &hmac_length))
+   if (azure_http_prepare(azure_path, &azure_host, azure_put_path, sizeof(azure_put_path), &connection))
    {
```
Amr-Shams, line 729: jut few debug logs would be great here

shashidhar-bm:
@jesperpedersen PTAL

PR #1242, opened 2026-08-21
[#1241] Fix user index lookup and guard the not-found case

### LINKED ISSUE #1241: User lookup: server index assigned to user index in main.c, and four unguarded users[-1] accesses
Two related defects in the "find the user index for a server" pattern, found by running `cppcheck` over `src/` (the repo has no static analysis in CI — `ci.yml` runs clang-format, a build and the test suite).

The pattern appears **8 times**. Four sites guard the sentinel, four do not.

## 1. `main.c` assigns the server index into the user index

`src/main.c:2712`:

```c
for (int i = 0; i < config->common.number_of_servers; i++)      // i = server
{
   ...
   for (int j = 0; usr == -1 && j < config->common.number_of_users; j++)   // j = user
   {
      if (pgmoneta_compare_string(config->common.servers[i].username, config->common.users[j].username))
      {
         usr = i;        // <-- assigns the SERVER index
      }
   }

   auth = pgmoneta_server_authenticate(i, "postgres",
                                       config->common.users[usr].username,
                                       config->common.users[usr].password, ...);
```

`usr` is used only to index `config->common.users[]`, but the match assigns `i`, the server index. Every other site in the codebase assigns the user loop variable.

Consequences when a user *is* found:

- if the server index differs from the matching user index, the connection is attempted with **another user's credentials**, and the failure is then logged against that other user's name;
- if the server index is >= `number_of_users`, this reads past the end of the array.

It looks like a rename of the inner loop variable to `j` that missed this line.

## 2. Four sites use `users[usr]` without checking for -1

`usr` stays `-1` when the server's username matches no configured user, and these sites index with it immediately:

| site | uses `users[usr]` at |
|---|---|
| `src/libpgmoneta/wf_backup.c:158` | 166 |
| `src/libpgmoneta/wf_backup_incremental.c:260` | 268 |
| `src/libpgmoneta/wf_backup_incremental.c:739` | 747 |
| `src/main.c:2712` | 2721 (same site as above) |

`config->common.users[-1]` is an out-of-bounds read.

These four already guard it, which is what makes the omission look accidental rather than an invariant:

| site | guard |
|---|---|
| `src/libpgmoneta/wal.c:148` | `if (usr == -1)` at 156, logs and `goto error` |
| `src/libpgmoneta/wf_extra.c:129` | `if (usr == -1)` at 137, logs and `goto error` |
| `src/libpgmoneta/extension.c:289` | `if (usr == -1)` at 297 |
| `src/main.c:3066` | `if (usr != -1)` at 3075 |

`wf_extra.c` and `wf_backup.c` are near-identical here — same comment, same `pgmoneta_server_authenticate` call — but only `wf_extra.c` has the guard.

## Reachability

I could not find anything in configuration loading that requires a server's `user` to exist in `pgmoneta_users.conf`; validation checks password encoding, not cross-references. The four existing guards log "Invalid user for %s" / "User not found for server: %d", so the authors evidently consider the case reachable at runtime — otherwise those branches would be dead.

I have verified this by reading the source and with cppcheck (`negativeIndex`, `Array 'config->common.users[64]' accessed at index -1`); I have not driven it with a live misconfiguration.

Happy to send a PR: fix the index in `main.c` and add the missing guards following the existing `wal.c` / `wf_extra.c` convention. Per CONTRIBUTING I can split it into two pull requests if you would prefer the index fix separate from the guards.

Pranav-error:
Fixes #1241.

## What

**`src/main.c`** — the outer loop indexes servers (`i`), the inner loop indexes users (`j`), but the match assigned `usr = i`. `usr` is only ever used to index `config->common.users[]`, so a successful match picked another user's credentials whenever the two indices differed, and read past the end of the array when the server index exceeded `number_of_users`. Now assigns `j`.

**Four sites indexed `users[usr]` without checking for the `-1` sentinel**, which reads `users[-1]` when a server's username matches no configured user:

- `src/libpgmoneta/wf_backup.c`
- `src/libpgmoneta/wf_backup_incremental.c` (two sites)
- `src/main.c` (the same site as above)

`wal.c`, `wf_extra.c`, `extension.c` and the other `main.c` site already guard this and log before bailing out, so the added guards follow that convention — `pgmoneta_log_error` then `goto error`, except in `main.c` where the lookup sits inside the per-server loop and `continue` is the equivalent.

`wf_extra.c` and `wf_backup.c` are near-identical around the lookup, same comment and same `pgmoneta_server_authenticate` call, with only `wf_extra.c` carrying the guard — which is what suggested these were omissions rather than an invariant.

## Verification

- Builds clean.
- `clang-format` reports no changes for the three touched files.
- `cppcheck --enable=warning -I src/include src/` previously reported four `negativeIndex` errors (`Array 'config->common.users[64]' accessed at index -1`); all four are gone, and no new findings are introduced.

I have not driven this with a live misconfigured server, so the runtime symptom is reasoned from the code path rather than observed.

## Notes

Kept to one commit per CONTRIBUTING. Happy to split the index fix from the guards if you would rather review them separately.

Unrelated, and deliberately **not** included here: `src/libpgmoneta/walfile/wal_summary.c` does not compile on macOS because `mkdtemp` is declared in `<unistd.h>` there while glibc declares it in `<stdlib.h>`, and neither header is included directly. I patched it locally to get a build. Say the word and I will open that as its own issue and PR.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -57,3 +57,4 @@ Harshit Shaw <shawharshit116@gmail.com>
 Ahmed Kamal <ahmedkamal200427@gmail.com>
 Rohan Mishra <kmrrohan29@gmail.com>
 Mostafa Mahmoud <mm1471800@gmail.com>
+R Sai Pranav <rajasaipranav0@gmail.com>
--- a/doc/manual/en/97-acknowledgement.md
+++ b/doc/manual/en/97-acknowledgement.md
@@ -64,6 +64,7 @@ Harshit Shaw <shawharshit116@gmail.com>
 Ahmed Kamal <ahmedkamal200427@gmail.com>
 Rohan Mishra <kmrrohan29@gmail.com>
 Mostafa Mahmoud <mm1471800@gmail.com>
+R Sai Pranav <rajasaipranav0@gmail.com>
 ```
 
 ## Committers
--- a/src/libpgmoneta/wf_backup.c
+++ b/src/libpgmoneta/wf_backup.c
@@ -162,6 +162,13 @@ basebackup_execute(char* name __attribute__((unused)), struct art* nodes)
          usr = i;
       }
    }
+
+   if (usr == -1)
+   {
+      pgmoneta_log_error("User not found for server: %d", server);
+      goto error;
+   }
+
    // establish a connection, with replication flag set
    if (pgmoneta_server_authenticate(server, "postgres", config->common.users[usr].username, config->common.users[usr].password, false, &ssl, &socket) != AUTH_SUCCESS)
    {
--- a/src/libpgmoneta/wf_backup_incremental.c
+++ b/src/libpgmoneta/wf_backup_incremental.c
@@ -264,6 +264,13 @@ incr_backup_execute_14_to_16(char* name __attribute__((unused)), struct art* nod
          usr = i;
       }
    }
+
+   if (usr == -1)
+   {
+      pgmoneta_log_error("User not found for server: %d", server);
+      goto error;
+   }
+
    // establish a connection, with replication flag set
    if (pgmoneta_server_authenticate(server, "postgres", config->common.users[usr].username, config->common.users[usr].password, false, &ssl, &socket) != AUTH_SUCCESS)
    {
@@ -743,6 +750,13 @@ incr_backup_execute_17_plus(char* name __attribute__((unused)), struct art* node
          usr = i;
       }
    }
+
+   if (usr == -1)
+   {
+      pgmoneta_log_error("User not found for server: %d", server);
+      goto error;
+   }
+
    // establish a connection, with replication flag set
    if (pgmoneta_server_authenticate(server, "postgres", config->common.users[usr].username, config->common.users[usr].password, false, &ssl, &socket) != AUTH_SUCCESS)
    {
--- a/src/main.c
+++ b/src/main.c
@@ -2713,8 +2713,14 @@ valid_cb(struct ev_loop* loop __attribute__((unused)), ev_periodic* w __attribut
             {
                if (pgmoneta_compare_string(config->common.servers[i].username, config->common.users[j].username))
                {
-                  usr = i;
+                  usr = j;
                }
+            }
+
+            if (usr == -1)
+            {
+               pgmoneta_log_error("User not found for server: %s", config->common.servers[i].name);
+               continue;
             }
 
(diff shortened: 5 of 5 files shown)


jesperpedersen:
@Pranav-error You will need to update the author files -- see DEVELOPERS.md

Pranav-error:
Thanks @jesperpedersen — done. Added myself to `AUTHORS` and `doc/manual/en/97-acknowledgement.md`, amended into the existing commit so the branch stays at one commit as DEVELOPERS.md asks.

Note the `AUTHORS` diff shows as `3 ++-` because the file has no trailing newline after the final entry; I kept that convention rather than introducing one.

Should I do the same on #1244 and #1246, or is the once-only entry from this PR enough?

jesperpedersen:
@Pranav-error Only necessary in this PR

jesperpedersen:
@Pranav-error Please, start to cross-port all your PRs to pgagroal, pgexporter and pgvictoria

Pranav-error:
Understood on the author files — thanks, left #1244 and #1246 alone then.

Cross-ports are up as you asked:

- pgagroal/pgagroal#999
- pgexporter/pgexporter#576
- pgvictoria/pgvictoria#51

Those are the `network.c` `star_fds` fix from #1246. `network.c` is character-identical across all four projects, so it applied unchanged; I added myself to `AUTHORS` and the acknowledgement page in each since they are first PRs there.

The other two do not have equivalents elsewhere — the user-index lookup here and the `mkdtemp` include in #1244 are both pgmoneta-specific, and cppcheck finds nothing matching in the other three.

CI on this PR is green across all eight jobs.

One thing worth deciding as a group rather than per-repo: all four projects report `nullPointerArithmeticOutOfMemory` from unchecked `malloc` (`pgagroal/security.c`, `pgexporter/queries.c`, `pgvictoria/security.c` and `utils.c`, `pgmoneta/restore.c` and `utils.c`). I left those out of #1245 on purpose — in pgmoneta roughly 204 of 342 allocations are checked, so it is a broad pattern rather than a few oversights, and picking off five sites seemed arbitrary. If you want that swept properly I am happy to do it across all four, or to add cppcheck to CI so the class stops reappearing. Your call on scope.

jesperpedersen:
Merged.

jesperpedersen:
Welcome to the community !

jesperpedersen:
@Pranav-error We should fix all warnings/errors found by clang or cppcheck, so go ahead with the PRs for all projects

PR #1244, opened 2026-08-21
[#1243] Include unistd.h in wal_summary.c for mkdtemp

### LINKED ISSUE #1243: wal_summary.c fails to compile on macOS: mkdtemp not declared
Building `main` on macOS (Apple clang, Xcode SDK) fails:

```
src/libpgmoneta/walfile/wal_summary.c:205:8: error: call to undeclared function 'mkdtemp';
    ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
src/libpgmoneta/walfile/wal_summary.c:205:33: error: comparison between pointer and integer
    ('int' and 'void *') [-Werror,-Wpointer-integer-compare]
2 errors generated.
```

The second error is a consequence of the first: with no declaration, `mkdtemp` is assumed to return `int`, and the `== NULL` comparison then trips `-Werror,-Wpointer-integer-compare`.

## Cause

`wal_summary.c` includes neither `<stdlib.h>` nor `<unistd.h>` directly. It picks up `<stdlib.h>` transitively via `pgmoneta.h:44`.

- On glibc, `<stdlib.h>` declares `mkdtemp`, so Linux CI compiles fine.
- On macOS it does not. The declaration lives in `<unistd.h>` only (`SDK/usr/include/unistd.h:679`).

Minimal reproduction on macOS:

```c
#include <stdlib.h>
int main(void){ char t[]="/tmp/xxXXXXXX"; return mkdtemp(t)==0; }
```

```
error: call to undeclared function 'mkdtemp'
```

Adding `#include <unistd.h>` compiles.

POSIX.1-2008 does specify `mkdtemp` in `<stdlib.h>`, so macOS is the one deviating here — but including `<unistd.h>` is portable, harmless on Linux, and is already what the rest of this codebase does.

## Consistent with the rest of the repo

Every other file that uses `mkdtemp`/`mkstemp` already includes `<unistd.h>`:

| file | `#include <unistd.h>` |
|---|---|
| `src/admin.c` | yes |
| `src/walinfo.c` | yes |
| `src/walfilter.c` | yes |
| `src/libpgmoneta/utils.c` | yes |
| `src/libpgmoneta/walfile/wal_summary.c` | **no** |

So this looks like a missed include rather than a deliberate choice.

## Fix

One line — add `#include <unistd.h>` to the system include block in `wal_summary.c`. With that applied the full build succeeds on macOS and produces all six binaries.

## Scope

I hit this on macOS. I have **not** been able to test FreeBSD, which the README lists as a supported platform, so I am not claiming it is affected — only noting that this is the class of header divergence an explicit include protects against.

Happy to send the one-line PR.

Pranav-error:
Fixes #1243.

One line. `wal_summary.c` includes neither `<stdlib.h>` nor `<unistd.h>` directly and picks up `<stdlib.h>` transitively through `pgmoneta.h`. glibc declares `mkdtemp` there, which is why Linux CI is unaffected, but macOS declares it only in `<unistd.h>`, so the build fails:

```
wal_summary.c:205:8: error: call to undeclared function 'mkdtemp' [-Wimplicit-function-declaration]
wal_summary.c:205:33: error: comparison between pointer and integer ('int' and 'void *')
                            [-Werror,-Wpointer-integer-compare]
```

The second error follows from the first: with no declaration `mkdtemp` is assumed to return `int`, so `== NULL` compares an integer against a pointer.

Every other file in the tree that uses `mkdtemp` or `mkstemp` — `admin.c`, `walinfo.c`, `walfilter.c`, `utils.c` — already includes `<unistd.h>`; this file was the only one that did not.

## Verification

- Before: build fails on macOS at `wal_summary.c`.
- After: full build succeeds and produces all six binaries (`pgmoneta`, `pgmoneta-admin`, `pgmoneta-cli`, `pgmoneta-config`, `pgmoneta-walfilter`, `pgmoneta-walinfo`).
- `clang-format` reports no changes for the file.
- No behaviour change on Linux — the declaration is already visible there via `<stdlib.h>`.

Independent of #1242, which touches different files.

### DIFF
--- a/src/libpgmoneta/walfile/wal_summary.c
+++ b/src/libpgmoneta/walfile/wal_summary.c
@@ -41,6 +41,7 @@
 #include <dirent.h>
 #include <errno.h>
 #include <libgen.h>
+#include <unistd.h>
 
 static char* summary_file_name(uint64_t s_lsn, uint64_t e_lsn);
 static int summarize_walfile(char* path, uint64_t start_lsn, uint64_t end_lsn, block_ref_table* brt);


jesperpedersen:
Rebased and merged.

PR #1246, opened 2026-08-21
[#1245] Check realloc results before using them

### LINKED ISSUE #1245: Three unchecked realloc/malloc results leak the original buffer and then dereference NULL
Continuing the `cppcheck` pass from #1241 (the repo has no static analysis in CI).

Of **18 realloc assignments** in `src/`, **15 already check the result**, and they all use the safe shape — realloc into a *new* variable, test it, and only then commit to it. Three do not, and each one both leaks the original allocation and immediately dereferences the NULL result.

| site | shape | on failure |
|---|---|---|
| `src/libpgmoneta/csv.c:81` | `cs = realloc(cs, ...)` | original `cs` leaked, then `cs[num] = col` writes through NULL |
| `src/libpgmoneta/network.c:115` | `star_fds = realloc(star_fds, ...)` | original leaked, then `memcpy(star_fds + star_length, ...)` |
| `src/libpgmoneta/memory.c:126` | `d = realloc(orig, s)` | `orig` leaked (caller replaces its pointer with the return value), then `memcpy(d + orig_size, ...)` |

The self-assigning form (`p = realloc(p, n)`) is the classic version of this: on failure `realloc` returns NULL *without* freeing the original, so overwriting the only pointer to it loses the buffer.

`src/libpgmoneta/network.c:108` has the same problem one branch earlier — `star_fds = malloc(...)` followed directly by `memcpy` with no check.

## The convention already in the tree

`pgmoneta_append` in `src/libpgmoneta/utils.c:1542`:

```c
n = (char*)realloc(orig, orig_length + s_length + 1);

if (n == NULL)
{
   return orig;
}

memcpy(n + orig_length, s, s_length);
```

New variable, checked, and on failure the caller keeps its original pointer. The three sites above can follow exactly this.

## Scope — deliberately narrow

I am **not** proposing to check every allocation in the tree. A quick pass says roughly 204 of 342 allocations are followed by a NULL check, so unchecked allocation is a broader pattern and changing 138 sites would be a different discussion. These three are worth separating because they are the realloc-specific case, the codebase is already 15-for-18 consistent on it, and each one leaks *and* crashes rather than only crashing.

These are OOM-only paths, so the practical severity is low. Reported because they are cheap to fix and inconsistent with the surrounding code.

PR to follow.

Pranav-error:
Fixes #1245.

Three sites overwrote the only pointer to a buffer with the result of `realloc` and then dereferenced it unchecked. Since `realloc` returns NULL *without* freeing the original, each one leaked the original allocation and then wrote through NULL.

| file | before | after |
|---|---|---|
| `csv.c:81` | `cs = realloc(cs, ...)` then `cs[num] = col` | realloc into `tmp`, on NULL `goto error` — the existing label frees the still-valid `cs` |
| `network.c:115` | `star_fds = realloc(star_fds, ...)` then `memcpy` | realloc into `tmp_fds`, on NULL free `new_fds` and `continue`, matching what the loop already does on bind failure |
| `network.c:108` | `star_fds = malloc(...)` then `memcpy` | same treatment — this branch had the identical problem |
| `memory.c:126` | `d = realloc(orig, s)` then `memcpy` | on NULL, restore `*new_size = orig_size` and return `orig` unchanged |

This follows the shape already used by `pgmoneta_append` (`utils.c:1542`): realloc into a new variable, check it, and only then commit to it. 15 of the 18 realloc assignments in the tree already do this.

The `memory.c` case needed a decision about ownership, since the caller (`message.c:1200`) does `data = pgmoneta_memory_dynamic_append(data, ...)`. Returning the original pointer and the original size keeps the caller's buffer valid and drops only the append, which is the same choice `pgmoneta_append` already makes.

## Verification

- Builds clean.
- `cppcheck --enable=warning -I src/include src/` previously reported two `memleakOnRealloc` errors; both are gone and no new findings appear.
- `clang-format` reports no changes for the three files.

## Scope

Deliberately limited to the realloc-specific case. Roughly 204 of 342 allocations in the tree are followed by a NULL check, so unchecked allocation generally is a broader pattern I have not touched here — see #1245 for that reasoning.

These are OOM-only paths, so practical severity is low.

### DIFF
--- a/src/libpgmoneta/csv.c
+++ b/src/libpgmoneta/csv.c
@@ -62,6 +62,7 @@ bool
 pgmoneta_csv_next_row(struct csv_reader* reader, int* num_col, char*** cols)
 {
    char** cs = NULL;
+   char** tmp = NULL;
    char* col = NULL;
    char* last_tok = NULL;
    int num = 0;
@@ -78,7 +79,12 @@ pgmoneta_csv_next_row(struct csv_reader* reader, int* num_col, char*** cols)
    col = strtok_r(reader->line, ",", &reader->saveptr);
    while (col != NULL)
    {
-      cs = realloc(cs, (num + 1) * sizeof(char*));
+      tmp = realloc(cs, (num + 1) * sizeof(char*));
+      if (tmp == NULL)
+      {
+         goto error;
+      }
+      cs = tmp;
       cs[num] = col;
       num++;
       col = strtok_r(NULL, ",", &reader->saveptr);
--- a/src/libpgmoneta/memory.c
+++ b/src/libpgmoneta/memory.c
@@ -124,6 +124,11 @@ pgmoneta_memory_dynamic_append(void* orig, size_t orig_size, void* append, size_
    {
       s = orig_size + append_size;
       d = realloc(orig, s);
+      if (d == NULL)
+      {
+         *new_size = orig_size;
+         return orig;
+      }
       memcpy(d + orig_size, append, append_size);
    }
    else
--- a/src/libpgmoneta/network.c
+++ b/src/libpgmoneta/network.c
@@ -83,6 +83,7 @@ pgmoneta_bind(char* hostname, int port, int** fds, int* length)
              (ifa->ifa_flags & IFF_UP))
          {
             int* new_fds = NULL;
+            int* tmp_fds = NULL;
             int new_length = 0;
 
             memset(addr, 0, sizeof(addr));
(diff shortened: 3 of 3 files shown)


Pranav-error:
Thanks @jesperpedersen — cross-ported this one:

- pgagroal/pgagroal#999
- pgexporter/pgexporter#576
- pgvictoria/pgvictoria#51

`network.c` is character-identical across all four, so the same `star_fds` fix applies unchanged. Added myself to `AUTHORS` and the acknowledgement page in each, since they are first PRs there.

#1242 and #1244 do not port over — the user-index lookup and `mkdtemp` are pgmoneta-specific and cppcheck finds no equivalent in the other three. What the others do share is `nullPointerArithmeticOutOfMemory` on unchecked `malloc` (`pgagroal/security.c`, `pgexporter/queries.c`, `pgvictoria/security.c` and `utils.c`), which is the same broader unchecked-allocation pattern I described in #1245 and deliberately did not touch there. Happy to take that on across all four if you want it as a single sweep.

Pranav-error:
The red check here is not from this change — `build-linux (clang, Debug, 15)` is failing on `main` as well.

- This PR, run [32503745387](https://github.com/pgmoneta/pgmoneta/actions/runs/32503745387): `The action 'Run Tests' has timed out after 45 minutes`, last test through was `test_aes_cbc_decrypt_failures`.
- `main`, run [32370667187](https://github.com/pgmoneta/pgmoneta/actions/runs/32370667187): same job, same `timed out after 45 minutes`, last test through was `test_not_double_compression_incremental_backup`.

Same configuration, same timeout, different last test — so it looks like a hang somewhere in the suite on that one matrix entry rather than a single bad test, and it predates this branch. Every other job here is green, including the other clang Debug entries (14 and 17).

For what it is worth I did check the obvious way this change could hang: the `continue` I added on allocation failure sits inside `for (ifa = ifaddr; ifa != NULL; ifa = ifa->ifa_next)`, so the increment still runs and it cannot spin. The other two hunks are in `csv.c` and `memory.c` and are not reachable from the test that stalled.

Happy to look into the timeout separately if it is not already known.

PR #1248, opened 2026-08-21
[#1247] Clear freed pointers in add_user and update_user

### LINKED ISSUE #1247: Double free in pgmoneta-admin add_user and update_user on the error path
Found with the clang static analyzer (`clang --analyze` over a compile database). That is a different tool from the `cppcheck` pass in #1241/#1245 and finds a different class — it reports 10 `Attempt to free released memory` sites across the tree; the two in `src/admin.c` are the ones I have read through and confirmed.

## add_user (`src/admin.c:614`)

The success path frees six pointers and does not clear them:

```c
   fputs(entry, users_file);

   free(entry);
   free(master_key);
   free(encrypted);
   free(encoded);
   if (do_free)
   {
      free(password);
   }
   free(verify);

   fflush(users_file);
   fclose(users_file);
   ...
   if (pgmoneta_management_create_outcome_success(j, start_t, end_t, &outcome))
   {
      goto error;
   }

   if (create_response(users_path, j, &response))
   {
      goto error;
   }
```

and the `error:` label frees the same pointers again:

```c
error:

   free(entry);
   free(master_key);
   free(encrypted);
   free(encoded);
   if (do_free)
   {
      free(password);
   }
```

So if either `pgmoneta_management_create_outcome_success` or `create_response` fails — both of which run *after* the frees — each of those pointers is freed twice.

## update_user (`src/admin.c:959`)

Same shape. `master_key`, `encrypted`, `encoded`, `password` and `verify` are freed before `fclose(users_file)`, two `goto error` statements follow, and the `error:` label frees them again.

## Impact

Double free is undefined behaviour and corrupts the allocator; with glibc it typically aborts with `free(): double free detected in tcache 2`. It requires the error path to be taken, so it is not hit in normal operation — but this is the tool that writes credentials.

Unlike the `realloc` findings in #1245, this does **not** require an allocation failure to trigger.

## Fix

Clear each pointer after the free in the success path so the `error:` label's `free()` becomes a no-op. That matches what the surrounding code already does for file handles (`users_file = NULL;` immediately after `fclose`).

Verified by reading the code and with the analyzer, not by forcing `create_response` to fail at runtime.

PR to follow. The other eight `Attempt to free released memory` sites — `extension.c:402`, `management.c:1533`, `message.c:2639`, `utils.c:5170`, `walfile/rm_standby.c:191`, `walfile/wal_reader.c:1226`, `wf_backup_incremental.c:1508`, `wf_retention.c:230` — come from the same run but I have not read through them, so I am not claiming them here. Happy to work through those separately.

Pranav-error:
Fixes #1247.

`add_user` and `update_user` free `entry`, `master_key`, `encrypted`, `encoded`, `password` and `verify` on the success path, then call `pgmoneta_management_create_outcome_success` and `create_response`, both of which `goto error`. The `error:` label frees the same pointers again.

Setting them to NULL after the free makes the second free a no-op, which is what the surrounding code already does for `users_file`.

`clang --analyze` reported `Attempt to free released memory` at `admin.c:919` and `admin.c:1287` before the change and reports neither after. Builds clean, no `clang-format` changes.

### DIFF
--- a/src/admin.c
+++ b/src/admin.c
@@ -872,14 +872,20 @@ password:
    fputs(entry, users_file);
 
    free(entry);
+   entry = NULL;
    free(master_key);
+   master_key = NULL;
    free(encrypted);
+   encrypted = NULL;
    free(encoded);
+   encoded = NULL;
    if (do_free)
    {
       free(password);
+      password = NULL;
    }
    free(verify);
+   verify = NULL;
 
    fflush(users_file);
    fclose(users_file);
@@ -1237,13 +1243,18 @@ password:
    }
 
    free(master_key);
+   master_key = NULL;
    free(encrypted);
+   encrypted = NULL;
    free(encoded);
+   encoded = NULL;
    if (do_free)
    {
       free(password);
+      password = NULL;
    }
    free(verify);
+   verify = NULL;
 
    fclose(users_file);
    users_file = NULL;


Amr-Shams:
@Pranav-error great work, just small note, can you reduce the pr body a little (it is way too verbose), but keep it up

Pranav-error:
Thanks, that's fair — I've trimmed this one down. I'll keep the rest shorter too: what's broken, the fix, and the analyzer result.

PR #1250, opened 2026-08-21
Add a cppcheck job to CI, gated on a baseline

Pranav-error:
Following up on the offer in #1242 / #1247. Easy to close if you would rather not have this — it is a policy decision, not a bug fix.

CI currently runs clang-format, builds and tests, but no static analysis. That is why the defects behind #1241, #1245 and #1247 sat there: nothing was looking for them.

## Why it is a baseline rather than a plain run

`cppcheck --enable=warning` on `main` reports **244 findings**. Failing on any of them would break every build immediately, and dumping 244 warnings into the log that nobody can act on is worse than not running it.

So the job compares against a checked-in baseline of what already exists and fails only when a file gains findings of a kind it did not have, or gains **more** of a kind it already had. New code cannot add to the list; the existing 244 can be worked off whenever you feel like it. When a finding is fixed the job prints which baseline line to lower, so the file does not silently drift.

The baseline records counts per `id` and `file`, not line numbers, so ordinary edits do not churn it.

## One detail worth knowing

The job asks cppcheck for `{id}:{file}:{line}` and then strips the line number. That looks redundant but is not: **cppcheck collapses messages that format identically**, so with a `{id}:{file}` template a file reports at most one finding of each kind no matter how many it actually has.

I had written it that way first and the check silently could never fail. I found it by deliberately injecting a `p = realloc(p, n)` into `csv.c` and watching the job pass. Worth mentioning because the same trap applies to anyone regenerating the baseline by hand.

## Verified

- Clean tree: passes.
- With a `memleakOnRealloc` injected into `csv.c`: fails with
  `memleakOnRealloc:src/libpgmoneta/csv.c: baseline 1 -> now 2`.
- Removing the injection: passes again.
- The workflow YAML parses.

## Caveat I cannot check from here

The baseline was generated with **cppcheck 2.21.0 on macOS**. `ubuntu-latest` ships a different version and may report a slightly different set. If the first CI run fails on version drift rather than on a real regression, tell me and I will regenerate the baseline from that run's output — or pin the cppcheck version in the job, whichever you prefer.

## Interaction with the open fix PRs

#1242, #1246, #1248 and #1249 each remove findings that are in this baseline. Whichever lands first, the others will need their baseline lines lowered — the job prints exactly which. Happy to sequence this one last to keep that simple.

### DIFF
--- a/.github/cppcheck-baseline.txt
+++ b/.github/cppcheck-baseline.txt
@@ -1,0 +1,76 @@
+# cppcheck findings known to exist on main, as 'id:file count' pairs.
+#
+# CI fails when a file gains findings of a kind it did not have, or gains more of a
+# kind it already had, so new code cannot add to this list.
+#
+# Counts rather than line numbers, so ordinary edits do not churn the file. The
+# line number is asked for and then stripped, because cppcheck collapses messages
+# that format identically - without it every file would report at most one finding
+# per kind and the check could never fail.
+#
+# When a finding is fixed, lower the count or delete the line in the same PR.
+#
+# Regenerate with:
+#   cppcheck --enable=warning --inline-suppr --quiet --template='{id}:{file}:{line}' \
+#       -I src/include src/ 2>&1 | grep -E '^[a-zA-Z]+:' | sed 's/:[0-9]*$//' \
+#       | sort | uniq -c | awk '{printf "%s %s
+#
+", $2, $1}' | sort
+IOWithoutPositioning:src/admin.c 1
+arrayIndexOutOfBoundsCond:src/libpgmoneta/utils.c 2
+ctunullpointer:src/libpgmoneta/message.c 2
+ctunullpointer:src/libpgmoneta/wf_backup_incremental.c 1
+ctunullpointer:src/libpgmoneta/workflow.c 1
+ctunullpointerOutOfMemory:src/libpgmoneta/info.c 1
+ctunullpointerOutOfMemory:src/libpgmoneta/json.c 1
+identicalInnerCondition:src/libpgmoneta/se_ssh.c 1
+identicalInnerCondition:src/libpgmoneta/wf_link.c 1
+invalidscanf:src/libpgmoneta/utils.c 1
+invalidscanf:src/libpgmoneta/wf_backup_incremental.c 2
+memleakOnRealloc:src/libpgmoneta/csv.c 1
+memleakOnRealloc:src/libpgmoneta/network.c 1
+nullPointer:src/libpgmoneta/logging.c 1
+nullPointerArithmeticOutOfMemory:src/libpgmoneta/restore.c 2
+nullPointerArithmeticOutOfMemory:src/libpgmoneta/utils.c 1
+nullPointerOutOfMemory:src/libpgmoneta/art.c 17
+nullPointerOutOfMemory:src/libpgmoneta/brt.c 3
+nullPointerOutOfMemory:src/libpgmoneta/bzip2_compression.c 4
+nullPointerOutOfMemory:src/libpgmoneta/compression.c 4
+nullPointerOutOfMemory:src/libpgmoneta/console.c 1
+nullPointerOutOfMemory:src/libpgmoneta/deque.c 9
+nullPointerOutOfMemory:src/libpgmoneta/gzip_compression.c 4
+nullPointerOutOfMemory:src/libpgmoneta/json.c 9
+nullPointerOutOfMemory:src/libpgmoneta/keep.c 1
+nullPointerOutOfMemory:src/libpgmoneta/lz4_compression.c 6
+nullPointerOutOfMemory:src/libpgmoneta/message.c 13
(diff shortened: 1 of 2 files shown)


Pranav-error:
Rebased onto `main` now that #1242 and #1243 have landed, and updated the baseline accordingly.

#1242 removed all five `negativeIndex` findings, so these three lines are gone from `.github/cppcheck-baseline.txt`:

```
negativeIndex:src/libpgmoneta/wf_backup.c 1
negativeIndex:src/libpgmoneta/wf_backup_incremental.c 2
negativeIndex:src/main.c 2
```

Baseline is now 54 entries covering 239 findings, and matches `main` exactly — the job reports no drift in either direction.

Re-verified after the rebase: clean tree passes, an injected `p = realloc(p, n)` in `csv.c` still fails the job with `memleakOnRealloc:src/libpgmoneta/csv.c: baseline 1 -> now 2`, and removing it passes again.

#1246, #1248 and #1249 will each remove more entries when they land. Rather than have this PR go stale every time, I am happy to hold it until those are decided and rebase once at the end — just say which you prefer.

Pranav-error:
The first run failed, which is the check doing its job — and it caught exactly the drift I flagged in the description.

`ubuntu-latest` ships a different cppcheck than the 2.21 I generated the baseline with, and it reports nine findings mine did not:

```
nullPointerRedundantCheck:src/libpgmoneta/info.c: 0 in the baseline, 7 now
nullPointerRedundantCheck:src/libpgmoneta/wf_restore.c: 0 in the baseline, 1 now
nullPointerRedundantCheck:src/libpgmoneta/wf_sha512.c: 0 in the baseline, 1 now
```

Those are now in the baseline, so the job reflects what CI actually sees rather than what my laptop sees.

It also reported `toomanyconfigs:nofile`. That one is cppcheck describing its own configuration limit, not a finding in the tree, so the filter now requires a path under `src/` and ignores anything attributed to the pseudo-file `nofile`.

The job now also prints `cppcheck --version`, so if the runner image bumps the version and the baseline drifts again, the log says so immediately instead of looking like a code regression.

Two things worth deciding, and I am happy either way:

1. **Pinning the version.** Baselining against whatever `apt` currently offers means a future image bump will fail the job for reasons unrelated to any PR. Pinning cppcheck would make it deterministic at the cost of pinning maintenance. I have left it unpinned for now since that matches how `clang-format` is handled here, but that one is pinned to 21 explicitly, so say the word if you want the same treatment.
2. **The nine `nullPointerRedundantCheck` findings** are now recorded rather than fixed. They are in `info.c`, `wf_restore.c` and `wf_sha512.c` and look like genuine redundant NULL checks. I can send those as a separate PR under the sweep you asked for, which would shrink the baseline again.

Unrelated, in case it is useful: `build-linux (clang, Debug, 15)` is timing out after 45 minutes on `main` as well as on my branches — details on #1246.

jesperpedersen:
@Pranav-error You need to follow project guide lines - create an issue and reference it in the squashed commit message

Pranav-error:
Understood, and sorry for the omission — CONTRIBUTING.md is clear on it and I stopped following it partway through.

Filed #1270 for this one and squashed to a single commit, `[#1270] Add a cppcheck job to CI, gated on a baseline`.

I had the same gap on my other six open PRs, so I fixed those too rather than leaving you to point it out again — an issue each and one squashed commit referencing it:

- #1256 → #1271
- #1260 → #1272
- #1266 → #1273
- #1267 → #1274
- #1268 → #1275
- #1269 → #1276

Diffs are unchanged in all seven; only the commit messages moved.

Pranav-error:
`build-linux (clang, Debug, 15)` is red here and I do not think it can be from this change, but I would rather lay out the reasoning than just assert it.

The job did not fail a test — it hung. The last line is `test_cli_verify ... [PASS]` at 18:11:18, then nothing until `The action 'Run Tests' has timed out after 45 minutes`. Every test that ran, passed.

This PR adds one new independent job (`static-analysis`) and one new file (`.github/cppcheck-baseline.txt`). It changes no source, and nothing in the `build-linux` job reads either. The other five `build-linux` variants on the same commit are green, as are `storage-engines`, `format-check` and the new job itself — a real defect would not single out clang-15.

I cannot re-run jobs here. Say the word and I will push an empty commit to retrigger it.

Pranav-error:
Adding to my note above, since the picture is clearer now: four of my PRs went red in the same window and three of them — #1256, #1267 and this one — are the identical `The action 'Run Tests' has timed out after 45 minutes` failure, on two different clang versions and at different points in the suite. In every case each test that ran passed; the step hung.

For #1256 and #1267 the trees are byte-identical to their last green runs (same tree hash, only the commit message changed), so the same content was green and is now red. This PR changes no source at all.

All seven were pushed within about four minutes of each other, so seven full CI runs were competing for runners at once. That may well be the whole story.

Sorry for the noise — that was my doing, from applying the `[#issue]` convention to all seven at once.

PR #1252, opened 2026-08-21
Validate the hex and xlogpos conversions instead of trusting sscanf

Pranav-error:
Follow-up to #1251, which flagged these four `invalidScanfArgType_int` findings but left them out — neither is really an argument-type problem.

## `server.c` — the failure check never fired

```c
int hi, lo;
sscanf(hex_chr_str, "%x", &hi);
...
if (hi == -1 || lo == -1)
```

`sscanf` never writes `-1`; on a failed conversion it returns 0 and leaves the variable untouched. So an invalid hex character was never detected — the loop silently reused the previous iteration's value — and on the first iteration `hi` and `lo` are uninitialised, so the comparison read indeterminate memory. Now checks each return value and initialises both.

Making them `unsigned` without this would have been worse: the `== -1` test would become always-false, which reads as intentional.

## `wal.c` — `strtok` result used without a NULL check

The guard above only proves a `/` is present. An `xlogpos` of `"ABC/"` satisfies it, and then the second `strtok` returns NULL and `sscanf` dereferences it. Both pointers and both conversions are now checked, failing the way the function already fails on bad input. The offset mask is now `~((unsigned int)segsize - 1)` so the operand types match after `num` became unsigned.

## Verification

`cppcheck` reported `invalidScanfArgType_int` four times and reports none after. Builds clean, no `clang-format` changes. Verified by reading, not by driving a malformed `xlogpos` through at runtime.

### DIFF
--- a/src/libpgmoneta/wal.c
+++ b/src/libpgmoneta/wal.c
@@ -1158,21 +1158,17 @@ wal_xlog_offset(size_t xlogptr, int segsize)
 static int
 wal_convert_xlogpos(char* xlogpos, int segsize, uint32_t* high32, uint32_t* low32)
 {
-   char* ptr = NULL;
-   int num = 0;
-   if (xlogpos == NULL || !pgmoneta_contains(xlogpos, "/"))
+   uint64_t lsn = 0;
+
+   if (pgmoneta_string_to_lsn(xlogpos, &lsn))
    {
       pgmoneta_log_error("WAL unable to convert xlogpos");
       return 1;
    }
-   ptr = strtok(xlogpos, "/");
-   sscanf(ptr, "%x", &num);
-   *high32 = num;
 
-   ptr = strtok(NULL, "/");
-   sscanf(ptr, "%x", &num);
+   *high32 = (uint32_t)(lsn >> 32);
    // discard in-segment offset
-   *low32 = num & (~(segsize - 1));
+   *low32 = (uint32_t)lsn & ~((uint32_t)segsize - 1);
    return 0;
 }
 
--- a/src/include/utils.h
+++ b/src/include/utils.h
@@ -1350,10 +1350,11 @@ pgmoneta_lsn_to_string(uint64_t lsn);
 /**
  * Generate the lsn integer given a lsn string format %X/%X
  * @param lsn string value
- * @return The lsn integer
+ * @param lsn_out [out] The lsn integer, set to 0 if lsn could not be parsed
+ * @return 0 upon success, otherwise 1
  */
-uint64_t
-pgmoneta_string_to_lsn(char* lsn);
+int
+pgmoneta_string_to_lsn(char* lsn, uint64_t* lsn_out);
 
 /**
  * Check if the path to a file is an incremental path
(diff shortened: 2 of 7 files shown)


Amr-Shams (approved):

src/libpgmoneta/wal.c
```diff
@@ -1159,20 +1159,30 @@ static int
 wal_convert_xlogpos(char* xlogpos, int segsize, uint32_t* high32, uint32_t* low32)
```
Amr-Shams, line 1159: there is a func in utils.c pgmoneta_string_to_lsn, i know it is not safe(it uses scanf and same error) so fixing this func and reuse it here would be the right call

Pranav-error: Agreed that `pgmoneta_string_to_lsn` has the same bug — `hi` and `lo` are pre-zeroed and the `sscanf` return is unchecked, so `"ABC/"` silently yields a partial LSN instead of an error.

It is not a drop-in replacement here though. It returns a combined `uint64_t` with no error channel, and 0 is indistinguishable from a valid `0/0`, whereas `wal_convert_xlogpos` has to be able to fail. Giving it an error channel means changing the signature at 11 call sites across `walinfo.c`, `wf_backup_incremental.c` and `server.c`, all of which currently treat a parse failure as 0 — so that change would quietly alter behaviour at each of them.

I would rather not fold an API change that size into a PR whose point is the NULL dereference. Happy to do it as a follow-up: fix `pgmoneta_string_to_lsn`, update the call sites deliberately, then reuse it here and drop the local parsing. Shall I open an issue for that?

jesperpedersen: @Pranav-error You can do it as part of this PR

Pranav-error: Done in this PR.

`pgmoneta_string_to_lsn` now returns 0 or 1 and takes the value as an out parameter, set to 0 when the string cannot be parsed, and `wal_convert_xlogpos` uses it instead of parsing the string itself.

For the 11 call sites: `server.c` and `wf_backup_incremental.c` treat a failure as an error, since a wrong LSN there affects the backup. The `walinfo.c` ones keep the existing behaviour, where an unparsable bound simply does not filter — except that the filter is now only marked present when the input actually parsed, so a malformed `--end-lsn` no longer filters on 0 and silently returns nothing.

It also fixed something I had not spotted: `wal_convert_xlogpos` called `strtok` on the string, and two of its three callers pass `pgmoneta_query_response_get_data()` in directly, so the parse was writing a NUL over the `/` in the response buffer. The `sscanf` version leaves it alone.

Builds clean, `clang-format` clean, and the four `invalidScanfArgType_int` findings are gone. Squashed to one commit.

Amr-Shams (approved):

Amr-Shams:
@Pranav-error sign the commit

Pranav-error:
Signed off and squeezed the message to a single line, same as #1253. Force-pushed as `de67a01`; the diff is unchanged.

jesperpedersen:
Rebased and merged.

PR #1256, opened 2026-08-22
Remove the inner checks that cannot be false

Pranav-error:
`ssh_storage_backup_execute()` and `link_execute()` both walk the backup list with `next_newest == -1` **in the for condition**:

```c
for (int j = number_of_backups - 2; j >= 0 && next_newest == -1; j--)
{
   if (pgmoneta_is_backup_struct_valid(server, backups[j]))
   {
      if (next_newest == -1)   /* always true when reached */
      {
         next_newest = j;
      }
   }
}
```

The inner test can only be reached while the loop condition still holds, so it is always true; and once it assigns, the loop condition ends the loop before the next body runs. Removing it keeps the same first-match behaviour.

cppcheck 2.21.0 reports these as `identicalInnerCondition` and reports neither after. No behaviour change, so nothing to test beyond a clean build; `clang-format` reports no changes.

Dropped the `uselessAssignmentPtrArg` half of the original PR per @jesperpedersen — `p = NULL;` after `free(p)` is a house convention here, so those three lines are restored.

### DIFF
--- a/src/libpgmoneta/se_ssh.c
+++ b/src/libpgmoneta/se_ssh.c
@@ -328,10 +328,7 @@ ssh_storage_backup_execute(char* name __attribute__((unused)), struct art* nodes
       {
          if (pgmoneta_is_backup_struct_valid(server, backups[j]))
          {
-            if (next_newest == -1)
-            {
-               next_newest = j;
-            }
+            next_newest = j;
          }
       }
    }
--- a/src/libpgmoneta/wf_link.c
+++ b/src/libpgmoneta/wf_link.c
@@ -142,10 +142,7 @@ link_execute(char* name __attribute__((unused)), struct art* nodes)
          if (pgmoneta_is_backup_struct_valid(server, backups[j]) &&
              backups[j]->major_version == backups[number_of_backups - 1]->major_version)
          {
-            if (next_newest == -1)
-            {
-               next_newest = j;
-            }
+            next_newest = j;
          }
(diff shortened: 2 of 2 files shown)


jesperpedersen:
@Pranav-error `p = NULL;` is a house convention

Pranav-error:
Understood — restored all three `p = NULL;` lines and dropped that half of the PR. It is now only the two `identicalInnerCondition` cleanups in `se_ssh.c` and `wf_link.c`, where the same test is already in the enclosing `for` condition. Retitled to match, force-pushed as `24f184c`.

Pranav-error:
`build-linux (clang, Debug, 14)` is red, and it is not from this change.

It did not fail a test — the step hung. Every test that ran passed; the log stops after `test_pgmoneta_delete_force_retained_backup_rejects_regular_delete [PASS]` and then `The action 'Run Tests' has timed out after 45 minutes`.

The tree here is byte-identical to the one that passed this job before — `24f184c` and `ce54ee3` have the same tree hash. The only difference is the commit message, which now carries the `[#1271]` reference you asked for. Same content, green then red.

Three of my other PRs hit the same 45-minute `Run Tests` timeout in the same window, at different points in the suite and on different clang versions, including one that changes no source at all. I cannot re-run jobs here — happy to push an empty commit to retrigger if that is easier.

jesperpedersen:
Rebased and merged.

PR #1258, opened 2026-08-24
[#1257] Fix infinite loop in pgmoneta_remove_prefix when the input does not start with the prefix

### LINKED ISSUE #1257: pgmoneta_remove_prefix() spins forever when the input does not start with the prefix (no mismatch handling in scan loop)
## Summary

`pgmoneta_remove_prefix()` in `src/libpgmoneta/utils.c` contains an unbounded scan loop that **never terminates when the input string does not start with the given prefix**: the loop only advances its index on a matching character and has no `else` branch, so the first mismatching character spins the loop forever (100% CPU).

Static-analysis finding against current `master`; not executed here.

## Location

- File: `src/libpgmoneta/utils.c`
- Function: `pgmoneta_remove_prefix(char* orig, char* prefix)` (~lines 1695-1727)

```c
while (idx < len1 && idx < len2)
{
    if (orig[idx] == prefix[idx])
    {
        idx++;
    }
    // no else: on a mismatch idx never changes -> infinite loop
}
```

## Problem

The loop is meant to find the length of the common prefix of `orig` and `prefix`. However, when `orig[idx] != prefix[idx]`, neither `idx` nor any other state changes and the loop condition remains true, so execution hangs indefinitely. The function is exported via `src/include/utils.h` and is part of the general string utility API.

For inputs where `orig` genuinely starts with `prefix` (or one is empty), the loop exits through the bounds check and the result is correct - which is why existing call sites and tests (`test/testcases/test_utils.c` line ~443) do not catch it.

## Trigger / Reproduction

Based on static analysis; no runtime run was performed:

```c
pgmoneta_remove_prefix("hello", "world");   // hangs forever
pgmoneta_remove_prefix("data/x", "dat0/");  // hangs forever ('a' vs '0' at index 3)
```

Any pair of non-empty strings whose first `min(len1,len2)` characters diverge before the shorter one ends triggers the hang.

Current in-tree callers happen to be safe by construction:
- `src/libpgmoneta/link.c:89` - `from_entry` is built by concatenating `base_from` with directory entries, so it always starts with `base_from`.
- `src/libpgmoneta/restore.c:343` - guarded by `pgmoneta_starts_with(normalized, "data/")`.

But nothing prevents future callers (the utility is public API) from passing arbitrary strings.

## Expected Behavior

On mismatch the function should stop scanning at the divergence point. Reasonable semantics: return a copy of `orig` unchanged when it does not start with `prefix` (mirroring how `remove_suffix` returns the original string when `ends_with` fails), or return an empty string - either way, terminate.

## Actual Behavior

Infinite loop with no exit condition on the first differing byte; the process must be killed externally.

## Impact

A single call with unexpected input freezes the whole pgmoneta process (main server process or worker thread). Since the helper sits in the shared utility layer used by backup linking and restore workflows, any future call site that does not strictly guarantee the prefix relationship will introduce an unrecoverable hang.

## Suggested Direction

Break out of the loop on the first mismatch and treat "does not start with prefix" as "return copy of orig" (or document+assert the precondition). E.g.:

```c
while (idx < len1 && idx < len2 && orig[idx] == prefix[idx])
{
    idx++;
}
```

and keep the rest of the allocation logic as-is. Also consider checking the `malloc` return value in the same function while touching it.

## Evidence

- `utils.c` ~lines 1712-1718: loop body increments `idx` only inside the equality branch.
- `test/testcases/test_utils.c`: only the matching-prefix case ("pre_test"/"pre_") is exercised.
- Exported in `src/include/utils.h`; consumed by `link.c` and `restore.c` under conditions that currently guarantee the prefix property.

krishna3554:
## Summary

`pgmoneta_remove_prefix()` in `src/libpgmoneta/utils.c` contained an unbounded scan loop that never terminated when the input string did not start with the given prefix: the loop only advanced its index inside the equality branch, so the first mismatching character spun the loop forever (100% CPU).

Fixes #1257

## Changes

- Move the equality check into the loop condition (`while (idx < len1 && idx < len2 && orig[idx] == prefix[idx])`) so scanning stops at the first divergence point.
- If `orig` does not start with `prefix` (`idx < len2` after the loop), return a copy of `orig` unchanged, mirroring how `pgmoneta_remove_suffix()` returns the original string when `pgmoneta_ends_with()` fails.
- Compute `strlen(orig)`/`strlen(prefix)` only after the `NULL` checks (previously `strlen(NULL)` was evaluated before the guard, which is undefined behavior).
- Check the `malloc` return value in the same function, as suggested in the issue.
- Add regression tests to `test/testcases/test_utils.c`: full mismatch (`"hello""/"world"`) and partial common prefix (`"data/x"`/`"dat0/"`).

## Behavior

| Input | Before | After |
| --- | --- | --- |
| `(\"hello\", \"world\")` | hangs forever | returns copy of `orig` |
| `(\"data/x\", \"dat0/\")` | hangs forever | returns copy of `orig` |
| `(\"pre_test\", \"pre_\")` | `test` | `test` (unchanged) |
| `(\"abc\", \"\")`, `(x, NULL)` | copy of `orig` | copy of `orig` (unchanged) |

Current in-tree callers are safe by construction and keep their existing behavior:
- `src/libpgmoneta/link.c:89`
- `src/libpgmoneta/restore.c:343`

## Verification

- Standalone harness comparing old vs new implementation confirms the old loop spins on a mismatch while the new code terminates and returns the expected values for mismatch, partial-match, empty-prefix, shorter-orig, and NULL cases.
- Regression tests added next to the existing `remove_prefix` case in `MCTF_TEST(test_utils_string_manipulation)`.
- Full build not run locally due to missing build dependencies (cmake/libev); changes are limited to this function plus tests.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -58,3 +58,4 @@ Ahmed Kamal <ahmedkamal200427@gmail.com>
 Rohan Mishra <kmrrohan29@gmail.com>
 Mostafa Mahmoud <mm1471800@gmail.com>
 R Sai Pranav <rajasaipranav0@gmail.com>
+Krishna Lokhande <krishlokhande45@gmail.com>
--- a/doc/manual/en/97-acknowledgement.md
+++ b/doc/manual/en/97-acknowledgement.md
@@ -65,6 +65,7 @@ Ahmed Kamal <ahmedkamal200427@gmail.com>
 Rohan Mishra <kmrrohan29@gmail.com>
 Mostafa Mahmoud <mm1471800@gmail.com>
 R Sai Pranav <rajasaipranav0@gmail.com>
+Krishna Lokhande <krishlokhande45@gmail.com> 
 ```
 
 ## Committers
--- a/src/libpgmoneta/utils.c
+++ b/src/libpgmoneta/utils.c
@@ -1696,8 +1696,8 @@ pgmoneta_remove_prefix(char* orig, char* prefix)
 {
    char* res = NULL;
    int idx = 0;
-   int len1 = strlen(orig);
-   int len2 = strlen(prefix);
+   int len1 = 0;
+   int len2 = 0;
    int len = 0;
    if (orig == NULL)
    {
@@ -1709,15 +1709,24 @@ pgmoneta_remove_prefix(char* orig, char* prefix)
       res = pgmoneta_append(res, orig);
       return res;
    }
-   while (idx < len1 && idx < len2)
+   len1 = strlen(orig);
+   len2 = strlen(prefix);
+   while (idx < len1 && idx < len2 && orig[idx] == prefix[idx])
    {
-      if (orig[idx] == prefix[idx])
-      {
-         idx++;
-      }
+      idx++;
+   }
+   if (idx < len2)
+   {
+      // does not start with the prefix
+      res = pgmoneta_append(res, orig);
+      return res;
    }
    len = len1 - idx + 1;
    res = malloc(len);
+   if (res == NULL)
+   {
+      return NULL;
+   }
    res[len - 1] = 0;
    if (len > 1)
    {
--- a/test/testcases/test_utils.c
+++ b/test/testcases/test_utils.c
@@ -448,6 +448,30 @@ MCTF_TEST(test_utils_string_manipulation)
    free(res);
    res = NULL;
 
+   // test remove_prefix without a matching prefix
+
+   s = strdup("hello");
+   MCTF_ASSERT_PTR_NONNULL(s, cleanup, "strdup failed");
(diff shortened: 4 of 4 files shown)


jesperpedersen:
@krishna3554 You need to add your name to the author files - see DEVELOPERS.md

jesperpedersen:
@krishna3554 Start to cross-port to pgagroal, pgexporter and pgvictoria

krishna3554 (commented):
Done with adding my name

jesperpedersen:
@krishna3554 You need to squash - read DEVELOPERS.md and have orangu help you

Issue #1042, opened 2026-03-20
I/O: Buffer size


jesperpedersen:
Investigate using different buffer sizes for all disk I/O.

It would mean using

```
int setvbuf(FILE *restrict stream, char buf[restrict .size],
            int mode, size_t size);
```

Lower amount would be from `st_blksize` (`stat()` call).

Unlikely to lead to a performance improvement - just documentation explaining why we don't support disk buffer size

Sakshii-27:
@jesperpedersen can I work on this?

jesperpedersen:
@Sakshii-27 Go for it !

jesperpedersen:
@Sakshii-27 I would just make a standalone binary that tests read/write of 1Gb files with different buffer sizes. See if you can test with different disk types / file systems

Sakshii-27:
@jesperpedersen I've investigated using different buffer sizes for disk I/O as requested. I created a standalone benchmark tool to test `setvbuf()` performance variations.

I tested a **1GB file** read/write with 13 different buffer sizes on a Mac (SSD/APFS), running with `sudo` to drop filesystem caches between tests.

### Benchmark Results (1GB File)

<img width="690" height="580" alt="Image" src="https://github.com/user-attachments/assets/a84f0ca0-fed6-4aae-9d49-262fe634ea11" />

<img width="577" height="516" alt="Image" src="https://github.com/user-attachments/assets/81f77a76-2bd7-42e4-abdd-dfdd32e9650b" />

### Investigation Summary
- **No clear advantage**: While there are minor fluctuations (e.g., 16KB was faster in this run), there isn't a consistent or significant performance gain from increasing the Userspace buffer size.
- **Kernel Efficiency**: Modern OS kernels (like macOS and Linux) and high-speed SSDs already use sophisticated caching, read-ahead, and write-behind mechanisms that outweigh the effects of `setvbuf()`.
- **Conclusion**: This benchmark confirms that adding complexity to support custom disk buffer sizes is unlikely to improve `pgmoneta`'s real-world performance.

I have implemented the benchmark tool in a new `contrib/benchmark_buffer` directory. Let me know if you would like me to submit a PR to include this tool in the repository!

Purushottam-y:
Hi, I want to work on this issue, Please assign to me.

jesperpedersen:
@Purushottam-y Go for it !

Purushottam-y:
[...]

PR #1262, opened 2026-08-29
Add I/O buffer size benchmark for issue #1042

Purushottam-y:
Implemented an I/O buffer size benchmark for issue #1042.

The benchmark:
- Tests 1 GB file read/write performance
- Tests buffer sizes from 4 KB to 16 MB
- Records write and read times
- Includes three benchmark runs for comparison


jesperpedersen:
@Purushottam-y You can keep these local - so I'll close

PR #1265, opened 2026-08-30
Tests: Add replica backup/restore test cases

### LINKED ISSUE #1214: Test suite: Replica backup/restore
We need to create a full test suite that will test

* Test case: Full backup from a replica (PostgreSQL 17+)
* Test case: Incremental backup from a replica (PostgreSQL 17+)
* Test case: Full backup from a replica (PostgreSQL 14)
* Test case: Incremental backup from a replica (PostgreSQL 14)

So, we need to configure a primary/replica setup, and make sure that backups can be taken from the replica instance instead of the primary one.

Then we need to do the same for a restore based on these backups

AhmedMabrouk22:
Closes #1214 
Hi @jesperpedersen,
I sincerely apologize for the long delay. I had some unexpected hardware issues and got heavily tied up with some commitments.
I have already started exploring the issue and made some initial progress. I added the replica setup and configuration in the test environment, running the primary and replica in two separate containers, currently not running in CI mode. I also wrote a test for a full backup from the replica.

when I try to run an incremental backup, I face the following error `(SQLSTATE code: 55000)` with message : `manifest requires WAL from final timeline 1 ending at 0/130000D8, but this backup starts at 0/13000060`

I am currently researching this error to get more details. Could you point me to any resources or specific parts of the codebase that might help me resolve this?

Thanks!


AhmedMabrouk22:
While searching about the incremental backup issue, I found in the postgresql doc that this is not a bug, it's the limitation when work with replica and it's normal.
so in incremental backup test i took the full backup from replica, inserted rows on the primary and execute checkpoint on both primary and replica

AhmedMabrouk22:
@jesperpedersen 
I wrote full and incremental backup tests from replica for postgresql 14 and 17. They work correctly, but i skipped the incremental backup test in postgresql 14 because when i run the incremental backup manually outside the tests i got this error `ERROR wf_backup_incremental.c:356 Incremental backup: Error during backup of: backup_manifest`
and this error occurred when `write_full_file()` is called. so i think this is bug when run incremental backup from replica in postgresql 14 so i skipped this test for now.

jesperpedersen:
@AhmedMabrouk22 See if you can find the root cause

AhmedMabrouk22:
> @AhmedMabrouk22 See if you can find the root cause
@jesperpedersen 

The problem occurs when pgmoneta tries to read the `backup_manifest` file and because the data is in hex format, size is doubled. and also the buffer size is small, the file not read fully and throw the error: `server.c:1581 invalid hex bytea (partial data)`
i found this only happens when taking an incremental backup from a replica. It works correctly on the primary because this file does not exist there; it is only generated when the replica is created using `pg_basebackup`

My initial solution is skip this file, but I faced a new issue: the backup hangs and takes a very long time. I think it enters an infinite loop, but I am not sure if it's a new bug or related to skipping the file. I will investigate this in more detail

jesperpedersen:
@AhmedMabrouk22 You need to rebase, squash and reference the issue in the commit message - see DEVELOPERS.md

AhmedMabrouk22:
> @AhmedMabrouk22 You need to rebase, squash and reference the issue in the commit message - see DEVELOPERS.md

done

AhmedMabrouk22:
> > @AhmedMabrouk22 See if you can find the root cause
> > @jesperpedersen
> 
> The problem occurs when pgmoneta tries to read the `backup_manifest` file and because the data is in hex format, size is doubled. and also the buffer size is small, the file not read fully and throw the error: `server.c:1581 invalid hex bytea (partial data)` i found this only happens when taking an incremental backup from a replica. It works correctly on the primary because this file does not exist there; it is only generated when the replica is created using `pg_basebackup`
> 
> My initial solution is skip this file, but I faced a new issue: the backup hangs and takes a very long time. I think it enters an infinite loop, but I am not sure if it's a new bug or related to skipping the file. I will investigate this in more detail

@jesperpedersen 
i found the incremental backup enters in infinity loop in the `wait_for_wal_switch` function. and the root cause is when taking the incremental backup from primary, function `pg_stop_backup()` forces a wal switch automatically , so the current wal closing. but in replica this function cannot trigger a wal switch and the file remains open

my solution is first execute `checkpoint` in primary, then start backup and after `pg_stop_backup()` execute `pg_switch_wal()` in primary to force switch wal file

PR #1266, opened 2026-08-31
Check the allocations in the block reference table

Pranav-error:
`brt.c` builds the block reference table used for incremental backups. Nine allocations in it are used without being checked, and every one is followed immediately by a write through the returned pointer.

`brt_mark_block_modified()` has eight of them:

```c
entry->chunk_size = (uint16_t*)realloc(entry->chunk_size, sizeof(uint16_t) * max_chunks);
memset(&entry->chunk_size[entry->nchunks], 0, extra_chunks * sizeof(uint16_t));
```

On failure that overwrites the only pointer to the existing array — leaking it — and then `memset`s through NULL. The `malloc` branch above it, the first chunk allocation, the bitmap conversion and the array doubling all have the same shape. The doubling one also sets `entry->chunk_size[chunkno] = newsize` after a failed `realloc`, so the recorded size would describe memory that was never allocated.

`pgmoneta_brt_create_empty()` has the ninth: `malloc` then `brtab->table` without a check.

## How failures are reported

`brt_mark_block_modified()` was `static void`, but its only caller `pgmoneta_brt_mark_block_modified()` already returns `int` with a `goto error` path, and *its* callers already test it — `brt.c:441` does `if (pgmoneta_brt_mark_block_modified(...))`. So the channel already existed; this just uses it. The helper becomes `static int`, and nothing outside `brt.c` changes.

In the grow branch each `realloc` result is stored as soon as it succeeds, so `entry` never holds a pointer that `realloc` has already freed, and `nchunks` is raised only after all three have grown. A failure there leaves the entry consistent, just not yet enlarged.

## Verification

`clang -fsyntax-only -I src/include` is clean — it caught a missing `return 0` on the way — and `clang-format` reports no changes. I have not driven a real allocation failure through an incremental backup, so this is verified by reading plus the compiler, not at runtime.

## One I left alone deliberately

`brt_read_next_relation()` at `brt.c:841` allocates `reader->chunk_size` unchecked and passes it straight to `brt_read()`. I did not touch it because that function reports through `bool`, where `false` already means "no more relations" — returning `false` on an allocation failure would make an out-of-memory look like a complete read, which for a backup manifest seems worse than the crash. That needs a decision about the semantics rather than a guard, so I would rather ask: would you prefer a separate error channel there, or is `false` acceptable?

### DIFF
--- a/src/libpgmoneta/brt.c
+++ b/src/libpgmoneta/brt.c
@@ -41,7 +41,7 @@ static int brt_insert(block_ref_table* brt, block_ref_table_key key, block_ref_t
 static block_ref_table_entry* brt_lookup(block_ref_table* brt, block_ref_table_key key);
 
 static void brt_set_limit_block(block_ref_table_entry* entry, block_number limit_block);
-static void brt_mark_block_modified(block_ref_table_entry* entry, block_number blocknum);
+static int brt_mark_block_modified(block_ref_table_entry* entry, block_number blocknum);
 
 static void brt_write(FILE* f, block_ref_table_buffer* buffer, void* data, int length);
 static int brt_read(FILE* f, struct block_ref_table_reader* reader, void* data, int length);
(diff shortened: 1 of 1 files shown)


jesperpedersen:
Rebased and merged.

PR #1267, opened 2026-08-31
Check the allocations that grow the YAML rule arrays

Pranav-error:
`handle_scalar_event()` grows two arrays without checking the results:

```c
config->operations = realloc(config->operations, (config->operation_count + 1) * sizeof(char*));
config->operations[config->operation_count] = strdup(value);
config->operation_count++;
```

On failure `realloc` returns NULL while leaving the old array allocated, so overwriting `config->operations` leaks everything parsed so far and the next line writes through the NULL. The `xids` case a few lines below has the same shape, and the `strdup` result was unchecked too.

Same class as #1266 and the merged #1246.

## How the failure is reported

`handle_scalar_event()` was `void`, but the parse loop it runs in belongs to `pgmoneta_parse_yaml_config()`, which already returns `int`, and its caller already tests that — `walfilter.c:576` does `if (pgmoneta_parse_yaml_config(yaml_file, &yaml_config))`. So the channel existed; this uses it. The handler now returns non-zero on an allocation failure, the loop stops and the existing cleanup at the end of the function runs before returning -1.

The declaration in `yaml_utils.h` changes from `void` to `int` accordingly. `handle_scalar_event()` has no callers outside `yaml_utils.c`, so nothing else is affected, and ignoring an `int` return is source-compatible in any case.

## Verification

`clang -fsyntax-only -I src/include` is clean and `clang-format` reports no changes. I could not drive a real allocation failure through a walfilter run, so this is verified by reading plus the compiler rather than at runtime.

### DIFF
--- a/src/include/yaml_utils.h
+++ b/src/include/yaml_utils.h
@@ -88,8 +88,8 @@ int pgmoneta_parse_yaml_config(const char* filename, config_t* config);
  * @param current_key Pointer to the current key being processed.
  * @param config Pointer to the configuration structure to populate.
  */
-void handle_scalar_event(yaml_event_t* event, parser_state_t* state,
-                         char** current_key, config_t* config);
+int handle_scalar_event(yaml_event_t* event, parser_state_t* state,
+                        char** current_key, config_t* config);
 
 /**
  * Cleans up the configuration structure.
--- a/src/libpgmoneta/yaml_utils.c
+++ b/src/libpgmoneta/yaml_utils.c
@@ -65,6 +65,7 @@ pgmoneta_parse_yaml_config(const char* filename, config_t* config)
    parser_state_t state = STATE_START;
    char* current_key = NULL;
    int done = 0;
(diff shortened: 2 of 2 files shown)


Pranav-error:
`build-linux (clang, Debug, 15)` is red, and it is not from this change.

The step hung rather than failing a test: every test that ran passed, the log stops after `test_cli_verify [PASS]` (the next case would be `test_cli_archive`), then `The action 'Run Tests' has timed out after 45 minutes`.

The tree here is byte-identical to the one that passed this job before — `b329433` and `b115ea2` have the same tree hash; only the commit message changed, to add the `[#1274]` reference. And this change touches `yaml_utils.c` and its header, which the CLI tests do not exercise.

Three of my other PRs timed out the same way in the same window, on different clang versions and at different points in the suite. I cannot re-run jobs here — happy to push an empty commit if that helps.

jesperpedersen:
Rebased and merged.

PR #1268, opened 2026-09-01
Bound the directory listing in the walinfo browser

Pranav-error:
`show_wal_file_selector()` in `walinfo.c` collects a directory listing into fixed stack arrays and never checks the indices:

```c
char temp_dirs[999][256];
char temp_files[999][256];
int temp_dir_idx = 0;
int temp_file_idx = 0;

{
   ...
   strcpy(temp_dirs[temp_dir_idx++], entry->d_name);
   ...
   strcpy(temp_files[temp_file_idx++], entry->d_name);
}
```

Neither index is bounded, so a directory with more than 999 entries writes past the end of a stack array. The names themselves are fine — `d_name` is at most 255 bytes and the rows are 256 — it is the row count that is unchecked.

This is reachable in ordinary use rather than as a corner case: the files collected here are the ones whose first 24 characters are hex, which is exactly the WAL segment naming scheme, and a WAL archive of any age holds far more than 999 segments.

There is a third write with the same problem. `entries` holds 1000 rows, but the two copy loops after the scan can move up to 999 directories **and** 999 files into it, plus the `".."` entry added first:

```c
for (int i = 0; i < temp_dir_idx; i++)
{
   strcpy(entries[dir_count].name, temp_dirs[i]);
   ...
}
for (int i = 0; i < temp_file_idx; i++)
{
   strcpy(entries[dir_count + file_count].name, temp_files[i]);
   ...
}
```

So even with the first two bounded, the total can still reach 1999 rows in a 1000-row array.

This bounds all three. A listing larger than the arrays is truncated rather than overflowing; I kept the existing arrays and behaviour rather than moving to a dynamic listing, since that is a bigger change and you may prefer a different shape for it. Happy to do that instead if you would rather the browser showed everything.

## Verification

`clang -fsyntax-only -I src/include` is clean, and `clang-format` reports no changes — I checked that `walinfo.c` is format-clean upstream first, so this is not adding drift. I have not run the browser against a WAL directory of more than 999 segments, so the overflow is established by reading the bounds rather than by triggering it.

### DIFF
--- a/src/walinfo.c
+++ b/src/walinfo.c
@@ -3983,7 +3983,8 @@ show_wal_file_selector(struct ui_state* state)
       int dir_count = 0;
       int file_count = 0;
 
-      strcpy(entries[dir_count].name, "..");
+      pgmoneta_snprintf(entries[dir_count].name, sizeof(entries[dir_count].name),
+                        "%s", "..");
       entries[dir_count].is_dir = true;
       dir_count++;
 
@@ -4008,7 +4009,13 @@ show_wal_file_selector(struct ui_state* state)
          {
             if (S_ISDIR(st.st_mode))
             {
-               strcpy(temp_dirs[temp_dir_idx++], entry->d_name);
+               if (temp_dir_idx < (int)(sizeof(temp_dirs) / sizeof(temp_dirs[0])))
+               {
+                  pgmoneta_snprintf(temp_dirs[temp_dir_idx],
+                                    sizeof(temp_dirs[temp_dir_idx]), "%s",
+                                    entry->d_name);
+                  temp_dir_idx++;
+               }
             }
             else if (S_ISREG(st.st_mode))
             {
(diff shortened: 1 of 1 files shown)


jesperpedersen:
@Pranav-error This looks wrong, and we don't use `strcpy`
@saraanbih Can you help with this ?

Pranav-error:
Fixed — the branch now removes all 11 `strcpy` calls in that function and uses `pgmoneta_snprintf` instead, which is what the rest of `walinfo.c` already uses. The diff adds no `strcpy` and removes 11.

To be clear about what the first version did, since the diff made it look worse than it was: those `strcpy` lines were already there, and I had only wrapped two of them in bounds checks, which re-indented them and made them show up as added lines. That was a bad way to present it, and switching them to `pgmoneta_snprintf` is the better change anyway.

On "this looks wrong" — if you mean the truncation, I would agree it is the weaker half. Silently dropping entries past 999 keeps the browser from overflowing the stack, but it also means a WAL directory with more segments than that cannot be fully browsed, which for `walinfo` is the normal case rather than the exception. The arrays are `temp_dirs[999][256]`, `temp_files[999][256]` and `entries[1000]`, so a real archive exceeds them easily.

The straightforward version is to count the entries first, allocate, and drop the fixed arrays entirely. I did not do that in this PR because it changes more of the function than a bounds fix and I would rather agree the shape first. Happy to send that instead if it is what you want — or if @saraanbih is already on it, I am glad to leave it with her and close this.

jesperpedersen:
Rebased and merged.

Issue #1143, opened 2026-05-14
Zombie processes are left on SIGINT


ahmed-kamal2004:
- Before killing pgmoneta

```
ahmed-k+  298420 35.2  3.6 21474987912 514352 pts/4 S 11:06   0:45 pgmoneta: verification
ahmed-k+  303202 34.3  3.4 21474987912 487160 pts/4 S 11:07   0:14 pgmoneta: verification
ahmed-k+  304595  7.9  3.6 21474986888 522264 pts/4 S+ 11:07   0:01 pgmoneta: main
ahmed-k+  304752  0.3  0.0      0     0 pts/4    Z+   11:07   0:00 [pgmoneta] <defunct>
ahmed-k+  304756 37.1  3.3 21474987912 479340 pts/4 R+ 11:07   0:08 pgmoneta: verification
ahmed-k+  306969  300  0.1  33836 26564 pts/4    R    11:08   0:00 addr2line -e /usr/local/lib/libpgmoneta.so.0.21.0 -fC 0x4887c5
ahmed-k+  306972  0.0  0.0   9380  2324 pts/1    S+   11:08   0:00 grep --color=auto --exclude-dir=.bzr --exclude-dir=CVS --exclude-dir=.git --exclude-dir=.hg --exclude-dir=.svn --exclude-dir=.idea --exclude-dir=.tox --exclude-dir=.venv --exclude-dir=venv pgmoneta
ahmed-k+  306973  0.0  0.1  22812 15580 pts/4    R    11:08   0:00 addr2line -e /usr/local/lib/libpgmoneta.so.0.21.0 -fC 0x30fcc1
```

- After killing pgmoneta using `SIGINT`  ->  `CTLR + C`

```
ahmed-k+  298420 35.2  3.6 21474987912 516100 pts/4 R 11:06   0:47 pgmoneta: verification
ahmed-k+  303202 34.4  3.4 21474987912 489232 pts/4 S 11:07   0:17 pgmoneta: verification
ahmed-k+  304756 36.6  3.4 21474987912 481428 pts/4 R 11:07   0:10 pgmoneta: verification
[...]

PR #1269, opened 2026-09-03
Reap exited children when the event loop stops

Pranav-error:
Relates to #1143.

`sigchld_cb` is an `ev_signal` watcher, so it only runs while the event loop is running. `shutdown_cb` calls `ev_break(loop, EVBREAK_ALL)` straight away, and the loop is gone for the rest of `main()` — through `shutdown_management()`, `shutdown_metrics()`, `shutdown_nagios()`, `shutdown_console()`, `shutdown_mgt()`, `ev_loop_destroy()`, the frees, `remove_pidfile()`, `pgmoneta_stop_logging()` and both `pgmoneta_destroy_shared_memory()` calls. Nothing reaps on that path, so a child exiting during shutdown stays a zombie until the process finally returns.

This drains what has already exited, immediately after the loop stops.

**This is not the fix for #1143**, and I would rather not have it read as one. The `ps` output in that report shows a zombie next to a *running* `pgmoneta: main`, which is a zombie during normal operation — something this does not explain, since `sigchld_cb` should be reaping then. I raised that distinction on the issue on 2026-08-28 along with two questions that would separate the causes, and offered this piece separately; nobody objected, so here it is on its own. Happy to close it if you would rather keep everything in one change once the main cause is known.

## Verification

`clang -fsyntax-only -I src/include` is clean and `clang-format` reports no changes. `sys/wait.h` is already included. I have not reproduced the zombie state, so this is verified by reading the shutdown path rather than at runtime.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1002,6 +1002,15 @@ main(int argc, char** argv)
       ev_loop(main_loop, 0);
    }
 
+   /* sigchld_cb is an event watcher, so it stops running the moment the loop
+    * is broken. Any child exiting during the shutdown sequence below would
+    * therefore stay a zombie for the rest of the process's life. Reap what has
+    * already exited before starting it. */
+   while (waitpid(-1, NULL, WNOHANG) > 0)
+   {
+      ;
+   }
+
    pgmoneta_log_info("Shutdown");
 #ifdef HAVE_SYSTEMD
    sd_notify(0, "STOPPING=1");


Pranav-error:
The point is fair in principle — `waitpid()` can return `-1` with `EINTR` even under `WNOHANG`, and this loop would stop there.

I wrote it this way deliberately though: it is character-for-character the drain that already exists in `sigchld_cb` at `main.c:2483`, and `wal.c:736` uses the same shape. So the concern applies equally to all three, and fixing only the new one would leave the codebase with two different idioms for the same operation.

Happy to handle it — but as a separate change covering all three, so they stay consistent. Something like retrying on `EINTR` and logging anything that is not `ECHILD`. Say the word and I will send that; I did not want to fold a codebase-wide idiom change into a PR that is otherwise four lines.

In practical terms the cost here is small either way: this drain runs immediately before the process tears itself down, and anything still unreaped is reparented to init a moment later. The reason it is worth having at all is the window during the shutdown sequence, not correctness of the final exit.

Pranav-error:
On the Copilot note about `waitpid()` error handling: the loop here is deliberately character-identical to the existing drain in `sigchld_cb` (`main.c:2483`), which uses the same `while (waitpid(-1, NULL, WNOHANG) > 0)` form. Diverging from it would mean two different reaping idioms in the same file for no gain, and `EINTR` is not really reachable on this path — it runs after the loop has stopped, with no blocking wait to interrupt, and `WNOHANG` returns immediately. If you would rather both were hardened, I am happy to change them together in a separate PR so they stay the same shape.

Pranav-error:
`storage-engines` is red on this run, and it is not from this change. The five failures are all `storage backend setup failed` in `test_remote.c`, each timing out at exactly 10.001s before any pgmoneta code under test runs:

```
test_remote_backup_lands_on_all_backends (00:00:10,001) [FAIL] (test_remote.c:128)
test_remote_uploads_run_in_parallel      (00:00:10,001) [FAIL] (test_remote.c:165)
...
```

The tree here is byte-identical to the one that passed the same job at 02:32 today — `bce86cc` and `c4bc23c` have the same tree hash; only the commit message changed, to add the `[#1276]` reference. So the same content passed and then failed with no code difference between the two runs.

I cannot re-run jobs on this repo. Happy to push an empty commit to retrigger if that is easier than a maintainer re-run.

jesperpedersen:
@Userfrom1995 PTAL

Pranav-error:
@Userfrom1995 one thing to save you time before you look: the red `storage-engines` check is a week-old environmental failure, not something in this diff.

All five failures are `storage backend setup failed` in `test_remote.c`, each timing out at exactly 10.001s before any pgmoneta code under test runs. The tree that failed is byte-identical to one that passed the same job earlier the same day — same tree hash, only the commit message differed.

The diff itself is four lines, and it is character-identical to the drain that already exists in `sigchld_cb` at `main.c:2483`.

I can push an empty commit to retrigger CI if a green run would help the review - just say so and I will.

Pranav-error:
Follow-up on the red `storage-engines` check here: the same failure hit `main` on 2026-09-10, taking out three `test_ssh.c` cases with the same "storage backend setup failed". I have filed the broader pattern as #1295.

PR #1282, opened 2026-09-07
Cross-port: Build the number strings with format_and_append

### LINKED ISSUE #1281: append_int and append_ulong truncate the largest value
`pgmoneta_append_int()` and `pgmoneta_append_ulong()` in `utils.c` pass the digit count where `snprintf` wants the buffer size, so each is one byte short and drops the last digit of the largest value of its type:

```c
char number[12];
pgmoneta_snprintf(&number[0], 11, "%d", i);

char number[21];
pgmoneta_snprintf(&number[0], 20, "%lu", l);
```

`INT_MIN` is `-2147483648`, 11 characters plus a terminator, and `ULONG_MAX` is 20 digits plus a terminator, so:

```
append_int(INT_MIN)      size 11 -> "-214748364"            correct: -2147483648
append_ulong(ULONG_MAX)  size 20 -> "1844674407370955161"   correct: 18446744073709551615
```

There is no overflow — `snprintf` truncates safely. The result is a wrong number, not a crash, and the practical reachability is low: the current callers pass ports, versions and counters that will not reach those values. It is a trap for a future caller rather than a live bug.

Cross-port of pgagroal/pgagroal#1027, which @jesperpedersen asked to have carried across. The fix there follows his suggestion of using the `_append_` functions: the three helpers become one-line calls to `pgmoneta_format_and_append()`, which sizes its buffer from `vsnprintf(NULL, 0, ...)` and so cannot truncate at all. That removes the fixed buffers rather than correcting their sizes.

`pgmoneta_append_ullong()` does not exist here, so only the two functions are affected.

Pranav-error:
> **This is a cross-port, not a new proposal.** The original is pgagroal/pgagroal#1028, already reviewed and **merged**. The change here is the same apart from the project prefix, so it should only need a committer's eye rather than a full review.

Fixes #1281.

Cross-port of pgagroal/pgagroal#1028, which @jesperpedersen asked to have carried across on pgagroal/pgagroal#1027.

`pgmoneta_append_int()` and `pgmoneta_append_ulong()` pass the digit count where `snprintf` wants the buffer size, so each is a byte short and drops the last digit of the largest value of its type:

```
append_int(INT_MIN)      size 11 -> "-214748364"            correct: -2147483648
append_ulong(ULONG_MAX)  size 20 -> "1844674407370955161"   correct: 18446744073709551615
```

Rather than correcting the two sizes, this follows the shape jesperpedersen asked for on the pgagroal PR: both helpers become one-line calls to `pgmoneta_format_and_append()`, which sizes its buffer from `vsnprintf(NULL, 0, ...)` and therefore cannot truncate for any value. The fixed buffers go away entirely.

`pgmoneta_append_ullong()` does not exist here, so only two functions change.

There is no overflow in the current code — `snprintf` truncates safely — so this produces a wrong number rather than a crash, and the reachable impact today is small. I would rather say that than present it as more than it is.

## Verification

Ran the rewritten helpers with `format_and_append` and `append` copied verbatim:

```
append_int(INT_MIN)      = "-2147483648"           len=11
append_int(INT_MAX)      = "2147483647"
append_int(0)            = "0"
append_ulong(ULONG_MAX)  = "18446744073709551615"  len=20
append_int(append_int(NULL, 42), 7) = "427"
```

The last line is the regression check — appending onto an existing string still concatenates.

`clang-format` reports no changes. `clang -fsyntax-only -I src/include` reports the same count of pre-existing errors before and after the change (build-generated defines this environment does not supply), compared against the upstream file rather than a stash, so the count is a real before/after.

### DIFF
--- a/src/include/utils.h
+++ b/src/include/utils.h
@@ -631,6 +631,15 @@ char*
 pgmoneta_append_ulong(char* orig, unsigned long l);
 
 /**
+ * Append an unsigned long long
+ * @param orig The original string
+ * @param l The long
+ * @return The resulting string
+ */
+char*
+pgmoneta_append_ullong(char* orig, unsigned long long l);
+
+/**
  * Append a double
  * @param orig The original string
  * @param d The double
--- a/src/libpgmoneta/utils.c
+++ b/src/libpgmoneta/utils.c
@@ -1593,53 +1593,37 @@ pgmoneta_append_char(char* orig, char c)
 char*
 pgmoneta_append_int(char* orig, int i)
 {
-   char number[12];
-
-   memset(&number[0], 0, sizeof(number));
-   pgmoneta_snprintf(&number[0], 11, "%d", i);
-   orig = pgmoneta_append(orig, number);
-
-   return orig;
+   return pgmoneta_format_and_append(orig, "%d", i);
 }
 
 char*
 pgmoneta_append_ulong(char* orig, unsigned long l)
 {
-   char number[21];
+   return pgmoneta_format_and_append(orig, "%lu", l);
+}
 
-   memset(&number[0], 0, sizeof(number));
-   pgmoneta_snprintf(&number[0], 20, "%lu", l);
-   orig = pgmoneta_append(orig, number);
-
-   return orig;
+char*
+pgmoneta_append_ullong(char* orig, unsigned long long l)
+{
+   return pgmoneta_format_and_append(orig, "%llu", l);
 }
 
 char*
 pgmoneta_append_double(char* orig, double d)
 {
-   char number[21];
-
-   memset(&number[0], 0, sizeof(number));
-   pgmoneta_snprintf(&number[0], 20, "%lf", d);
-   orig = pgmoneta_append(orig, number);
-
-   return orig;
+   return pgmoneta_format_and_append(orig, "%lf", d);
 }
 
 char*
 pgmoneta_append_double_precision(char* orig, double d, int precision)
 {
-   char number[21];
-
    char* format = NULL;
    format = pgmoneta_append_char(format, '%');
    format = pgmoneta_append_char(format, '.');
    format = pgmoneta_append_int(format, precision);
    format = pgmoneta_append_char(format, 'f');
 
-   memset(&number[0], 0, sizeof(number));
-   pgmoneta_snprintf(&number[0], 20, format, d);
-   orig = pgmoneta_append(orig, number);
+   orig = pgmoneta_format_and_append(orig, format, d);
 
    free(format);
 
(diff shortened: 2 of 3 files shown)


Pranav-error:
Hardened `format_and_append()` in the same change, so the review point is addressed rather than deferred.

This was my mistake: I spotted the unchecked `malloc` while writing the pgagroal PR and only mentioned it there instead of fixing it. Routing more callers through the function is exactly when it should have been dealt with.

```c
va_start(args, format);
len = vsnprintf(NULL, 0, format, args);
va_end(args);

// Leave buf as it is on failure, like _append() does
if (len < 0)
{
   return buf;
}

formatted_str = malloc((size_t)len + 1);

if (formatted_str == NULL)
{
   return buf;
}
```

Returning `buf` untouched on either failure matches what `_append()` already does for its own allocation failure, so callers see no change in contract. It also fixes a second thing the old code had: `size_needed` was `vsnprintf(...) + 1`, so a negative return became `0`, and `malloc(0)` followed by use of the result was undefined rather than detected.

Driven through both paths with the function copied verbatim and `malloc` forced to fail:

```
normal   : "-2147483648" len=11
appended : "-214748364818446744073709551615"
on OOM   : "-214748364818446744073709551615"  (pointer unchanged, no crash)
```

`clang -fsyntax-only` and `clang-format` clean; pre-existing error counts unchanged against the parent commit.

Pranav-error:
Added `append_ullong()` here too, so all four projects expose the same three helpers.

```c
char*
pgmoneta_append_ullong(char* orig, unsigned long long l)
{
   return pgmoneta_format_and_append(orig, "%llu", l);
}
```

Declared in `utils.h` next to `append_ulong` with the same doc-comment shape as pgagroal's. It has no callers in this project yet — it is there for parity, matching what pgagroal already has.

`clang -fsyntax-only` and `clang-format` clean on both files; pre-existing error count unchanged against the parent commit.

Pranav-error:
`build-linux (clang, Debug, 14)` is red on this run and it is not from the change.

It did not fail a test — the step hung. Every test that ran passed; the log stops on entering the `delete` module at 15:29 and then `The action 'Run Tests' has timed out after 45 minutes`.

The tree here is byte-identical to the one that passed this job at 11:06 yesterday — `f9e1fd7e` and `9fee6a2c` have the same tree hash. The only difference is the commit message, which I rewrote to carry the description after @fluca1978 asked for that on pgagroal/pgagroal#1036. So the same content was green and is now red with no code difference.

This is also the fourth time I have seen this specific 45-minute `Run Tests` timeout on this repo, at different points in the suite and on different clang versions, including once on a branch that changed no source at all. `feat/gcs-backup`, `feat/empty-folders` and `Test/Replica-backup` have red runs in the same window, so it does not look specific to my branches either.

I cannot re-run jobs here. Happy to push an empty commit to retrigger if that is easier than a maintainer re-run.

jesperpedersen:
Rebased and merged.

PR #1283, opened 2026-09-07
[#1261] Preserve empty directories in the backup manifest

### LINKED ISSUE #1261: Restored S3 backup fails to start
When I restore a backup from S3 and then try to load that data into PostgreSQL, the server fails to start:

`pg_ctl -D /tmp/restore/primary-20260829135838 start`
```
waiting for server to start.... stopped waiting
pg_ctl: could not start server
Examine the log output.

```
and pg logs says: 
```
2026-08-29 14:08:36.352 IST [30157] FATAL:  could not open directory "pg_notify": No such file or directory
2026-08-29 14:08:36.352 IST [30157] LOG:  database system is shut down
```
On looking though, the cause is empty directories. The S3 upload is driven by `backup.manifest`, which records files only, so directories containing no files are never uploaded, and the restore has no way to recreate them.

addititional detail: 
A local backup and restore is working fine, since it copies the actual directory tree rather than going through the manifest.

shashidhar-bm:
-  fixes: #1261 
- related issue: #1154 

what changes??
(for S3)
At backup, it records the data directory's folder layout into a new `backup.dirs` file and uploads it with the other metadata. At restore, it reads that file back and rebuilds those folders, since the manifest lists only files and so loses any folder that is empty.

new testcases: 
- `test_s3_restored_cluster_starts`: which boots a PostgreSQL on the restored directory (#1154 for s3) cc: @Jubilee101 ,
- `test_s3_restore_recreates_empty_directories`: restores a backup and checks the normally-empty PostgreSQL folders

### DIFF
--- a/src/libpgmoneta/se_s3.c
+++ b/src/libpgmoneta/se_s3.c
@@ -87,6 +87,7 @@ static int s3_sign_request(char* method, char* canonical_uri, char* query_string
 static int s3_apply_signed_headers(struct http_request* request, struct deque* headers, char* auth_value);
 
 static char* s3_get_host(int server);
+static int s3_restore_directories(char* local_root);
 static char* s3_get_basepath(int server, char* identifier);
 static char* s3_url_encode(char* str);
 static char* s3_label_from_common_prefix(char* prefix);
@@ -864,6 +865,12 @@ s3_storage_restore(char* name __attribute__((unused)), struct art* nodes)
       goto error;
    }
 
+   /* directory rows are excluded from the file list, so recreate them here */
+   if (s3_restore_directories(local_root))
+   {
+      goto error;
+   }
+
    manifest_tmp = pgmoneta_append(pgmoneta_append(NULL, local_root), "backup.manifest.tmp");
    manifest_final = pgmoneta_append(pgmoneta_append(NULL, local_root), "backup.manifest");
    sha512_tmp = pgmoneta_append(pgmoneta_append(NULL, local_root), "backup.sha512.tmp");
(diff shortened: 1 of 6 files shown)


src/libpgmoneta/se_s3.c
```diff
@@ -881,6 +890,22 @@ s3_storage_restore(char* name __attribute__((unused)), struct art* nodes)
       goto error;
    }
 
+   {
+      char* dirs_tmp = pgmoneta_append(pgmoneta_append(NULL, local_root), "backup.dirs.tmp");
+      char* dirs_final = pgmoneta_append(pgmoneta_append(NULL, local_root), "backup.dirs");
```
jesperpedersen: This looks wrong - we have the manifest

shashidhar-bm: yes @jesperpedersen, but the problem is that the manifest is built to work at the file level.
eg, lets say we added this empty dir `pg_notify,<any checksum>` to the manifest. Every line looks the same to the code, so

- The download loop reads that line and tries to fetch `pg_notify` from S3 -> fails, no such object
- The verify code reads that line and tries to open `pg_notify` as a file to fingerprint it -> fails, it's a folder

Is there any other approach you have in mind?

Jubilee101: ..I've been away for too long so I may be missing context. Can't you use `/` to distinguish a directory? Put `pg_notify/, <some checksum?>` into the manifest for instance.

shashidhar-bm: oh yess, I tried with `pg_notify/, -` and it worked!, thank you, i'll update the code,

shashidhar-bm:
@jesperpedersen , Once this is in place, I’ll update #1235  to have a testcase that starts PostgreSQL on the restored Azure backup,

jesperpedersen:
Rebased and merged.

PR #1287, opened 2026-09-08
[#1278]feat: Add Google Cloud Storage backup support

### LINKED ISSUE #1278: feat: Add Google Cloud Storage (GCS) backup support

shashidhar-bm:
- closes #1278


Abdelrhmansersawy:
@shashidhar-bm 
I would take a look at "progress" part tommorow

src/libpgmoneta/se_gcs.c
```diff
@@ -0,0 +1,1419 @@
...
+   {
+      pgmoneta_log_error("GCS upload: failed to read manifest %s", manifest_path);
+      goto error;
+   }
+
+   pgmoneta_deque_iterator_create(paths, &iter);
+
+   if (pgmoneta_is_progress_enabled(server))
+   {
+      pgmoneta_progress_set_total(server, pgmoneta_deque_size(paths));
```
Abdelrhmansersawy, line 567: These numbers never reach the user. The upload runs in the "Remote" step, which is not a progress phase, so done is never reset - it still holds SHA512's count, which is bigger than the N set here.

```
    SHA512 ends:    done=1002, total=1002  ->  100%
    GCS file 1:     done=1003, total=1003  ->  100%
    GCS file 1000:  done=2002, total=2002  ->  100%
```

S3 and Azure are the same today, so not a blocker. So we need a "Remote" phase - I think we can open an issue for.

src/libpgmoneta/se_gcs.c
```diff
@@ -0,0 +1,1419 @@
...
+   if (gcs_send_upload_request(task->local_root, task->gcs_root, task->remote_path,
+                               strlen(task->file_sha512) > 0 ? task->file_sha512 : NULL, task->server))
+   {
+      pgmoneta_log_error("GCS upload: failed %s", task->remote_path);
+      return 1;
+   }
+
+   if (task->progress_enabled)
+   {
+      pgmoneta_progress_increment(task->server, 1);
```
Abdelrhmansersawy, line 510: Same as previous, runs once per file after a successful upload, so the count is right. But it adds into SHA512's leftover done (Not remote phase)

src/libpgmoneta/se_remote.c
```diff
@@ -72,6 +73,9 @@ static const struct storage_engine engines[] = {
    {"Azure", STORAGE_ENGINE_AZURE,
     STORAGE_CAP_RANGE_GET | STORAGE_CAP_PARALLEL_SAFE,
     azure_upload, NULL},
+   {"GCS", STORAGE_ENGINE_GCS,
+    STORAGE_CAP_RANGE_GET | STORAGE_CAP_PARALLEL_SAFE,
+    gcs_upload, NULL},
```
Abdelrhmansersawy, line 78: `storage_engine = s3`, gcs both engines upload at the same time and share one counter, so done/total end at about 2N. Not visible today. Better fixed when the Remote phase lands: set the total once.

Abdelrhmansersawy (changes requested):
Great Work, I have left some suggestion.

shashidhar-bm:
thank you @Abdelrhmansersawy , i've created issue #1303 to track it.

PR #1289, opened 2026-09-09
Cross-port: Build the value strings with format_and_append

### LINKED ISSUE #1288: value.c callbacks use fixed buffers; the float and double ones truncate
The twelve `*_to_string_cb` callbacks in `value.c` each format into a fixed `char buf[MISC_LENGTH]` and then append it:

```c
char* ret = NULL;
char buf[MISC_LENGTH];

ret = pgmoneta_indent(ret, tag, indent);
memset(buf, 0, MISC_LENGTH);
pgmoneta_snprintf(buf, MISC_LENGTH, "%" PRId8, (int8_t)data);
ret = pgmoneta_append(ret, buf);
```

`pgmoneta_format_and_append()` does this in one line and sizes the allocation from `vsnprintf(NULL, 0, ...)` instead of a fixed buffer.

**Two of them truncate.** `MISC_LENGTH` is 128, but `%f` prints the whole integer part, so a large value needs far more:

```
%f of 1e300 needs : 308 chars
buffer holds      : 127
produced strlen   : 127  -> TRUNCATED
```

`float_to_string_cb` and `double_to_string_cb` therefore emit a silently wrong number rather than failing. The other ten cannot overflow — an integer or a pointer always fits in 128 — so for those this is simplification only.

Cross-port of pgagroal/pgagroal#1040, per @jesperpedersen's "all projects should be the same within this area".

Pranav-error:
> **This is a cross-port, not a new proposal.** The original is pgagroal/pgagroal#1041, already reviewed and **merged**. The change here is identical apart from the `pgmoneta_` prefix, so it should only need a committer's eye rather than a full review.

Fixes #1288.

The twelve `*_to_string_cb` callbacks each declared a fixed `char buf[MISC_LENGTH]`, memset it, formatted into it and appended it. `pgmoneta_format_and_append()` does that in one line and sizes the allocation from `vsnprintf(NULL, 0, ...)`, so the fixed buffers go away entirely: 48 lines removed, 13 added.

It also fixes a truncation. `MISC_LENGTH` is 128, but `%f` prints the whole integer part:

```
%f of 1e300 needs : 308 chars
buffer holds      : 127
produced strlen   : 127  -> TRUNCATED
```

So `float_to_string_cb` and `double_to_string_cb` were emitting a silently wrong number for large values. The other ten format an integer or a pointer, which always fit in 128, so for those this is simplification only.

## Verification

Every removed line belongs to the pattern: 12 x `char buf[MISC_LENGTH];`, 12 x `memset`, 12 x `append(ret, buf)`, and the 12 `snprintf` calls that became the `format_and_append` arguments. No other line in the file changed.

### DIFF
--- a/src/libpgmoneta/value.c
+++ b/src/libpgmoneta/value.c
@@ -419,12 +419,9 @@ static char*
 int8_to_string_cb(uintptr_t data, int32_t format __attribute__((unused)), char* tag, int indent)
 {
    char* ret = NULL;
-   char buf[MISC_LENGTH];
 
    ret = pgmoneta_indent(ret, tag, indent);
-   memset(buf, 0, MISC_LENGTH);
-   pgmoneta_snprintf(buf, MISC_LENGTH, "%" PRId8, (int8_t)data);
-   ret = pgmoneta_append(ret, buf);
+   ret = pgmoneta_format_and_append(ret, "%" PRId8, (int8_t)data);
    return ret;
 }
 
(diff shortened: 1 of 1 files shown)


Pranav-error:
For reference: the pgagroal original, pgagroal/pgagroal#1041, was merged today. This is the identical change — same twelve callbacks, 48 lines removed and 13 added.

jesperpedersen:
Rebased and merged.

PR #1291, opened 2026-09-10
Check the allocations in rfile.c

### LINKED ISSUE #1290: Unchecked allocations in rfile.c
Two allocations in `rfile.c` are used without being checked.

**`pgmoneta_rfile_create()`** — the `rfile` itself:

```c
fp = fopen(extracted_file_path, "r");

if (fp == NULL)
{
   ...
   goto error;
}
rf = (struct rfile*)malloc(sizeof(struct rfile));
memset(rf, 0, sizeof(struct rfile));

rf->fp = fp;
```

On failure `memset` writes through NULL. There is a second consequence: `fp` is an open handle at that point and does not become owned by `rf` until the line after the `memset`, so the `error:` label cannot recover it — `pgmoneta_rfile_destroy(rf)` returns immediately for a NULL `rf`, and the file descriptor leaks.

**`pgmoneta_rfile_initialize()`** — the block-number array:

```c
rf->relative_block_numbers = malloc(sizeof(uint32_t) * rf->num_blocks);
nread = fread(rf->relative_block_numbers, sizeof(uint32_t), rf->num_blocks, rf->fp);
```

The result goes straight into `fread` as the destination. Worth noting what is *not* wrong here: `rf->num_blocks` is read from the file header but is validated against `relsegsz` a few lines earlier, so the size cannot be attacker-inflated — this is an ordinary unchecked allocation, not an untrusted-size one.

Both are OOM-only paths, so practical severity is low. Same class as the allocations fixed in #1246.

Pranav-error:
Fixes #1290.

`pgmoneta_rfile_create()` allocated the `rfile` and went straight into `memset`, and `pgmoneta_rfile_initialize()` passed an unchecked `malloc` result to `fread` as the destination.

The first one has a consequence beyond the null dereference. `fp` is an open handle by that point and only becomes owned by `rf` on the line *after* the `memset`, so the `error:` label cannot recover it — `pgmoneta_rfile_destroy()` returns immediately for a NULL `rf` and the descriptor leaks. The guard closes `fp` before taking the error path.

One thing I checked and found already correct, so it is not part of this: `rf->num_blocks` is read from the file header, but it is validated against `relsegsz` a few lines above the allocation, so a corrupt backup file cannot inflate the size. This is an ordinary allocation check, not an untrusted-size one.

Both are out-of-memory paths, so practical severity is low. Same class as #1246.

## Verification

Builds clean and `clang-format` reports no changes. I have not driven a real allocation failure through a restore, so this is verified by reading plus the compiler rather than at runtime.

### DIFF
--- a/src/libpgmoneta/rfile.c
+++ b/src/libpgmoneta/rfile.c
@@ -117,6 +117,14 @@ pgmoneta_rfile_create(int server, char* label, char* relative_dir, char* base_fi
       goto error;
    }
    rf = (struct rfile*)malloc(sizeof(struct rfile));
+
+   if (rf == NULL)
+   {
+      pgmoneta_record_failure(failures, "rfile_create: could not allocate rfile for %s (label %s)", extracted_file_path, label);
+      fclose(fp);
+      goto error;
+   }
+
    memset(rf, 0, sizeof(struct rfile));
 
    rf->fp = fp;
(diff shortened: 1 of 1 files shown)


jesperpedersen:
Merged.

PR #1294, opened 2026-09-10
Report failure when sasl_prep cannot allocate

### LINKED ISSUE #1293: sasl_prep returns success when its allocation fails
`sasl_prep()` in `security.c` reports success even when its allocation fails:

```c
*password_prep = strdup(password);
return 0;

error:
   *password_prep = NULL;
   return 1;
```

If `strdup` returns NULL, the function still returns 0. The `error:` label sets `*password_prep = NULL` and returns 1, so the contract is clearly "non-zero means the output is not usable" — but the success path can produce exactly the same NULL output while claiming it worked.

The caller trusts the return value, and the NULL then travels three levels down:

```
security.c:480   status = sasl_prep(password, &password_prep);
                 if (status) { goto error; }        <- status is 0, so it continues
security.c:548   client_proof(password_prep, ...)
security.c:2263  salted_password(password, ...)
security.c:2375  password_length = strlen(password);   <- NULL dereference
```

So an allocation failure during SCRAM authentication becomes a crash in `strlen` rather than a clean authentication failure.

This is an out-of-memory path, so practical severity is low. It is the *shape* that seems worth fixing: every other exit from this function reports failure when the output is unusable, and this one does not.

Pranav-error:
Fixes #1293.

`sasl_prep()` returned 0 after `strdup` without checking it. Every other exit from that function reports failure when the output is unusable — the `error:` label sets `*password_prep = NULL` and returns 1 — so a caller that trusts the return value gets a NULL out-parameter while being told the call succeeded.

The NULL then travels three levels before anything touches it:

```
security.c:480   status = sasl_prep(password, &password_prep);
                 if (status) { goto error; }        <- status is 0, continues
security.c:548   client_proof(password_prep, ...)
security.c:2263  salted_password(password, ...)
security.c:2375  password_length = strlen(password);   <- NULL dereference
```

So an allocation failure during SCRAM authentication became a crash in `strlen` rather than a clean authentication failure.

This is an out-of-memory path, so the practical severity is low — the inconsistent contract is what makes it worth fixing rather than the likelihood.

## Verification

Builds clean and `clang-format` reports no changes. I have not driven a real allocation failure through an authentication handshake, so the call chain above is established by reading rather than at runtime.

### DIFF
--- a/src/libpgmoneta/security.c
+++ b/src/libpgmoneta/security.c
@@ -2064,6 +2064,12 @@ sasl_prep(char* password, char** password_prep)
    }
 
    *password_prep = strdup(password);
+
+   if (*password_prep == NULL)
+   {
+      goto error;
+   }
+
    return 0;
 
 error:


Pranav-error:
The red `build-linux (clang, Debug, 14)` here is the CI timeout, not this change. Every test that ran passed; the log stops after `test_cli_backup [PASS]` and then `The action 'Run Tests' has timed out after 45 minutes`.

I have now filed #1295 with the wider picture: 19 failures in the last 60 CI runs, across five contributors, and `main` itself fails — so a red run currently carries little information about the change under it.

shashidhar-bm (approved):
hi @Pranav-error , I don’t think this can happen in normal use, but then yes if there is such failure,  it says success but hands back NULL value, so its worth having this fix :)

PR #1298, opened 2026-09-11
Cross-port anonymous mmap fd fix from pgagroal

AbdulrahmanXSO25:
Use -1 instead of 0 as fd for the fallback anonymous mmap() in the shared memory creation function to match the initial call and fix portability on FreeBSD and OpenBSD, which require -1 for MAP_ANONYMOUS.

This cross-ports the fix from pgagroal PR 1044:
https://github.com/pgagroal/pgagroal/pull/1044

Update AUTHORS and doc/manual/en/97-acknowledgement.md.


AbdulrahmanXSO25:
@jesperpedersen

jesperpedersen:
@AbdulrahmanXSO25 Same deal here

Issue #1297, opened 2026-09-11
Add path.h/path.c API for path manipulation


Mostafa-Ashraf0:
Create a path.h / path.c API to provide a common abstraction for working with directory and file paths in pgmoneta.

The goal is to make path manipulation easier and reduce duplicated path-handling logic throughout the codebase.

For example, an API such as:

pgmoneta_path_directory_add(..., "..", ...)

could transform:

/pgmoneta/backups/backup/

into:

/pgmoneta/backups/

PR #1301, opened 2026-09-12
[#1297] Add path directory API

Mostafa-Ashraf0:


jesperpedersen:
@Mostafa-Ashraf0 You need author files, documentation, ... Read through DEVELOPERS.md

jesperpedersen:
@Mostafa-Ashraf0 Then we can look at the API

Mostafa-Ashraf0:
@jesperpedersen I have updated the author information and `97-acknowledgement.md` by adding my name and email.

Jubilee101:
@Mostafa-Ashraf0 Thanks for working on this. We need something more object oriented. Think about std::path in Rust. From top off my head we should need
1. Easier path initialization and construction, support both append and snprintf, and 
2. Easy to destroy, lower memory leak risk
3. Able to tell a file from a directory
4. Able to tell a relative path from an absolute path, and conversion between each other
5. Able to trim prefix or suffix, one level of path at a time
6. Get the file extension of a file path
7. Joining two paths

You don't have to put in all of them at one time. But design with such requirements in mind so that you can extend the functionality later.

Mostafa-Ashraf0:
@Jubilee101  Thanks for the feedback. I will spend more time going through the project and thinking about the suitable design.

PR #1311, opened 2026-09-16
[#1310] Delete target base directory on restore failure

### LINKED ISSUE #1310: Failed full restore leaves the target base directory behind
**Describe the bug**

`restore_backup_full` creates `<directory>/<server>-<label>/` at `restore.c` file , so on failure, the `error:` block at `restore` file frees the strings and destroys the workflow, but never removes that directory, so an empty `<server>-<label>/` is left behind

also the `restore_backup_incremental` already handles this at file `restore.c` aroubd (`pgmoneta_delete_directory(target_base_combine)`), and the full path doesn't.

The cleanup node can't cover it also, the `carry_out_workflow` jumps to `error:` on any node failure, and skipping the teardown loop. so any restore failure hits this, also the `storage_engine = s3` is just the easiest trigger. `archive` inherits it in archive.c file around 166 line.

**To Reproduce**

With `storage_engine = s3`, `data/` is pruned locally after a backup, so restore fails at `opendir`:
```
pgmoneta-cli -c pgmoneta.conf backup primary
pgmoneta-cli -c pgmoneta.conf restore primary <label> current /tmp/r1
ls -la /tmp/r1
```
Error 600, and `/tmp/r1/primary-<label>/` remains, empty.

**Version**

0.22.0 (main)

**PostgreSQL**

17.11

**libev**

4.33

**OpenSSL**

3.0.13

**OS**

Pop!_OS 24.04

**Configuration**

* pgmoneta.conf

```
[pgmoneta]
base_dir = /tmp/s3lab/backup
compression = zstd
encryption = none
storage_engine = s3
log_level = debug5
workers = 4

[primary]
host = localhost
port = 6432
user = repl
s3_endpoint = localhost
s3_port = 3900
s3_region = garage
s3_use_tls = off
s3_bucket = pgmoneta-test-bucket
```

**Debug logs**

```
ERROR wf_restore.c:224 Restore: Could not restore primary/<label>
ERROR restore.c:2830 execute/Restore backup
WARN  restore.c:676 Restore: No identifier for primary/<label>
```

LoayTarek5:
Closes #1310

The target base is removed when this run created it, or when the workflow started, so the workflow wipes and repopulates the target regardless of who created it, also before it runs, a pre-existing directory is left alone also.

Tablespace directories are removed only when the workflow started, since `copy_tablespaces_restore` is the only thing that creates them.

aslo archive deletes its target directory on failure, and a restore failure inside archive returns an error instead of exiting silently.


shashidhar-bm (commented):
Can you also cover archive with this? It mkdir’s the target before restore, so an early restore failure still leaves the empty dir.

LoayTarek5:
> Can you also cover archive with this? It mkdir’s the target before restore, so an early restore failure still leaves the empty dir.

Ok, it's really good point

shashidhar-bm:
LGTM, @jesperpedersen , but we need to add testcases for this and similar cases in future, when we introduce fault injection concept into our MCTF

